PR #6, opened 2017-02-21
included a constants class.

lackolaxton:
This is allow the program to store all conversion units etc in one place. Although no classes refer to this i can look at adding that in if you are happy with that.

### DIFF
--- a/Constants.cs
+++ b/Constants.cs
@@ -1,0 +1,29 @@
+﻿using System;
+
+namespace AgOpenGPS
+{
+    public class Constants
+    {
+        public const byte SET_1 = 1;
+        public const byte SET_2 = 2;
+        public const byte SET_3 = 4;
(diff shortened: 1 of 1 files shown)


lackolaxton:
I would like to help out where i can, although i am not part of the forums i have read them and would like to help first by helping refactor but also to help implement features.

farmerbriantee:
Hey that's great! I'm pretty slow at keeping track of where i need to reply. I will take a look at the constants. So much changes so quickly on this project, there are many people interested and it is really hard to find examples - as in pretty much none. So we are learning as we go.

lackolaxton:
ah that is good! If you are happy with the idea of it i will start applying the class it current code. Although it doesn't add "features" as such, it will allow easier implementation later on with features....hopefully!

PR #45, opened 2018-07-11
speed comp steering

apmartin00:
Bit of code that may work to slow steer response at higher speeds but still allow a fast slow speed reaction. arrays will need to be modified for each machine as to the speeds that overcontrol happens.
Would appreciate a simple yes or no as to whether this is useful


apmartin00:
Needs more work

Issue #64, opened 2019-07-29
[FR] Add feature: fertilizer spreader


hagre:
Would it be possible to add a different/larger marking to show spreaded fertilizer? Timing/Distance to the vehicle has to be adjusted as well.
I am thinking of such an equipment https://rauch.de/english/fertiliser-spreaders/mds/mds-17.1-19.1/index.html

ArBe-AU:
Hagre,

Can you provide some more information - alternatively ask your question in the AgOpenGPS thread of the Combine Forum https://www.thecombineforum.com/forums/31-technology/278810-agopengps.html

For the equipment in the link you provided I would make the section width the width the spreader is able to spread the product you are using - according the the website 10 - 18m.

If that doesn't help either post your question in the Combine Forum or provide more details here.

hagre:
Thank you for your replay.

I will post this and further requests on the forum.

Yes, the section size should be up to 15m in my case, but i can use (as a walkaround) more then one section to simulate the full width. The real problem is to mark the backwards sprading area (segment of a ellipse/circle - a triangle could be a good approximation) and calculate the right time to switch the sections on/off. 

Advanced thought ->show the reduced/different intensity of the fertilizer on the edge oft the spreading area of each disc in a differnet color/shading. The spreading areas are normaly overlaping in the middle of the two spreader discs and between the different tracks. To get a better sprading pattern we are driving on tacks with half the max sprading distance (see picture on link - "Abbildung 4" and "Abbildung 5").  An additionaly feature woul be the display of border and boundary spreading (limits the sprading width for one side (left) - smaler triangle.

A good discription and pictures are in german at https://www.dlg.org/de/landwirtschaft/themen/technik/technik-in-der-pflanzenproduktion/dlg-merkblatt-410/

hagre:
Link: https://www.thecombineforum.com/forums/31-technology/278810-agopengps-392.html

farmerbriantee:
Have suspended all pulls for now till we get the procedure and resource issues fixed

PR #67, opened 2019-08-11
New Feature: Disk Spreader Fertilizer

hagre:
Introduction of Disk Spreader Fertilizer.
 to solve my  #64

-> New checkbox in vehicle settings to activate "DiskSpreader"
-> Set Distance of throwing back fertilizer
-> This will be shown on map when switching on section(s)
-> In AUTO mode just outer boundaries will be checked to be inside before switching on individual sections

+ read and save of vehicle with new fearure, but still can read old config files.
+ supersection is disabled to avoid some disturbance of the mapview

ToDo:
* AUTO switching off sections in Turn/Headlands
* overlapping of spreadingarea by the half of the throwing width (or the outer 2 of all 4 sections)
* special combination of 2 sections on each side to build something like an limited width spreading feature close to boundaries
* improve map view
* better checking of already applied areas
* fertilizer rate adjustment

* Arduino hard/software for controler


apmartin00:
Hagre, New feature pull requests should be made to.the AgOpenGps_dev repository,.

hagre:
is now pull request on AgOpenGPS_Dev

PR #71, opened 2019-09-04
Delete .vs and obj directories, update .gitignore

travistymartini:
Deleted ".vs" directory as it is not intended to be source controlled.
Deleted "obj" as it contains only binary files generated by the compiler
and build process.

Adjusted ".gitignore" to ignore the two above directories by matching
directory name, instead of absolute path. Should should ensure that the
directories are ignored regardless of directory structure.


farmerbriantee:
Merged

PR #79, opened 2019-10-03
fixed runtime error when trying to calculate turn headings where no t…

buttontate:
I noticed a runtime error when trying to creating a field boundary in a straight line. The app didn't seem to recover will so this should help.

Is the Development branch the correct branch to PR into?


farmerbriantee:
We have been having so much trouble with pull requests, I just added the if statement. Thank you so much for doing this - I hope it is ok to do this. I added a line in the code accrediting Tate Button to the code mod. Git updates and winforms is a real challenge!

buttontate:
It's no problem. 

Ultimately the Application directory should be ignored from source control. The Dependencies and and YouTurnShapes that live in Application could live in the solution and copied to output directory. This way you wouldn't have anything compiled being shown as a diff in the PR. 

If you wanted somebody to be able to download the code we could just use a build agent to build the solution and create GitHub Releases. Releases could be split out (AgOpenGps, Arduino, and PCB) then zipped independently. If an installer was created then a separate installer could be included as well. Would look something like:
Release 1.X.X.X
      AgOpenGps.zip
      AgOpenGps.msi
      PCB.zip
      Arduino.zip

I could setup and Azure Devops pipeline that could do all of this. Additionally this would add PR Verification (Build and Unit Testing) for free and would keep code that has problems out of the Development and master branches. Would be nice to see all PR's do a Build and Test, all merges into Development would build, test, and create a Nightly Release, all merges into master build build, test and create a Stable Release.

I could set this up if you would like?

PR #86, opened 2019-10-07
[Do Not Merge] Continuous Integration/Continuous Deployment Pipeline 

buttontate:
Do Not Merge... quite yet. 

I nuked the Application folder which contained the compiled code as well as update the git ignore to exclude the build directly.

All static references were either brought in via nuget or switched to reside in the GPS/dependencies directory.

One the pipeline is using the base repository we will no longer need the Application folder as the compiled source code will be tracked by releases.


SourceCode/GPS/Forms/OpenGL.Designer.cs
```diff
@@ -30,7 +30,7 @@ private void oglMain_Load(object sender, EventArgs e)
             oglMain.MakeCurrent();
             LoadGLTextures();
             GL.ClearColor(0.5122f, 0.58f, 0.75f, 1.0f);
-            GL.BlendFunc(BlendingFactorSrc.SrcAlpha, BlendingFactorDest.OneMinusSrcAlpha);
+            GL.BlendFunc(BlendingFactor.SrcAlpha, BlendingFactor.OneMinusSrcAlpha);
```
buttontate: @farmerbriantee I had to change this up a bit. The enum has the same value so I think its fine? I have no experience with OpenGL so I'm going to have to do some reading

PR #93, opened 2019-10-12
Remove compiled files in repo

buttontate:


farmerbriantee:
There are many conflicts.

buttontate:
Fortunately its all the files we are removing. Every time there is a merge into master this branch will have merge conflicts. I can merge those changes though.

Once this gets merged in I'll need a gitHub dev token with the whole repo scope. You will have control of this token but should not be shared publicly.  This is so Azure can write releases to GitHub as well as make pipeline builds available to GitHub. It's not a big deal that this get's done straight away as we can build manually and create the release.

Can I walk you through what the pipeline does? Maybe there needs to be some discussion before this goes in?

buttontate:
@farmerbriantee looks like you did something similar. I can close this PR. 

I saw you kept OpenTK references as hard references. Are you opposed to using Nuget?

PR #95, opened 2019-10-14
Adding of the last gStr variables with french translations

Urepom:


farmerbriantee:
There are many conflicts. winforms makes merging extremely difficult

PR #97, opened 2019-10-17
updated git ignore file to exclude application among other visual stu…

buttontate:
…dio ignorables


buttontate:
This PR adds the /Application to the git ignore as well as adds other visual studio recommendations. Basically this is the default visual studio git ignore with /Application added as well.

farmerbriantee:
There no longer is an application with source, the exe is now part of releases and tags

PR #113, opened 2019-11-12
Topmost on Picker

farmerbriantee:
Show in taskbar not set


farmerbriantee:
MAtched up with release 3.0.7

PR #133, opened 2019-12-10
Added flag edit and save

KentStuff:


KentStuff:
The individual files are gone.  Just changes the flagpts[].
AOG  flags

Issue #144, opened 2020-01-12
millis() returns unsigned long


mnltake:
https://www.arduino.cc/reference/jp/language/functions/time/millis/

AutosteerPCBv2.ino　line:163
```
  unsigned int lastTime = LOOP_TIME;
  unsigned int currentTime = LOOP_TIME;
```
↓
```
  unsigned long lastTime = LOOP_TIME;
  unsigned long currentTime = LOOP_TIME;
```

mnltake:
updated V4

PR #145, opened 2020-01-16
use unsigned longs for handling value from millis()

paulortman:
addresses #144


farmerbriantee:
Outdated but already modified

PR #146, opened 2020-02-11
Hungarian translation added and TCP connection possibility

robi772:
Hungarian translation and TCP connection possibility to RTKLib TCP server


farmerbriantee:
Can the RTKLib be connected by udp? 

The language translation is very picky to import and get working. There have been many new keys added so unless you add the individual translations to the most current  Excel files it usually corrupts the resource file since Github won't index the translations. 

There is a topic here, https://agopengps.discourse.group/t/languages-and-translations/1018/21 where people have requested to add a language, the current language file is posted, modified with new langauge, none of the keys are changed, the filled in file is posted, imported, and then the language is in. Long procedure, but the only way we have had continued success.

farmerbriantee:
Also, I noticed you are using the individual resx languages for forms renders the pull completely unable to work. It's a known winforms issue, which is why there is a single string file for each language for the entire program. There can't be anything in the resx files for the forms

robi772:
Unfortunatly the RTKLIB cant support the UDP protocol. :( without the translation can merge just only the TCP connection?

MrPoke21:
Windows or linux version?

(Linux egyszerűen rávehető az UDP re. Windows verzió is szerintem csak azt nem néztem még. RTKLIB támogatja az UDP-t csak az opció a bekapcsoláshoz hiányzik.....)

farmerbriantee:
Cannot import pull, is impossible

PR #152, opened 2020-03-31
Added Keyboard

ArBe-AU:
Brian,

I have 'duplicated' your Keypad to create a Keyboard.

I have also added the ability to enable or disable both the Keypad and Keyboard.  Keypad is enabled by default but the Keyboard must be enabled before it can be used.  This will ensure that the way AgOpenGPS currently functions will continue and end users will only see the addition of the keyboard if they choose to enable it.

For now it is only an English keyboard as I wanted to get this out so I can get my tractor ready for seeding shortly.

Will endeavor to add other languages.


farmerbriantee:
Pull requests need to go to secondary branches

PR #168, opened 2020-06-04
Added pulse/on-off/bang-bang valve support to AgOpenGPS

canadaboy25:
Hi Brian,
I have spent some time getting the on-off pulse valves working with AOG and now have them working very well.  I added a tab to the autosteer setting page in AOG to adjust the settings required for using the valves.  I also added a new Arduino settings sentence that is sent to the Arduino when one of the new settings are changed.  I tried to follow your style of programming as closely as possible so that it fit into the original code quite well.
I posted a description of what all the settings do and how they work here: https://agopengps.discourse.group/t/added-pulse-on-off-bang-bang-valve-support-to-agopengps/2233

If you could take a look at it that would be great because if this or something similar could be added to the main program it would save me having to manually code it in each time there is an update and I know there were some others looking for this functionality too.

Thanks


farmerbriantee:
v5 Now

PR #232, opened 2021-12-25
Record Path:

stroblhofwarte:
Hi,
i have extended the record path feature to be able to have more than one path per field. For my usage the record path feature is more usable than the AB line/curve feature, I use AgOpenGPS for mowing a solar system. The new code is tested with simulation only (winter time). 

- File dialog added to be able to handle more than one recorded path per field
- A recorded path must be named after record
- Start replay of a recorded path tries to lock on the path on the position nearest by the current position of the vehicle. The path is not started always from the beginning.
- Starting replay a path near by the end will resume the vehicle to the beginning of the path (old behavior).

** Tested with simulation only yet **
** Also tested with real world tractor **


farmerbriantee:
We haven't forgot about this - but there weren't going to be any changes in v5.4 so all the docs could be done. But also this is a good addition.

farmerbriantee:
Created a branch on AgOpen called RecordedPathOthmar. Are you going to make more changes?  

Found a few bugs, like when opening an existing field it crashes when opening a non existent recorded path hitting select, but that is just the fun of this - all the things you can do besides using it properly and old field compatibility.

dairytech:
This could be really useful to me.  Thanks for the addition!!

farmerbriantee:
I am just taking a few bugs out and it should be on github on the branch RecordPathOthnar

farmerbriantee:
Branch was manually pulled in.

PR #248, opened 2022-02-27
AgIO with some new functions

GormR:
- COM port auto detection (sometimes Win10 seem to be a bitch concerning the COM ports list)
- dual-GPS merging (not sure, if the data mapping is fully correct; the GPS position calculated is in the middle of the two antennas)
- sensor fusion with BNO085 RVC mode (open question: what is better? taking the last IMU dataset, the last 3 sets or all of the last 100ms????)

### DIFF
--- a/SourceCode/AgIO/Source/AgIO.csproj
+++ b/SourceCode/AgIO/Source/AgIO.csproj
@@ -61,6 +61,8 @@
     </Reference>
     <Reference Include="System" />
     <Reference Include="System.Configuration" />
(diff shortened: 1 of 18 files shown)


GormR:
to do (optional): input mask for further device VID and PID values to be more flexible

farmerbriantee:
Be sure to have fully tested and confirmed by others that all is functioning 100% as well as not affecting or limiting existing function and operation. Would be much better just to make a separate branch.

PR #272, opened 2022-07-30
Witmotion support

Xilofeon:
added support for Witmotion IMU (wt61p or wt901)


richardklasens:
There are some changes in the autosteer PGN data, that won't work

PR #297, opened 2023-01-29
Replace UDP broadcast by unicast for localhost

sonicpp:
Generally it is not good idea to use broadcast on localhost, moreover it is not enabled on most Linuxes.

Without this patch, AOG does not work on Linux with Wine, neither with Mono.

### DIFF
--- a/SourceCode/AgIO/Source/Forms/UDP.designer.cs
+++ b/SourceCode/AgIO/Source/Forms/UDP.designer.cs
@@ -29,8 +29,8 @@ namespace AgIO
         public bool isUDPNetworkConnected;
 
         //2 endpoints for local and 2 udp
-        private IPEndPoint epAgOpen = new IPEndPoint(IPAddress.Parse("127.255.255.255"), 15555);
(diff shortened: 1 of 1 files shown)


farmerbriantee:
The only problem is that it doesn't work in Windows if it isn't broadcast. If you can figure out why that is - then can change to straight localhost 127.

farmerbriantee:
The other applications, namely AgDiag and AgRate and any other plugins people have written all use 127.0... with different ports. But you can't use the same address with different ports on the same machine. And broadcast is used so you can talk to all the plugins (which is the purpose of broadcast)

gunicsba:
@sonicpp  Maybe this variable could be refactored and loaded from configuration ? That way on linux you could override it?

farmerbriantee:
That should be possible yes

farmerbriantee:
Only you can answer that. Your pull request breaks the communication and makes AOG not work. 

Would you like me to delete/close this pull request until you figure out how to make it work?

sonicpp:
Yes, I see that you added option to change the loopback settings, which is nice. And since I do not have the fix which works also on Windows, I am closing it for now.

PR #299, opened 2023-02-03
Fix file names for case sensitive systems

sonicpp:
Fixes building on Linux


farmerbriantee:
The trouble is windows overwrites these files when you add an image etc thru the designer - it will still use any name it wants, this is going to be a constant problem.

sonicpp:
I understand. Hmm. So we can be careful when adding new files :) (maybe there is an option in the IDE somewhere for that? I do not know) or I can just send these MR from time to time to fix that for other OSes.
Or you can ignore this and I will keep these fixes in some separate branch in the fork. What do you prefer?

farmerbriantee:
I'm not sure what is best to do. If fixed in Master what i could fix like the project file. 

Should we delete this pull?

sonicpp:
Well personally I would prefer to merge it, since it should not break things on Windows, only fix things on Linux. And we could see how often the files breaks again.

But it is up to you of course. If you do not want it, I can (or you) close it without merging.

sonicpp:
Damn, I accidentally hit the close and comment instead just comment. Sorry.

farmerbriantee:
Will be doing an update to 5.7.4 some time soon. There are too many changes to add this commit, but after that if you are interested the names could be set to linux happy. Does that work?

sonicpp:
Yes, thank you very much!

jmnc2:
Sorry if this is kind of off-topic.
But does this mean you can run AgOpenGPS on Linux instead of Windows, is there any documentation for Linux and what computer you would need?

sonicpp:
> Sorry if this is kind of off-topic. But does this mean you can run AgOpenGPS on Linux instead of Windows, is there any documentation for Linux and what computer you would need?

Here is the discussion forum: [discourse.agopengps.com](https://discourse.agopengps.com/t/wine-bug-fixed-finally-can-again-run-agopengps-under-wine-on-linux/5820/34)

In short: it somehow works on both AMD64 and RPi hardware for me, but there are few glitches and not working saving configuration right now, also some other things (check the link, I have WIP branch with few hot-fixes, will try to send more MR here when fixes makes sense)

PR #300, opened 2023-02-04
Converted the pdf version into markdown.

gunicsba:
Tried to use as much of the images from the SourceCode as possible.

I used this editor with Visual Studio 2022 to get preview. https://marketplace.visualstudio.com/items?itemName=MadsKristensen.MarkdownEditor64

Images that I couldn't grab from the SourceCode I put to images folder.

Should make translation easier as we can do comparison between files more easier.

However maybe I should factor this a bit more and aim to the Book repository?


gunicsba:
@lansalot  WDYT?

PR #301, opened 2023-02-05
Added Hungarian language

gunicsba:
During testing the translation I noticed that the [?] button is missing from the Steering Form. I've also fixed that.


gunicsba:
As this was merged to the Nightly branch I'll be closing it.
Thank you Brian!

PR #303, opened 2023-02-16
Fix lag with faster refresh

### LINKED ISSUE #264: 5.2.0 and above versions lag
I'm using AOG on a Z8350 windows 10 tablet, connected through usb/serial to a simpleRTK2B-F9P set to 10hz navigation rate.  
5.1.5 is just butter smooth while moving, but everything after that noticeably stutters. 

Yes, I have seen that its release notes specifically mentioned some kind of cap to "fix delays". But even assuming that was the real solution, there's something absolutely fishy going on with the way updates/refreshes happen. 
I have monitored every single component of the system, and nothing is anywhere close to being a bottleneck (be it gpu, disk or any single core of the cpu). 

*Yet* where the older version can paint a new frame to screen in more or less 90 to 110ms (exactly where you'd expect 10hz to be) everything newer (even with Minimum Frame Pause set to 90ms) takes 200-400ms. 
I definitively cannot confirm any overload because of "making a zillion frame updates". And I get that on a general purpose pc other processes may seldom steal resources and degrade user experience for the foreground app, but this isn't about flukes. It's always like that despite no apparent cause.

mirh:
I don't know why it works. 
I don't know what the code was even for. 
I don't know how GPS processing could choke graphics without even a hair of CPU usage. 

Yet this *actually* fixes #264.

### DIFF
--- a/SourceCode/AgIO/Source/Forms/NMEA.Designer.cs
+++ b/SourceCode/AgIO/Source/Forms/NMEA.Designer.cs
@@ -118,9 +118,6 @@ namespace AgIO
                     logNMEASentence.Append("\r\n" +
                         DateTime.UtcNow.ToString(" ->>  mm:ss.fff ", CultureInfo.InvariantCulture) + "\r\n" + rawBuffer + "\r\n");
                 }
-
-                rawBuffer = "";
-                return;
             }
 
             if (isLogMonitorOn)


farmerbriantee:
If there is greater then 301 bytes - that means the lag is somewhere else. By deleting the rawbuffer = "" forces agio to process all those old sentences which are completely old and useless. 

If your code is getting to this point, there is another problem - it should never get here in a properly setup system

farmerbriantee:
Did you post on the forum or telegram the problem?

mirh:
No, even because other people had already reported the problem in vain by the time I posted #264.

farmerbriantee:
Can you provide a link to where others have reported the problem in vain?

You have to realize, processing old sentences that no longer apply doesn't speed anything up.

mirh:
Seriously, couldn't you read [github.com/farmerbriantee/AgOpenGPS/issues/264](https://github.com/farmerbriantee/AgOpenGPS/issues/264)?
I provided both links (whatever they may be useful for, I also just found [this](https://discourse.agopengps.com/t/dell-venue-11-pro-charging/2720/15) but idk) ***and*** steps to replicate the issue on any pc. 

Processing older sentences may not speed up anything, but yet it makes a difference (I can make you a video if somehow my words aren't still accurate enough). 
As I said maybe you have a race condition, perhaps there's something wrong with the criteria for exclusion.. and almost surely it's *exactly* the fact that you are purposefully deciding to drop them from rendering that creates the slideshow impression.

richardklasens:
Provide a screen shot of what you're sending to AgIO on serial at 9600bps

farmerbriantee:
image

What the heck is your computer doing that is so busy? No wonder you have problems....

farmerbriantee:
image

This is what you want to see - not 40 some percent.

mirh:
> Did you notice you only have 4 frames per second?

That's, like, literally the whole problem here to begin with

> This is what you want to see - not 40 some percent.

And even twice as much usage would be limiting.. *because* what?
I don't know what the tablet may have been doing last week (could have just been checking for updates too) but of course an Atom is going to have proportionally more resources busy than whatever new fancy i7 or something. 

Regardless, 25% is more like it, and the problem is still there. 

> I did look at your link you posted, it is just your computer struggling.

It is struggling.. by doing basically nothing?
And not just in a "subjective" sense that I myself didn't open any program, but even by measuring cpu usage and all?
Hell, it's not even using a full core (let alone in my artificial conditions with the 9600K).

> Doing more operations will never make a computer run faster.

Yet this is what happens. Again, do you need a video to believe me?
You don't have to accept my code, but at least try to entertain the possibility that something may be seriously eluding you. 
I spent countless hours trying to make sense of this, it's not just randomly whining about the first thing that stuck to the wall.

Pat-I:
This seem to be hardware or GPS configuration issue.
What GPS you use and are you REALLY sure you just send GGA and VTG, noting else?
AOG wont be good at handle more messages since v5. Is that the point you want to push?

mirh:
My point was more about trying to find whatever kind of solution/rationale, but then #264 was closed as "completed" and after weeks of silence I decided to push this. 

As I said there, my GPS is just a simpleRTK2B-F9P set to a 10hz refresh rate with their normal rover config. 
I have two in my hands (ordered a year apart from each other, and thus presumably also belonging to different build batches) and I can reproduce this with both. So I don't really think the issue is with the receiver. 

I'm with brian in thinking "something" more or less specific about my [tablet](https://4logistic.de/industrielle-tablets/tablets/436/t935-industrietablet) must be at play here (it is harder to make the problem appear on my desktop, after all) but by the gods it's not ***just*** that the cpu is a slowpoke. 
Then yes, of course I could just (I don't know) throw more money at the hardware, or run with my own fork and call it a day.  
But putting aside that doesn't really seem the right attitude for a community project, the fact that it is happening suggests me that something is rotten in the state of Denmark (even though it rarely comes to notice).

Pat-I:
You cannot use ardusimple s configuration and expect that this is what AOG needs.
Load the configuration in the support folder and report back.

lansalot:
Your processor (7 year old) is _largely_ going to be running at 1.4ghz, which isn't that fast. According to the specs at Intel, the maximum memory it supports is 2gb (despite your link showing some models are 4gb - how much do you have?), but even with 4gb your machine is going to be endlessly paging memory in and out. Fairly willing to bet enabling disk counters ("diskperf -y") and restarting task manager will show a lot of IO, and a lack of memory.

A 64gb drive is quite possibly flash storage (not a typical SSD/M2) and those are pretty slow.

And by your own admission, it's harder to replicate on the desktop, a machine with more horsepower.

Have you at least enabled the Performance power plan in Windows, rather than Optimised/Balanced? Stopped any un-needed services to try and release more CPU/memory to AOG?

mirh:
Seriously guys.. some of the stuff flying here sounds like you don't even know how a computer works. 
4GB (which I have) is more than usable
With paging, even 2 should be enough (we aren't trying to run RDR2 here)
Even if they weren't, that doesn't explain why the simulator is completely fine
Intel may have slow gpus and bad drivers, but usage is still clearly negligible (updated Gen8 is also "stabler" than most setups imo)
If AOG needed an ssd to work, something really wong would be going on (and you wouldn't manage to influence the bottleneck by modifying thread affinity then). 
And last but not least for the tenth time: you cannot claim slowness/potatoness/crappiness/ancientness if NOTHING is nowhere near being maxed out (the fact that I can somehow replicate this even on a gaming pc, should be enough of a crazy concern). 

...
With all of this said, I have to thank Pat because indeed using [your cfg](https://github.com/farmerbriantee/AgOpenGPS_Boards/blob/master/Ublox%20F9P/1.32%20SingleAntennaRover.txt) instead of [ardusimple's](https://www.ardusimple.com/wp-content/uploads/2022/09/simpleRTK2B_FW132_Rover_10Hz-00.txt) washed away every nuisance.
(you could have explicitly mentioned that ***even with my task manager trick*** you couldn't notice anything at all).
After much ado, I could even track down the fateful settings in the text file, to some combination of factors between "C1" and "D1" in the ninth `CFG-VALGET` line. 
Do you happen to know what those entries control (is there some documentation to decode ublox cfgs?), and how in the heaven that could lead to this paradoxical behaviour in AOG?

lansalot:
I know perfectly well how computers work - the information offered was to help you as despite everyone else being fine, you alone had this problem and yet according to you, it was the fault of the code.

Now it's been clearly shown to be your error, I guess this PR can be cancelled.

mirh:
You don't think there's really anything to get at the bottom of here?
Be as it may that the difference (some *01* that you have to turn into *00*) is about some "extra information" more or less useless that "burdens" processing (trying to guess which from just the u-center GUI is kind of a lost cause)...

Why is it that AOG can choke on it, ***unless*** you let it parse \*more\* data?
This is the logical bug that just doesn't make sense.
-------- Messaggio originale --------
Il 2 apr 2023, 18:53, lansalot ha scritto:

mirh:
Ok, so.. The [interface description](https://content.u-blox.com/sites/default/files/documents/u-blox-F9-HPG-1.32_InterfaceDescription_UBX-22008968.pdf) lead me to the culprits in UBX-CFG-MSG.

Hence the final question is: why is it that when GSV, GSA or GLL are enabled, AOG starts be incredibly sensitive even just to "available thread count" (let alone the cpu actually having any cycle to spare or not) ***unless*** you remove the two lines in this PR?

mirh:
According to what brian said above, my mod just has AOG keep even old (as in "processed too late") sentences. 
So it's not even the parsing/filtering phase that I'm crossing off here, but literally just the "drop  useless cruft" step (i.e. paradoxically this could only increase load)

Then on the other hand, I do guess it may make sense for "more data points" to lead to "more displayed frames".. except you would at least expect this to affect latency and responsiveness big time?
Yet I can see no such handicap, no difference whatsoever with the experience of the (now  figured out) "official configuration". 

Plus (and by the gods I cannot underline this enough) how can even there be lag when it's not even \*trying\* to use the entirety of the processing power available?

mirh:
> Any computer from the last 30-years should be able to sink serial data at 115200 baud. And most any 32-bit MCU for that matter, should be fine. 
If the application can’t sink the data there’s either an issue with the app or the OS ignores the serial ports

> The receiver has a finite pool of buffers, they retire as the messages leave all the interfaces they’ve been enabled on. Slow baud rates, if left on UART1 or UART2 with NMEA can cause a log-jam. If this is occurring you’ll get a “$GxTXT txbuf,alloc” warning and it will dump packets. You really need to cull all messages, and interfaces you are not using. This will solve the majority of issues. Application software should recognize the “txbuf,alloc” messages, ideally counting and alerting the users. As more satellites are tracked the more data volume NMEA will generate, issues may appear to come and go.

> In apps check for data flow and things that might wait or block. Watch for dependency on message ordering.

Ok so, I got resident u-blox top expert @cturvey to comment on the issue here. 

I had kind of suspected that the only thing that could still make sense of the situation (apparent stuttering with no particular stress on any system component) was if the "rate limiting" happened at the GPS data source, but before this explanation I just couldn't imagine a mechanism for an application *downstream* to "retrocausally" affect it. 
And after playing with [`IDLEDISABLE`](https://superuser.com/a/1642140/567466) (i.e. turn off the CPU sleep states and every kind of power management, for minimum latency) I also figured out a way programs that sync to some external stimulus/interrupt/signal can be bottlenecked without actually placing any strain on the processor. 

The long story short of my educate bet, is that an unfortunate combination of overly tight timings has AOG trip over itself (like this, minus the fact that I don't know how and where its logics to read/*consummate*/parse new sentences work) 
```
GPS:______update1_______________________________________update2_______________________________________update3____________
APP:_____________wakeup__________*long time*__________process___"we late now, drop it"___"nothing to do, sleep"__wakeup__
```
The extra information by GSV & friends, probably lengthens the "process" step enough, that it gets *caught slightly long*. 
When pruning is kept, this in turn has AgIO wait for the next "totally fresh" update, which in turns causes sentences to pile up in the queue, creating an unstable loop that only has a working window of opportunity every now and then. 

Now, as already guessed and hinted, a [sizeable](https://github.com/torvalds/linux/blob/v6.3/drivers/idle/intel_idle.c#L360L392) part of this effect is due to whatever loony core unparking sluggishness my netbook hardware has (though let's please remember that I can somewhat reproduce this even on my beefy desktop). ***But*** if this is even possible to happen at all, clearly there's something wrong with the "old sentences cut off criteria", its priority inside of code, and/or the way the receiver is allowed to stall and block.

PR #307, opened 2023-04-01
5.7.2 drive in from template

LearningLinuxCNC:
Brian,

Here is my changes to allow for having template fields that are automatically copied to a new field when using the drive in option. I added a toggle button to the JobForm next to the drive in button. This toggle changes from using the templates to using normal drive in field function.

Using templates has another folder in the AgOpenGPS base directory. It is named TemplateFields. Once a field with a boundary basic abcurves, ablines etc. has been created it can be copied to the TemplateFields folder.

When the toggle is set to use template files and the user presses the "Drive In (From Template)" button it will check for the close fields just like before. Then it will bring up the select form if there are multiple fields or it will copy the only field that is close.

The name that is used for the new field is "TemplateFieldName_VehicleName_Date_Time"

This makes it simple for the user to use a single button to create a new field without coverage and have all of the basic field information already set up.

To make this work I added a new setting setDriveInFromTemplate that stores the users preference on if they want to use a template or not.

I also was wondering about the way the drive in function locates close fields. 
The initial point for the field may be all the way across from one entry point. I was thinking about getting the extreme points Lat and Long and comparing a square using the 4 corners of the field to the current position to determine if a field is close or not. The other option would be to allow multiple locations in the field.txt file instead of the single coordinate pair. This is a thought for future work but I thought I might ask you now.


LearningLinuxCNC:
Brian,

I just found a bug. I will have the code create the TemplateFields directory if it does not exist.

LearningLinuxCNC:
I added in the same type of code you have for creating the fields directory for creating the TemplateFields directory.

LearningLinuxCNC:
Well, I meant for these to be two separate pull requests. I guess they are combined at this point.

gunicsba:
WOW! This is awesome!

Pat-I:
But this seen all over-complicated and useless from my point of view, why not just open the field and select "save as" and put the new name? 
I see real advantage in this change.

LearningLinuxCNC:
> But this seen all over-complicated and useless from my point of view, why not just open the field and select "save as" and put the new name? I see real advantage in this change.

Pat, Please see this video showing one of the benefits.

https://youtu.be/DXdWv6W79dM

 You say it is over-complicated but from the user point of view it is much simpler. It is many less button clicks and it prevents the user from needing to type. It also gives a specific naming convention so it is easy to look at fields that are completed to see what they are, what tool was being used, and what date and time without the user having to enter anything.

Another benefit that you can see in this video is there are less template fields to sort thru than there are worked fields. If I have 4-5 pieces of equipment that go over a field in a year then I have 4-5 copies of the base field. I need to find the one I want to copy. In the template fields directory there is only one copy of each field.

I hope this helps clarify. We have used drive in from template this spring and it is a real user experience improvement. Especially with multiple users that want simple interaction (fewer clicks the better).

gunicsba:
this spring we did:
field prep 2 passes
planting
rolling

and this is just the last 2 months....
until autumn there'll be a few more passes

@LearningLinuxCNC  I'll make a separate PR so we could merge all these passes and there KML files into one big farm.kml so it could be 1 click and you'd see each operation on google maps, like I did here using java: https://discourse.agopengps.com/t/aog-kml-file-merger/12724

Pat-I:
I use "save as" around 2 to 4 time an hour when spraying.
How do you name your fields? I name it by "field name" then a space, then "operation" then the date stample and sometimes timestamp.

So in my case the "template" should use the same field name until a space, then use the new operation name.
For exemple: "ch50 semis" should become "ch50 arrosage1"

Do your code works a little like this?

LearningLinuxCNC:
The way they are named is "TemplateFieldName_VehicleName_Date Time"
The template field name is what the field was originally saved as
The VehicleName is what the name is for the equipment configuration in AOG
Then Date and Time is appended at the end.
Underscores are placed between the TemplateFieldName and VehicleName and Date Time

Example
TemplateFieldName = myfield
VehicleName = 8200_planter

Produces a field name = myfield_8200 planter_May.02 07_28

All of this information is readily available to AOG so no typing is needed. 

This assumes you have a different vehicle configurations for semis and for arrosage1.

PR #308, opened 2023-04-02
Added boundary section control on off button to FormGPS in lower butt…

LearningLinuxCNC:
…on bar. This button allows the user to turn on and off boundary section control without going into the settings screen. This is important for a particular application. It is for fertilizer spreading with a dry spreader bed on a truck. The process would be to do the headland pass first. On the headland pass the truck needs to keep applying even if it is applying slightly outside the boundary. A single headland pass is enough to turn in.

Then the inner part of the field is spread. To keep the truck from turning the bed on mid turn in the headlands when the spread width goes outside the boundary the boundary needs to control the section to be off.

Two different scenarios with the same piece of equipment in the same field. Don't want to go to the settings in the middle of doing application each field.

The actual change is minimal once I figured out how you were controlling it already.

I also updated the button image for this to include the triangle for the equipment.


LearningLinuxCNC:
I have found an issue in testing where this button needs to be made visible right after creating a boundary in a field that did not originally have a boundary. I will work on that tomorrow and push the update to this branch.

richardklasens:
Why not use the manual section control on/off here?

LearningLinuxCNC:
> Why not use the manual section control on/off here?

We want to use auto section control. However the path of the machine will put the spread pattern outside the boundary. We could use manual section control however our machine switch is set up to engage auto section control. So each time we turn it off using the machine switch we would then have to turn it on manual again. This would be a pain for the operator.

This button is working perfectly for our application. We could make a setting in the main screen options setting menu to enable or disable this button on screen.

PR #314, opened 2023-04-26
Hungarian Transalation Part 2 and some KML enhancement

gunicsba:
Refactored a few things to gStr so I can improve the UI and remove some english text.

There was also a request to see the worked area in the saves. I was able to add this as a description to KML.

image


gunicsba:
Self close as it was done via #329

PR #322, opened 2023-05-17
Avoid changing culture with language

mirh:
I was wondering why my clock was using the AM/PM convention, and I eventually figured the thing was tied to the UI locale. 
And since this detail isn't mentioned anywhere (and it doesn't seem the most of user-friendliness) I supposed it wasn't desired. 
I also wanted to add a comment in the code, but I just didn't feel comfortable speaking in the plural.


mirh:
God forbid dialogue/explanations happened?

PR #327, opened 2023-05-26
NTRIP async IP lookup

LearningLinuxCNC:
This change runs the DNS.GetHostAddressesAsync. The errors from running this async are handled as well as a null return value. When this is running async the value of isNTRIP_starting is true so no other code regarding the NTRIP connection is ran while this is running in the background.

I believe this handles the issue of AgIO not responding while trying to resolve the URL IP address.


LearningLinuxCNC:
I am closing this one and will put it up against SlowPoke

PR #331, opened 2023-09-25
update to 4.6.2

lackolaxton:


lackolaxton:
Update the app to be explicit. I assume there is an issue with goin to 4.8? It would be good to get the latest framework for performance and additional functionality.

lansalot:
Do you have metrics showing performance improvement, and what additional functionality does this provide?

I've ran it in 4.8 myself with no noticeable issues.

lackolaxton:
Nope, I only put 4.6.2 as its the same "major" version  to reduce any issues as that is what my VS2022 needed.

If people can test 4.8 I would update it to the latest. 

No metrics to show other than MS provided. Moving to 4.8 would make it easier to potentially move to .NET Core+ should it get there

lansalot:
OK, I wouldn't think this is an issue that particularly lends itself towards a PR at this time. No immediate improvement, it doesn't really solve anything. If you had the reworked code that necessitates it, that would be different, but just now it looks like "change for change sake".

Will see what other devs think, but if they want to change the target framework, I'm sure they will in time.

lackolaxton:
Makes sense, happy to hold off. Didnt want to just dump it in another PR when framework upgrades can make or break certain functionality.

Id just get it to 4.8 rather than mess around with 4.6

farmerbriantee:
Leaving it at 4.6 for now. Works and is very well tested so hate to change. Also more compatibility with older .net version and thousands of systems out there that don't get updated

lackolaxton:
Understandable.My thought process was to keep the ui on framework and try and pull some of the logic into different projects (.net standard?) . Simply to allow some better testingSent from my iPhoneOn 27 Oct 2023, at 18:20, Brian Tischler ***@***.***> wrote:﻿
Leaving it at 4.6 for now. Works and is very well tested so hate to change. Also more compatibility with older .net version and thousands of systems out there that don't get updated

PR #341, opened 2023-11-19
Better packet reading.

MrPoke21:
Packet lenght fix.
Add new line separator.
Thread safe read.

Now fix UDP wifi lagg, and work in AGOpenGPS is very old tablet.

Required a arduino firmware packet correction.


lansalot:
That's a lot to take in this time of night.. ;)

Why the CRLF on PGNs?

MrPoke21:
My firs idea, readLine function. This is dont work good.
It is necessary to manage the fragmented packet.
AgIO read NEMA Panda this use new line. I think use this.

mirh:
I seem to understand you only really did take a look at the UDP code. 
Is there any chance you could check the serial part too? I tested your PR early today, and it still didn't fix all the bad I reported in #303.

MrPoke21:
> I seem to understand you only really did take a look at the UDP code. Is there any chance you could check the serial part too? I tested your PR early today, and it still didn't fix all the bad I reported in #303.

It is only client side. Need change teensy etc code..

mirh:
Yes, the problem I was complaining about is client side (and it only pertains GPS data handling).

And my quick and dirty fix was for ParseNMEA to *not drop* the otherwise supposedly old sentences. 
Because (unlike UDP which doesn't even guarantee delivery) serial is FIFO and if you aren't fast enough pulling its data down, it can even pile up and block the flow. 

AgIO is doing something wrong with the way it sleeps between refreshes, but I cannot quite get to put my finger on it.

lansalot:
Are you saying this code depends on firmware changes as well - which aren't supplied?

MrPoke21:
Yes

mirh:
I think it would really help if you could squash some of the duplicate commits, and reword them in english.

PR #373, opened 2024-02-15
Rewritten uTurn algoritm

albvig:
The change I propose is to manage scenarios where the turning radius exceeds the turning radius of the tractor, such as during spraying or fertilizing. Currently, AgOpenGps takes the angle of the boundary hit by the ABline, executes a turn onto a straight line in the direction of the boundary, and then transitions onto the next ABline with a turn. Instead, I suggest that the straight part of the turn follows the boundary before transitioning to the next ABline. This modification would be particularly beneficial when driving in pre-made tracks on the headland or when following a curved boundary.

I am not sure if you can see this, but here is som pictures of what I mean!
Skärmbild 2024-02-15 171353
Skärmbild 2024-02-15 172757

In the code the turn is now made right away instead of one movement within each loop. This makes the calculations a little bit bigger but at the same time there were some simplifications made in the movement algoritm. While I haven't had the chance to evaluate it outside of the simulator mode due to the wet conditions in Sweden, I believe the changes work effectively.

If you have any questions, feel free to ask!

Best Regards
Albin Vigre
Sweden


MrPoke21:
This is very nice work!

albvig:
I forgot to mention that currently, it is only implemented on the AB lines and not on curves. However, it could probably be done in a similar way if there's interest! 

Additionally, the old code is still there but not used, so it could either be removed or this new type of turn could be implemented as a third uTurn option.

farmerbriantee:
Great work! Seems to work very well in my initial testing. As it was a pull into master - I stole the class and popped it into youturn branch.

Any thoughts on curve? Possible?

albvig:
Thank you very much!

If you believe it would be useful, I can certainly look into it this week and possibly next to find a solution to making it work with curves as well!

farmerbriantee:
More then useful! It is the way it should be. Are you on telegram? Want your name to be given credit?

In CTurnlines you changed the delta from 0.005 to 0.001. Is that critical? I left it at 5 and it seems to still work just fine. Less points is always a good thing.

farmerbriantee:
I've also added you as a collaborator. So if you want, use the YouTurn branch directly here and welcome to the craziness!!

albvig:
Happy to hear!

I would love to contribute more to AOG and join the development group, havn´t used telegram before but I don´t think that would be any problem! Regarding to the rather small amount I have contributed so far I don´t know if I deserve to be on the credit list but I would of course be very happy to be in the future! I am also kinda new to github but I have already started working with the curves so I will try to get it into that branch. 

The adjustment from 0.005 to 0.001 was made when in one of my test fields, the boundary points had too little difference between them to be recognized accurately, resulting in the turn line being a few centimeters off. However, this occurred only on one field (a field that was quickly made), and I believe further testing is needed to determine the optimal parameter value.

PR #383, opened 2024-03-26
Fixed typo in GPS Form

spidie:
Found a small typo in the app while playing with the app.


lansalot (approved):
Minor type, shouldn't be an issue

PR #384, opened 2024-03-30
Uturn no help

farmerbriantee:
All the fixes and changes


mirh:
Why was help deleted?

Davidwedel:
Because there was a ton of work translating all the strings. If you want
it, revert the commits back and offer to coordinate and maintain the
translation work.

mirh:
You mean [these](https://www.youtube.com/@AgOpenGPS/videos)?
They are certainly nice to have (especially to explain more complex concepts and with movement), but you sure aren't suggesting that a 10 minutes video is a replacement for a quick button function refresh in the field. 
My old man may be a boomer, but he's still not a post-literate tiktoker. 

And in the name of all that is holy, how is an *untranslated* tooltip worse than ***none at all***?

Davidwedel:
Yes. I sure am. It may have been quick for the end user, but Brian described it as a "frustrating circle of pain". 
> My old man may be a boomer, but he's still not a post-literate tiktoker.
> 
Well, that's your place to provide instructions for him then. 


Because then you would have to translate it. If you didn't know English, would you be happy with an untranslated tooltip? Much better to go with video help, and have YouTube translate.

mirh:
> It may have been quick for the end user, but Brian described it as a "frustrating circle of pain".

Then I'd like to understand that. 

Because if it's just about "needing translations", then 100% it doesn't make sense for *less* choice to be a better thing (even if it meant that that whole settings tabs are missing the tooltips). 
If on the other hand, it's difficult to maintain (not meaning human-localization, but code-wise or whatever) then I could see I'm just ignorant of the nitty-gritty details. 

I always thought that it was already incredible they existed to begin with, because even far more expensive and premium and staffed programs (not even related to agricolture) have something this immediate and effective. 

> Much better to go with video help, and have YouTube translate.

I mean.. if that's the token, then you may as well just machine translate all the help text and call it a day.

Davidwedel:
Yes. Every button has to have code handling for if help btn is activated.

mirh:
I mean.. obviously. 
But except when having to add it the first time (something that would likely happen concurrent to button creation itself) how does it really become an hindrance? Or is the torment only coming later, when you have to manually integrate translations (whenever they come) into the monstrous resx files?

PR #388, opened 2024-04-08
611 fixes

farmerbriantee:


mirh:
The icons are always superb, but it shouldn't come that I have to read the source code to know the name or confirm the function of tramline display override buttons.

PR #392, opened 2024-04-23
Change automatic steering to follow "local" nearest point instead of "global" while following a curve

GNSS-Stylist:
(Leaving this PR as a draft, because this is potentially dangerous change. Of course I tried to be careful, but as this is my first time touching this project, it's a real possibility that I forgot something obvious...)

The previous implementation led the vehicle to sometimes pick a wrong part of a curve when the curve was intersecting itself (AB-curve) or there were more than one parallel lines near to each other (typically occurs when driving along an inner curve formed from a boundary curve and the field's shape is not convex).

This changes the curve-following algorithm to use a search for "local" nearest point instead of "global" (= taking all the points on the curve into account) when already following the curve.

Look ahead-distance is used when searching the nearest "local" point to prevent problems in tight hairpins caused by the search sticking to a local minimum "on the wrong side" of the hairpin (especially when using large look ahead values causing the vehicle to take a shortcut).

Videos from testing (about the same conditions as in the issue):

https://github.com/farmerbriantee/AgOpenGPS/assets/50438441/bd927935-46a5-4962-b4bd-037c4aded5c9

https://github.com/farmerbriantee/AgOpenGPS/assets/50438441/6e0f7b96-86e1-44c7-956f-e09c205e312b

PS: As I now took a look at the code to make these changes, I have to say that I really like the coding style of AOG. Lot of descriptive comments and easy to read overall.

### DIFF
--- a/SourceCode/GPS/Classes/CABCurve.cs
+++ b/SourceCode/GPS/Classes/CABCurve.cs
@@ -58,6 +58,9 @@ namespace AgOpenGPS
 
         public double inty;
 
+        // Should we find the global nearest curve point (instead of local) on the next search.
+        private bool findGlobalNearestCurvePoint = false;
+
         public CABCurve(FormGPS _f)
         {
             //constructor
@@ -466,6 +469,7 @@ namespace AgOpenGPS
             }
 
             lastSecond = mf.secondsSinceStart;
+            findGlobalNearestCurvePoint = true;
         }
 
         public void GetCurrentCurveLine(vec3 pivot, vec3 steer)
@@ -484,6 +488,10 @@ namespace AgOpenGPS
 
             if (curList.Count > 0)
             {
+                // Update based on autosteer settings and distance from line
+                double goalPointDistance = mf.vehicle.UpdateGoalPointDistance();
+                bool ReverseHeading = mf.isReverse ? !isHeadingSameWay : isHeadingSameWay;
+
                 if (mf.yt.isYouTurnTriggered && mf.yt.DistanceFromYouTurnLine())//do the pure pursuit from youTurn
                 {
                     //now substitute what it thinks are AB line values with auto turn values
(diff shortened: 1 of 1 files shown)


GNSS-Stylist:
I drove around with this PR during last summer, first using the version 6.2.0 as a base version and later 6.3.3. Now also tested this on top of the current master ( 637963acb743e68f239f010b21a76c6429d24c71 ) in simulated environment. Didn't notice any side effects so removing the draft-status. Feel free to discard anyway if this doesn't fit or obfuscates the code (personally I don't really like the flag findGlobalNearestCurvePoint used here, but couldn't find any other way to relay this info).

Also removed some whitespace-changes that caused some conflicts with recent changes into the same source file (apparently I had some "autocorrect"-feature enabled on VS and "Ignore whitespace changes" enabled while committing, so didn't notice the changes).

GNSS-Stylist:
So apparently the force-pushed squash-merge (or whatever) with those whitespace-changes prevented this from being merged automatically ("Merging is blocked. The base branch does not allow updates")... Not sure how to fix this (sometimes git(hub) feels like an obstacle course, to be honest).

(Squashing could probably be enabled for the project by following step 4 here(?): https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests ).

richardklasens:
@farmerbriantee did you manually added this?

GNSS-Stylist:
I also thought that similar changes were added manually to the project by @farmerbriantee . Just tried the latest master (https://github.com/farmerbriantee/AgOpenGPS/commit/94d81cb40bb7c2c6e2c133b9a6fd0a24c43386bf) but the judging by the behavior the fix isn't there.

I can try to make a new clean PR based on the current master if you like(?) Brian seems to be the assignee on the issue now, so not going to do that by myself yet.

richardklasens:
@GNSS-Stylist take the shutdown branch as base and try to make a new branch with your changes

GNSS-Stylist:
Superseded by https://github.com/farmerbriantee/AgOpenGPS/pull/427

Tested if the "shutdown"-branch has this fix merged manually. It didn't, so made a new "clean" PR (above).

PR #420, opened 2024-10-28
Update GUI.Designer.cs to load settings for YouTurn SkipWidth

steran82:


farmerbriantee:
Added manually - thank you!!

PR #422, opened 2024-11-05
Update GUI.Designer.cs - fixed bug for UTurn mode after changing configuration

steran82:
fixed bug: 
precondition: UTurn is on, and UTurn mode was set to the alternative mode (colored icon for UTurn: YouSkipOn.png) then: going to the configuration menue and accept (green button). Settings are reloaded. UTurn is set to off (red icon on the right); icon for UTurn mode has still the alternative mode icon (YouSkipOn.png); but in AGO internaly UTurn mode is set to the normal mode. So the UTurn mode does not fit to the icon. When pressing the icon UTurn mode is vice versa.

### DIFF
--- a/SourceCode/GPS/Forms/GUI.Designer.cs
+++ b/SourceCode/GPS/Forms/GUI.Designer.cs
@@ -448,6 +448,7 @@ namespace AgOpenGPS
 
             //OGL control
             isUTurnOn = Properties.Settings.Default.setFeatures.isUTurnOn;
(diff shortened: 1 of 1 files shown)


gunicsba:
I was about to add this to the other PR that I think this button is also buggy.
Basically you have to turn it off then on again so it becomes on.

steran82:
Thanks - should i close the PR, or do you decline it?

farmerbriantee:
I added the code manually. Thankyou!!

PR #424, opened 2024-11-19
Update gStr.fr.resx

simmoinard:
replace "lignes de grandes largeurs" by "jalons" in french


farmerbriantee:
Manually added the translation. Thanks!!

PR #426, opened 2024-12-09
Convert projects to SDK-style

FantasyTeddy:
This converts the project files to the new SDK-style, using the [.NET Upgrade Assistant](https://dotnet.microsoft.com/en-us/platform/upgrade-assistant).

By using the new format, the following benefits are available:
- **Fewer obscure properties**
  - Many properties are set by default and do not need to be specified explicitly. This cleans up the project files and makes them more understandable.
- **Default includes and excludes**
  - Source files (and resources) will be included implicitly, which greatly reduces the necessary changes to the *.csproj files when adding or removing files. This also highlighted some files that were no longer included in the projects (which I removed).
- **Future proof**
  - The new SDK-style project format is necessary if a migration to any newer .NET version is planned. This is most often considered the first step when migrating.

---

PS: Are there any plans towards moving to a newer .NET version? If so, to which version?


lansalot:
What testing has been done to verify normal operation?

FantasyTeddy:
I only did minimal local testing, because I didn't expect anything to break.
What kind of tests would you like me to do?

richardklasens:
created new Branch called SDK

FantasyTeddy:
Okay and do I have to do something now? Sorry I am not familiar with this workflow. Could you explain how this works?

richardklasens:
Our current dev branch is called Shutdown.. compared to that we can’t do a merge due to conflicts. So pull shutdown branch and do a merge locally (use GitHub desktop, easy gui), when all conflicts are solved, you can do a new PR on the shutdown branch

FantasyTeddy:
Okay, I rebased my changes.

Is there a particular reason, why you are working on a branch called "Shutdown" that is essentially your dev branch?
This is very confusing for a new contributor like myself and if I had known in advance, I could have started my work there, instead of on "master".

richardklasens:
No it was just a feature inserted and we developed a lot more under the same branch. From now on we will use dev branch as our new development branch.

FantasyTeddy:
Okay, I understand.

Are you planning on using the [Git Flow](https://nvie.com/posts/a-successful-git-branching-model/) branching model?
Maybe it would help to state this somewhere, or at least that the "dev" branch should be used (maybe even make this the default branch).

PR #427, opened 2024-12-09
Local search instead of global when following a curve

### LINKED ISSUE #391: Curves parallel to each other or crossing cause automatic steering to get confused
Fields with shapes that are not convex may be divided into two or more "sections" when using a boundary curve and an "inner" curve is selected. These sections are connected with two straight lines that are very close to each other. This may lead to automatic steering getting confused so that it doesn't seem to always know which one of those lines it should follow.

Made more thorough testing and videos with this external simulator: https://github.com/GNSS-Stylist/AgOpenGPSSimPoC but the same issue also occurs with the internal simulator and in a real tractor (tested also in a field to make sure this is not just a simulator glitch). Actually very likely stumbled upon this while driving last summer with AOG V5.7.2.

Video:

https://github.com/farmerbriantee/AgOpenGPS/assets/50438441/edd3c106-9930-425e-8f0c-8f6d94d88be9

Based on the lookahead-pointer (yellow square) jumping back and forth it seems that the nearest point is always selected from the whole curve (also by looking at the code it seems to do so).

This also occurs with AB-curve that has self-crossing parts (may not be the most sensible curve though...):

https://github.com/farmerbriantee/AgOpenGPS/assets/50438441/1d97062b-e6e3-46de-b378-3322d0379111

I already took a look at the code and got a fix to this somewhat working. It does, however still have some problems with some tight hairpins when a big look ahead-value is used and tractor is taking a shortcut. I will continue investigating, so assigning this to myself for now (not sure if I can do that, though. But will try at least...). Edit: Can't assign myself, but will continue investigating anyway.

Field used in the videos above: 
[Field_Sim_Blob.zip](https://github.com/farmerbriantee/AgOpenGPS/files/15074113/Field_Sim_Blob.zip)

Reproducible in 6.2.0, 6.3.1, 6.3.3 and latest master ( 637963acb743e68f239f010b21a76c6429d24c71)

Edited: Updated "reproducible in"-list (tested V 6.3.1, 6.3.3 and latest master).

GNSS-Stylist:
The previous implementation led the vehicle to sometimes pick a wrong part of a curve when the curve was intersecting itself (AB-curve) or there were more than one parallel lines near to each other (typically occurs when driving along an inner curve formed from a boundary curve and the field's shape is not convex).

This changes the curve-following algorithm to use a search for "local" nearest point instead of "global" (= taking all the points on the curve into account) when already following the curve.

Look ahead-distance is used when searching the nearest "local" point to prevent problems in tight hairpins caused by the search sticking to a local minimum "on the wrong side" of the hairpin (especially when using large look ahead values causing the vehicle to take a shortcut).

NOTE: Superseded PR was tested in a real tractor but this only on an external simulator. Behavior in the simulator seemed to be identical with the superseded PR. Implementation is not, however, completely identical fix due to the other changes in the source file, especially https://github.com/farmerbriantee/AgOpenGPS/pull/419

See the superseded PR (below) for more info.

Fixes #391
Supersedes #392

### DIFF
--- a/SourceCode/GPS/Classes/CABCurve.cs
+++ b/SourceCode/GPS/Classes/CABCurve.cs
@@ -55,6 +55,9 @@ namespace AgOpenGPS
 
         public double inty;
 
+        // Should we find the global nearest curve point (instead of local) on the next search.
+        private bool findGlobalNearestCurvePoint = true;
+
         public CABCurve(FormGPS _f)
         {
             //constructor
@@ -75,6 +78,7 @@ namespace AgOpenGPS
             if (!isCurveValid || ((mf.secondsSinceStart - lastSecond) > 0.66 && (!mf.isBtnAutoSteerOn || mf.mc.steerSwitchHigh)))
             {
                 lastSecond = mf.secondsSinceStart;
+                findGlobalNearestCurvePoint = true;
                 if (track.mode != TrackMode.waterPivot)
                 {
                     int refCount = track.curvePts.Count;
@@ -182,6 +186,7 @@ namespace AgOpenGPS
 
                     curList = await Task.Run(() => BuildNewOffsetList(distAway, track));
                     isBusyWorking = false;
+                    findGlobalNearestCurvePoint = true;
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


richardklasens:
Merge is pushed to a new branch called follow-ab-curve-path-local-point.
Will close the PR and merge from there.

PR #432, opened 2024-12-14
Use NuGet packages for dependencies

FantasyTeddy:
- Migrate to `PackageReference` (see [Migrate from packages.config to PackageReference](https://learn.microsoft.com/en-us/nuget/consume-packages/migrate-packages-config-to-package-reference) for a list of benefits)
- Replace DLL references with their matching NuGet package where possible
- Reference the `Keypad` project directly


FantasyTeddy:
Thank you for merging my PR.

However, you have reverted it again? Can you tell me why?

FantasyTeddy:
@farmerbriantee, @richardklasens: If I can solve the performance issue introduced with this change, would you consider merging this again?

PR #458, opened 2024-12-21
Elevation.txt with a time stamp and Elevation in Flag forms

Rob2029:
Added a timestamp to the Elevation.txt file, useful to do data analysis with the text file and I hope some velocity, production (as in m2/s) data graphs

Added Elevation to the Flags form and text file, I've also put it in the KML file since VS did not let me build without it. This may be helpful.

Not quite sure if the pn.altitude needs a subtraction of the antenna height, I did not do that

Capture

### DIFF
--- a/SourceCode/GPS/Classes/CFlag.cs
+++ b/SourceCode/GPS/Classes/CFlag.cs
@@ -9,6 +9,8 @@ namespace AgOpenGPS
 
         public double longitude = 0;
 
+        public double elevation = 0; 
+
         //UTM coordinates
         public double northing = 0;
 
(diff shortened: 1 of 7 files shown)


farmerbriantee:
The trouble is now you have made all previous Flag files obsolete due to different structure.
As well as all the elevation files

Rob2029:
ohhh yup, not good

Davidwedel:
Would there be a way to assign a dummy elevation of 0.0 to all flags on startup? Of course, this would break things if a user downgraded from newer to older...

I don't personally see a timestamp on elevation being a crisis. Most users don't upgrade while recording elevation.

Rob2029:
Mmmm don't think there is a way to do that, since I modified the flags.txt structure, for example the KML does have a 0 elevation value so I only added a value without changing the structure

Pat-I:
I  think timestamp in the elevation log is not useful for many people (or only for you) but it doesn't create problems either I guess, since it's at the end of the line.
What is the printed timestamp format exactly?

For flags, an elevation would be nice. But must be made to be able to load old versions. 
I guess we don't need to support backward compatibility for Flags.

Rob2029:
Yes the time stamp mey not be useful to everybody, but I think geographical data always needs a time stamp, is not just where you did what thing but when, that could be useful to manage the data and some scenarios where you can create reports based on data
The format is: mm/dd/yyy hh:mm:ss, example 12/23/2024 23:11:52
If the field did not have any flags there is no problem. If it had flags it just keep adding them, but now has the old flags with 8 values and the new ones with 9 values like this:
$Flags
5
53.4360564,-111.160047,0,0,0,0,0,1,
53.4361859,-111.160047,0,14.4128848,0,0,2,2,
53.4362992,-111.160047,0,27.024159,0,0,1,3,
53.4361838,-111.1601104,262.9415,-4.2123653,14.1750596,5.9943294,0,4,4
53.4363135,-111.1601749,264.8843,-8.5026588,28.612399,5.9943301,0,5,5

PR #470, opened 2024-12-25
Use `Path.Combine` instead of string concatenation

FantasyTeddy:
By using [`Path.Combine`](https://learn.microsoft.com/en-us/dotnet/api/system.io.path.combine) instead of string concatenations, we don't rely on Windows path separator (`\`) and would be compatible with platforms using `/` as well.

### DIFF
--- a/SourceCode/AgIO/Source/Forms/Controls.Designer.cs
+++ b/SourceCode/AgIO/Source/Forms/Controls.Designer.cs
@@ -271,9 +271,7 @@ namespace AgIO
             if (processName.Length == 0)
             {
                 //Start application here
-                DirectoryInfo di = new DirectoryInfo(Application.StartupPath);
-                string strPath = di.ToString();
-                strPath += "\\ModSim.exe";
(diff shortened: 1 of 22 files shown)


richardklasens:
Maybe it’s better to change the workflow directly that the snippets of code aren’t in the designer files?

FantasyTeddy:
If you mean that all custom logic should be in the `FormExample.cs` and not the `FormExample.Designer.cs` files, I absolutely agree.
Would you be interested if I would open a PR that would change this?

farmerbriantee:
Needs to be rebased off the current dev changes

FantasyTeddy:
Thanks for rebasing and merging.

richardklasens:
I think there are more code snippets in designer files…

FantasyTeddy:
What do you mean? Did I forget some string concatenations that should use `Path.Combine`?
Or do you refer to the `*.Designer.cs` files that contain custom logic? If so, I can fix this in a separate PR if you want.

PR #473, opened 2024-12-25
Add dedicated `SystemEventsLogger`

FantasyTeddy:
This change encapsulates all the logic considering the system-events inside a new `SystemEventsLogger` class.

### DIFF
--- a/SourceCode/GPS/AgOpenGPS.csproj
+++ b/SourceCode/GPS/AgOpenGPS.csproj
@@ -665,6 +665,7 @@
       <DesignTime>True</DesignTime>
       <DependentUpon>gStr.zh-CHS.resx</DependentUpon>
     </Compile>
+    <Compile Include="Logging\SystemEventsLogger.cs" />
     <Compile Include="Program.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="Forms\UDPComm.Designer.cs">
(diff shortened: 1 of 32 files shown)


farmerbriantee:
Code gymnastics :( 
Lets get this release out and tested!!! no more changes like this please!

FantasyTeddy:
What you call "code gymnastics", I call "applying best practices".
By encapsulating the logic of writing system-events into their own class, we could change the way the logging is actually done, by changing the implementation of the `SystemEventsLogger`, without affecting anything outside of it.
Originally, I planned to not make the `SystemEventsLogger` property `public` and pass it down to everyone that needs it. This would improve things even more, because it would reduce the dependency to the main `FormGPS`.

If you first want to get a release out, fine, don't merge my PR yet. But calling it "code gymnastics" and rejecting it makes me feel like you don't want any help at all...

richardklasens:
Indeed. I reopened it. We can merge if the release is out and stable. The gymnastics Brian refers to is probably not the best word for it. I like your PR, so does he, but it’s just the wrong time. 

I would say, let’s create a group chat on Telegram with active contributors so we can talk first about our plans before we do a lot of work and then find out we end up in a million conflicts :)

cheers

FantasyTeddy:
Thank you for reopening the PR!
I will gladly rebase it, once the release is out.

Maybe we could introduce the concept of release branches, such that work can continue on `develop` and does not need to be halted when a release is pending. (Especially, because I imagine that "after a release is before a release".)

However, for bigger changes with "millions of conflicts", it could indeed be beneficial to have a channel where this can be coordinated :)

PR #481, opened 2024-12-27
Translation to Spanish 

Rob2029:
Form First
Form Help
Form About
Form SimCoords
Form GPS
Form SteerWiz


richardklasens:
Hi Rob. 
to translate stuff, it shouldn’t be necessary to change all those files. It will probably brake the program and other languages. Best is to only update gstr.es

now a lot of others files are changed. Not sure what this will do?

Rob2029:
I am not sure if I understand, I want to make this right, this is what I did:
Let's say I want to translate label2 in FormBndTool.cs, the label text is "Create", wich translates to "Crear"
I created a String in gStr.es.resx called gsCreate with the text "Crear"
Then in public FormBndTool(Form callingForm) I added this line:
`label2.Text = gStr.gsCreate;`

But I get an error:
Error CS0117	'gStr' does not contain a definition for 'gsCreate'	

Then I added the same String (gsCreate) in gStr.resx with the text "Create" and the error is gone.

The project compiles and runs, if Spanish is Selected and I open the Boundary Tool the label says "Crear"
If English or any other language is selected the label text is "Create"

richardklasens:
I understand now.. you created label symlinks for the resx. There hasn’t anything done yet with translations for the latest version(s) so it makes sense. I think it’s best we first create all symlinks in the designer files and in the gstr resx file, and then walk thru all the translations for multiple languages.

Rob2029:
Mmm ok so what do I need to change? 
I follow what was done to other labels and then in other forms I used the same format in the same function of the Form
At least you can keep the gStr.rex and gStr.es.rex Strings I hope

richardklasens:
I will look into it next days. Thanks for your work anyway!!

Rob2029:
Thanks, I am new to git so I didn't know how to just do a PR for 1 commit 😅

PR #483, opened 2024-12-27
gui fixes readme update

richardklasens:
changed color for sideguidelines
changed config buttons alignment in config
update readme.me

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,10 +36,12 @@ is actively being worked on and may not be ready for production use.
 
 In order to contribute to AgOpenGPS, follow these steps:
 
-1. Checkout the `develop` branch
-2. Create a new branch named after your feature
-3. Make your changes and commit to this branch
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -36,10 +36,12 @@ is actively being worked on and may not be ready for production use.
 
 In order to contribute to AgOpenGPS, follow these steps:
 
-1. Checkout the `develop` branch
-2. Create a new branch named after your feature
-3. Make your changes and commit to this branch
-4. Create a PR targeting the `develop` branch
+1. Please join the Coding Telegram group to talk about your plans
+	https://t.me/+YSawMBD3ybM5YmRk * an admin will give you full access.
```
FantasyTeddy, line 40: Do you really want to make it a requirement to join the Telegram group?
I think for smaller contributions, it should not be necessary to do so.

Maybe put this text after the list, such that it does not feel like it is mandatory?

PR #507, opened 2024-12-31
Merge 'release/6.6' into 'develop'

FantasyTeddy:


SourceCode/AgIO/Source/Classes/CLog.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace AgIO
+{
+    public static class Log
```
J005t67: You dropped the prefix 'C' from the class name. That's perfectly fine for me since it does not add anything in my opinion.
Please change the file name accordingly.

SourceCode/GPS/Classes/CABCurve.cs
```diff
@@ -516,14 +516,14 @@ public void GetCurrentCurveLine(vec3 pivot, vec3 steer)
                 else// Pure Pursuit ------------------------------------------
                 {
 
-                    double minDistA = double.MaxValue;
-                    double minDistB = double.MaxValue;
+                    double minDistA;
```
J005t67: The scope of these variables is too big. Move them inside the if.

SourceCode/GPS/Classes/CLog.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            //system event log file
+            FileInfo txtfile = new FileInfo(logFile);
+            if (txtfile.Exists)
+            {
+                if (txtfile.Length > (sizeLimit))       // ## NOTE: 0.5MB max file size
+                {
+                    StringBuilder sbF = new StringBuilder();
+                    long bytes = txtfile.Length - sizeLimit;
+                    bytes = (sizeLimit *2)/10 + bytes;
```
J005t67: Use spaces around operators

J005t67: Sorry, I don't undestand what the calculation is about. Please add comment line or improve naming of variables if that helps.

SourceCode/GPS/Classes/CLog.cs
```diff
@@ -0,0 +1,66 @@
...
+                    using (StreamReader reader = new StreamReader(logFile))
+                    {
+                        try
+                        {
+                            //Date time line
+                            while (!reader.EndOfStream)
+                            {
+                                bytesSoFar += reader.ReadLine().Length;
+                                if (bytesSoFar > bytes) 
+                                    break;
```
J005t67: Many coding standards forbid single line if statements like this. Either use {} or put the break on the same line as the if.

SourceCode/GPS/Forms/Field/FormFieldISOXML.cs
```diff
@@ -461,7 +460,7 @@ private void btnBuildFields_Click(object sender, EventArgs e)
                 }
                 catch (Exception ed)
                 {
-                    mf.LogEventWriter("Creating new iso field " + ed.ToString());
+                    Log.EventWriter("Creating new iso field " + ed.ToString());
 
                     MessageBox.Show(gStr.gsError, ed.ToString());
```
J005t67: It seems that most occurences of Messageboxes for errors also need a Log.EventWriter. In the future we could make an dedicated ErrorMessageBox that takes care of both the eventLog and the window.

SourceCode/GPS/Forms/FormGPS.cs
```diff
@@ -415,62 +415,23 @@ private void FormGPS_Load(object sender, EventArgs e)
...
             if (!string.IsNullOrEmpty(vehiclesDirectory) && !Directory.Exists(vehiclesDirectory)) { Directory.CreateDirectory(vehiclesDirectory);
-                sbSystemEvents.Append("Vehicles Dir Created\r");
+                Log.sbEvents.Append("Vehicles Dir Created\r");
             }
 
             //get the fields directory, if not exist, create
             logsDirectory = Path.Combine(baseDirectory, "Logs");
             if (!string.IsNullOrEmpty(logsDirectory) && !Directory.Exists(logsDirectory)) { Directory.CreateDirectory(logsDirectory);
-                sbSystemEvents.Append("Logs Dir Created\r");
+                Log.sbEvents.Append("Logs Dir Created\r");
```
J005t67: I see so many \r in the calling code. Maybe we could make that the resposibility of the logging class in the future

FantasyTeddy:
Sorry, these are all valid inputs, but this PR is only the merge of the release branch (`release/6.6`) into `develop` and I'm discarding all the changes, because they are already in `develop`.

SourceCode/GPS/Forms/FormGPS.cs
```diff
@@ -415,62 +415,23 @@ private void FormGPS_Load(object sender, EventArgs e)
...
             }
 
-            //system event log file
-            FileInfo txtfile = new FileInfo(Path.Combine(logsDirectory, "zSystemEventsLog_log.txt"));
-            if (txtfile.Exists)
-            {
-                if (txtfile.Length > (500000))       // ## NOTE: 0.5MB max file size
```
J005t67: No need for the additional () around 500000. I would prefer to write (512 * 1024), because that is the acutal number for 0.5MB and you don't need to count the zero's when checking the code.

J005t67:
> Sorry, these are all valid inputs, but this PR is only the merge of the release branch (`release/6.6`) into `develop` and I'm discarding all the changes, because they are already in `develop`.
> 
> Maybe we can add your proposed changes in a different PR directly to `develop`?

Yes, you are right, this PR is only about the merge.  Please ignore my comments.

FantasyTeddy:
> I think I was so enthusiastic about our new reviewing proces that I overdid it a bit ;-)

I totally understand and I really appreciate the comments. Maybe @farmerbriantee can incorporate these suggestions in a future PR?

SourceCode/GPS/Classes/CLog.cs
FantasyTeddy: This is the only comment I slightly disagree with: Yes, using brackets (`{}`) is probably the best way to format this, but if single-line `if` statements can be used, I don't think it matters if the same-line or the 2-line version is used (they are equally good/bad).

J005t67 (approved):
Feels good to start 2025 with a new review process!
Did not spot any merge problems.

PR #510, opened 2025-01-01
AutoRecBoundary

Tervuren:
Facilitates the creation of field borders, especially in small fields with implements carried on the rear of the tractor, by synchronizing the record/pause function with the activation of sections using the work switch


SourceCode/GPS/Forms/Position.designer.cs
```diff
@@ -1456,116 +1456,54 @@ public void AddBoundaryPoint()
             prevBoundaryPos.northing = pn.fix.northing;
 
             //build the boundary line
-            if (isAutoRecBoundary == true)
+            if ((isAutoRecBoundary == true & (manualBtnState == btnStates.On || autoBtnState == btnStates.Auto)) || isAutoRecBoundary == false)
```
J005t67, line 1459: Please, remove the '==true' if isAutoRecBoundary is a boolean.

J005t67, line 1459: Please use logical and '&&' not binary and '&'

SourceCode/GPS/Forms/Field/FormBoundaryPlayer.cs
```diff
@@ -157,24 +157,26 @@ private void btnStop_Click(object sender, EventArgs e)
...
+            mf.isAutoRecBoundary = chAutoRecBoundary.Checked;
             if (mf.bnd.isOkToAddPoints)
-            {
-                mf.bnd.isOkToAddPoints = false;
-                btnPausePlay.Image = Properties.Resources.BoundaryRecord;
-                //btnPausePlay.Text = gStr.gsRecord;
-                btnAddPoint.Enabled = true;
-                btnDeleteLast.Enabled = true;
+            { 
+            mf.bnd.isOkToAddPoints = false;
```
J005t67, line 163: No need to use an if and else branch to assign either true or false. You can directly assign mf.bnd.isOkToAddPoints or !mf.bnd.isOkToAddPoints

SourceCode/GPS/Forms/Position.designer.cs
```diff
@@ -1456,116 +1456,54 @@ public void AddBoundaryPoint()
...
-                            }
-
-                            //draw on left side
-                            else
-                            {
-                                //Right side
-                                vec3 point = new vec3(section[0].leftPoint.easting, section[0].leftPoint.northing, 0);
-                                bnd.bndBeingMadePts.Add(point);
-                            }
+                         //Right side
```
J005t67, line 1478: Is the comment correct? Line 1475 mentions left side.

SourceCode/GPS/Forms/Position.designer.cs
```diff
@@ -1456,116 +1456,54 @@ public void AddBoundaryPoint()
...
+                         vec3 point = new vec3(
```
J005t67, line 1479: I need to look 10 times at the code to spot the subtle difference between the code fragment on line 1480-1483 and the fragment 1468-1472. It turns out that only the sign of the offset is different. This is error prone, hard to maintain and hard to debug. Please rewrite. Only calculate the offset with an condition
For example:
double offset = bnd.isDrawRightSide ? -bnd.isDrawRightSide : bnd.isDrawRightSide;
And only write the rest of the code once.

richardklasens:
Try the PR as build but has to rebased first to current dev.
The Sim does strange stuff when recording the boundary
it doesnt record the boundary when manual work in on or i'm doing it wrong.

PR #511, opened 2025-01-01
Remove unused code

J005t67:
Remove unused code.  Note that part of the removed code in class vec3 was harmfull because it used vec3 (easting, northing, heading) for something like (easting, northing, somethingElse)


lansalot (approved):
Looks good to me

PR #512, opened 2025-01-02
fix save settings in InvariantCulture

richardklasens:
did some testing with this fix, seems to solve the invariant stuff. loading defaut vehicle from config works, save as new vehicle, load another one, and load that new vehicle again works also. 
Full Reset doesnt work yet. if you do it comes back to last loaded vehicle

### DIFF
--- a/SourceCode/GPS/Properties/Settings.cs
+++ b/SourceCode/GPS/Properties/Settings.cs
@@ -1,12 +1,14 @@
 ﻿using Microsoft.Win32;
 using System;
+using System.Collections;
+using System.Collections.Generic;
 using System.Drawing;
 using System.Globalization;
 using System.IO;
 using System.Reflection;
 using System.Text;
+using System.Windows.Forms;
 using System.Xml;
-using System.Xml.Serialization;
 
 namespace AgOpenGPS.Properties
 {
@@ -376,15 +378,14 @@ namespace AgOpenGPS.Properties
         public bool setDisplay_isSectionLinesOn = true;
         public bool setDisplay_isLineSmooth = false;
 
-        public bool Load()
+        public DialogResult Load()
         {
             string path = Path.Combine(RegistrySettings.WorkingDirectory, "AgOpenGPS", "Vehicles", RegistrySettings.VehicleFileName + ".XML");
-            if (!LoadXMLFile(path, this))
-            {
+            var result = LoadXMLFile(path, this);
+            if (result != DialogResult.OK)
                 RegistrySettings.Save("VehicleFileName", "Default Vehicle");
-                return false;
-            }
-            return true;
+
+            return result;
         }
 
         public void Save()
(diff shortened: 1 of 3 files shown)


SourceCode/GPS/Properties/Settings.cs
```diff
@@ -408,120 +410,64 @@ public static bool LoadXMLFile(string filePath, object obj)
...
                                                     {
-                                                        Errors = true;
-                                                        continue;
-                                                        throw new ArgumentException("type not found");
+                                                        string value = reader.ReadString();
+                                                        Errors = Errors || SetValue(obj, pinfo, value);
                                                     }
                                                 }
-                                                catch (Exception)
+                                                catch (Exception e)
```
FantasyTeddy, line 470: The exception object `e` is not actually used here and can be removed.

SourceCode/GPS/Properties/Settings.cs
```diff
@@ -544,13 +486,60 @@ public static bool LoadXMLFile(string filePath, object obj)
                     return !Errors;
                 }
             }
-            catch (Exception)
+            catch (Exception e)
```
FantasyTeddy, line 489: Same here

SourceCode/GPS/Properties/Settings.cs
```diff
@@ -544,13 +486,60 @@ public static bool LoadXMLFile(string filePath, object obj)
...
+                if (parts.Length == 3 && int.TryParse(parts[0], NumberStyles.Any, CultureInfo.InvariantCulture, out int r) && int.TryParse(parts[1], out int g) && int.TryParse(parts[2], NumberStyles.Any, CultureInfo.InvariantCulture, out int b))
+                {
+                    pinfo.SetValue(obj, Color.FromArgb(r, g, b));
+                }
+            }
+            else
+            {
+                return true;
+            }
+            return false;
```
FantasyTeddy, line 540: This confused me for a bit: it is quite unusual for a method to return `true` if something went wrong. Maybe invert this and change the calling code to `!SetValue()`?

SourceCode/GPS/Properties/Settings.cs
```diff
@@ -635,17 +626,43 @@ public static void SaveXMLFile(string filePath, object obj)
                 if (File.Exists(filePath + ".tmp"))
                     File.Move(filePath + ".tmp", filePath);
             }
-            catch (Exception)
+            catch (Exception e)
```
FantasyTeddy, line 629: Same here

SourceCode/GPS/Properties/Settings.cs
```diff
@@ -635,17 +626,43 @@ public static void SaveXMLFile(string filePath, object obj)
...
+                }
+                else if (value is Color dd)
+                {
+                    xml.WriteElementString(fld.Name, $"{dd.R.ToString(CultureInfo.InvariantCulture)}, {dd.G.ToString(CultureInfo.InvariantCulture)}, {dd.B.ToString(CultureInfo.InvariantCulture)}");
+                }
+                else
+                {
+                    string stringValue = Convert.ToString(value, CultureInfo.InvariantCulture);
+                    xml.WriteElementString(fld.Name, stringValue);
+                }
```
FantasyTeddy, line 662: This code looks exactly the same as in the `SaveXMLFile` method. Why do we need this twice?

SourceCode/GPS/Properties/Settings.cs
J005t67: Or return an integer that holds the number of bugs found. The boolean var Errors in the calling code could be replaced by int nErrors and keep track of the number of bugs.

FantasyTeddy: I'm not sure this would help. Even if we would display this to the end-user, I don't think that a message like "5 errors found in the file" will be useful, if there is no more information about the types of errors.

richardklasens: If you really want to make a difference it should give your the opportunity to repair it. Showing a table with all the loaded values and highlight the ones that are wrong.

richardklasens: If you really want to make a difference it should give your the opportunity to repair it. Showing a table with all the loaded values and highlight the ones that are wrong.

FantasyTeddy: Yes, but I think this is infeasible (and certainly out of scope for this PR).

SourceCode/GPS/Properties/Settings.cs
```diff
@@ -376,15 +378,14 @@ public static Settings Default
         public bool setDisplay_isSectionLinesOn = true;
         public bool setDisplay_isLineSmooth = false;
 
-        public bool Load()
+        public DialogResult Load()
```
FantasyTeddy, line 381: This is a smart idea, but there is a drawback: `DialogResult` belongs to WinForms and is usually reserved for results of dialogues (hence the name).
The `Load` method belongs to our backend-logic and should not rely on WinForms.
Maybe you could add a custom `enum` that defines the return states?

FantasyTeddy:
I'm sorry, I don't know how to say this, but the longer I look at this entire solution, the more worried I am about its maintainability.
You are essentially writing your own XML serializer/deserializer and I think this is just something we should not be doing ourselves.
Can we not somehow use the underlying functionality of the `.settings` stuff and use that instead of writing all this code?

PR #514, opened 2025-01-02
Use NuGet packages for dependencies

FantasyTeddy:
Copy of #432

---

- Migrate to `PackageReference` (see [Migrate from packages.config to PackageReference](https://learn.microsoft.com/en-us/nuget/consume-packages/migrate-packages-config-to-package-reference) for a list of benefits)
- Replace DLL references with their matching NuGet package where possible
- Reference the `Keypad` project directly


richardklasens (dismissed):

P00qwerty (approved):
Looks good

PR #520, opened 2025-01-03
InvariantCulture now its open for review

P00qwerty:
works only in AgOpenGPS for now(can just copy paste to agio),
but was thinking the loading/saving of the xml part should move to the AgLibrary?

### DIFF
--- a/SourceCode/GPS/Forms/FormGPS.cs
+++ b/SourceCode/GPS/Forms/FormGPS.cs
@@ -283,15 +283,18 @@ namespace AgOpenGPS
 
             baseDirectory = Path.Combine(RegistrySettings.WorkingDirectory, "AgOpenGPS");
 
-            if (!Settings.Default.Load())
+            var result = Settings.Default.Load();
+            if (result != LoadResult.Ok)
             {
-                Log.System.Write("Error loading settings XML: " + RegistrySettings.VehicleFileName);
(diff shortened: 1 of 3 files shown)


FantasyTeddy:
As stated in the previous PR (#512), I'm a little bit concerned about the fact that you are essentially writing your own XML serializer/deserializer here.
Of course, I'm not just talking about this PR, since this just fixes some shortcomings of the original, but I'm talking about the entire change introduced to how settings are being handled.

I will give my approval for this PR, but I think we should have another discussion, if the maintenance of all this code is really worth it.

FantasyTeddy (approved):

P00qwerty:
It's using the default serializer for classes/lists inside the settings

FantasyTeddy:
Yes, this is nicely generic, but you also have some very specific special-cases: you treat the three structs `Color`, `Point` & `Size` explicitly, which makes sense because they are actually currently being used. But what if I want to add another struct setting? Do I have to extend the serialization/deserialization?

FantasyTeddy:
I just had another look and maybe it isn't as bad as I thought by just looking at the diff.

Your suggestion to put the XML part into AgLibrary would certainly help to tidy things up and clearly separate the "settings loading/storing" part from the actual (application specific) settings.

PR #531, opened 2025-01-05
Fixed a few texts / labels in the Hungarian translation

gunicsba:
Nothing big


richardklasens:
i cannot pull this into the upcoming release. It will be okay for the future release.

PR #532, opened 2025-01-05
Soft delete / ignore corrupted settings files

gunicsba:
Instead of forcefully deleting the settings file when it has any kind of errors, this fix allow us to retain it but in the form it won't be visible.


richardklasens:
in the next release, the file won't be deleted

PR #534, opened 2025-01-06
Combined pr

P00qwerty:


FantasyTeddy:
@farmerbriantee just added parts of this change (the `RegistrySettings`) to `release/6.6`, which got merged back into `develop`.
Sorry for the additional work with fixing the conflicts :(

What do you think, should we introduce the change to AgIO as well in this PR? I'm re-adding the `AgLibrary` project in #542, such that we would not have to duplicate the XML code.

P00qwerty:
I can add it to AgLibrary after you do #542

FantasyTeddy:
Another small request: can we focus this PR on the settings stuff and move the unrelated changes to another PR?

richardklasens:
> I can add it to AgLibrary after you do #542

Where is #542?

PR #558, opened 2025-01-09
Nudge by machine: added PGN222 to UDPComm.designer

mtz8302:
Nudge by machine: added PGN222 to UDPComm.designer

### DIFF
--- a/SourceCode/GPS/Forms/UDPComm.Designer.cs
+++ b/SourceCode/GPS/Forms/UDPComm.Designer.cs
@@ -270,6 +270,19 @@ namespace AgOpenGPS
                             }
                             break;
                         }
+                    case 222: // DE
+                        {
+                            //{ 0x80, 0x81, 0x7f, 222, number bytes, mask, command CRC };
+                            if (data.Length < 6) break;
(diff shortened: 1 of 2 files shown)


richardklasens (commented):
There is only a strange comment with // DE.

everybting else seems fine

FantasyTeddy:
> There is only a strange comment with // DE.

I think that's the hexadecimal notation of 222. Maybe we could prefix this with "0x" to prevent confusion?

SourceCode/GPS/Forms/UDPComm.Designer.cs
```diff
@@ -270,6 +270,19 @@ private void ReceiveFromAgIO(byte[] data)
                             }
                             break;
                         }
+                    case 222: // DE
```
FantasyTeddy, line 273: I think the PGN documentation should also be extended with this. (The one above "221" is also not documented...)

FantasyTeddy (approved):

SourceCode/GPS/Forms/UDPComm.Designer.cs
richardklasens: ok in that case, good to pull in dev then.

richardklasens:
Added and ready for release 6.7.x and up

PR #567, opened 2025-01-10
Convert projects to SDK-style

FantasyTeddy:
See original PR #426


SourceCode/GPS_Out/Source/GPS_Out.csproj
```diff
@@ -1,124 +1,13 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
-    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-    <ProjectGuid>{A9BF85B8-2178-479E-B41B-4E7F84D56A79}</ProjectGuid>
+    <TargetFramework>net472</TargetFramework>
```
J005t67, line 3: GPS_Out.csproj is the only proj that targets net472. Is thi sintentional?

FantasyTeddy: I only converted the projects without changing the target frameworks. But I agree that we could/should probably align them to all target `net48`.

richardklasens (dismissed):
Good to do this

richardklasens:
Maybe wait until Brian has dumped 6.6.1

richardklasens (commented):
if we do this now, and there will be a bugfix again, can we continue merging master to dev when this one is merged?

FantasyTeddy:
> if we do this now, and there will be a bugfix again, can we continue merging master to dev when this one is merged?

Yes this shouldn't be too hard. It's basically: take the version from "develop", because most changes don't affect the `*.csproj` anymore.

richardklasens (approved):
Well, in that case, threw it in!

PR #577, opened 2025-01-14
Six six patch

farmerbriantee:
Tramlines and Bing map updates/fixes


SourceCode/GPS/Forms/FormMap.cs
```diff
@@ -388,14 +495,14 @@ private void ResetMapGrid()
...
-                btnGray.Image = Properties.Resources.MapGray;
-            }
+            //if (isColorMap)
+            //{
+            //    btnGray.Image = Properties.Resources.MapColor;
+            //}
+            //else
+            //{
+            //    btnGray.Image = Properties.Resources.MapGray;
+            //}
```
FantasyTeddy, line 505: Is this code still useful?
Commented out code can be quite difficult to interpret in the future: is this meant to be removed? be brought back at some point? does it work? why was it commented out?

farmerbriantee: I'm pretty sure someone will complain there isn't grayscale, so may end up putting it back.

farmerbriantee: The whole program needs formatting and deleting unused code. Given there is a full refactor and most likely in WPF by the look of it, this is time spent now like re-arranging deck chairs on the Titanic.

farmerbriantee: Master is unlocked from Read Only, so it can be pulled

FantasyTeddy (approved):

PR #583, opened 2025-01-15
newSettings

P00qwerty:


FantasyTeddy:
My comments in the previous PR (#534) are still unresolved:
- The settings in AgIO also need to be changed to the new infrastructure (and the XML code should probably live in `AgLibrary`).
- It is quite hard to follow the changes regarding the settings, if a lot of minor other changes are also done. Can you split this up into several commits, such that it is easier to see what belongs together?

SourceCode/GPS/Properties/RegistrySettings.cs
```diff
@@ -114,6 +104,29 @@ public static void Reset()
 
         public static void CreateDirectories()
         {
+            try
+            {
+                if (workingDirectory == "Default")
+                {
+                    baseDirectory = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments), "AgOpenGPS");
```
J005t67, line 111: I see a lot of hardcoded strings like "AgOpenGPS" and "Default". Like this, a small typo easily results in a bug.
Could you replace the hard coded string with constants? In that case the compiler will spot the typos.

J005t67, line 111: Both branches of the if statement do Path.Combine(something, "AgOpenGPS")
Why not move this duplicate part outside (after) the if

P00qwerty: Not sure why you would change it to outside the if statement?
as then you need another string set that in the if statement and do the combine outside?

SourceCode/AgIO/Source/Properties/RegistrySettings.cs
```diff
@@ -0,0 +1,139 @@
...
+        {
+            try
+            {
+                //opening the subkey
+                RegistryKey regKey = Registry.CurrentUser.CreateSubKey(@"SOFTWARE\AgIO");
+
+                //Profile File Name from Registry Key
+                if (regKey.GetValue(RegKeys.profileName) == null)
+                {
+                    regKey.SetValue(RegKeys.profileName, "");
```
FantasyTeddy, line 36: The fact that the default profile name is now an empty string (`""`) is a breaking change compared to before where it was `"Default Profile"`. Therefore we will loose the ability to seamlessly migrate to this version, correct?

SourceCode/AgIO/Source/Properties/RegistrySettings.cs
```diff
@@ -0,0 +1,139 @@
...
+                profileName = regKey.GetValue(RegKeys.profileName).ToString();
+
+                /*
+                //Culture from Registry Key
+                if (regKey.GetValue("Language").ToString() == "")
+                {
+                    regKey.SetValue("Language", "en");
+                    Log.EventWriter("Registry -> Culture was null and Created");
+                }
+                culture = regKey.GetValue("Language").ToString();
```
FantasyTeddy, line 48: This still needs to be added, correct?

SourceCode/AgIO/Source/Properties/RegistrySettings.cs
```diff
@@ -0,0 +1,139 @@
...
+            Properties.Settings.Default.Load();
+        }
+
+        public static void Save(string name, string value)
+        {
+            try
+            {
+                RegistryKey key = Registry.CurrentUser.CreateSubKey(@"SOFTWARE\AgIO");
+                key.SetValue(name, value);
+                Log.EventWriter("Registry -> Key " + name + " Saved to registry key");
```
FantasyTeddy, line 73: Maybe we could also add the `value` that is being set to the log message?

SourceCode/AgIO/Source/Properties/RegistrySettings.cs
```diff
@@ -0,0 +1,139 @@
...
+                Registry.CurrentUser.DeleteSubKeyTree(@"SOFTWARE\AgIO");
+
+                Log.EventWriter("Registry -> Resetting Registry SubKey Tree and Full Default Reset");
+            }
+            catch (Exception ex)
+            {
+                Log.EventWriter("Registry -> Catch, Serious Problem Resetting Registry keys: " + ex.ToString());
+
+                Log.FileSaveSystemEvents();
+                Environment.Exit(0);
```
FantasyTeddy, line 99: Why do we need to shutdown AgIO in this case?

SourceCode/AgIO/Source/Properties/RegistrySettings.cs
```diff
@@ -0,0 +1,139 @@
...
+            }
+        }
+
+        public static void CreateDirectories()
```
FantasyTeddy, line 103: I think this can be `private`.

SourceCode/AgLibrary/XmlSettingsHandler.cs
FantasyTeddy, line 1: Can we move this file into a folder/namespace `AgLibrary.Settings`?

SourceCode/AgLibrary/XmlSettingsHandler.cs
```diff
@@ -0,0 +1,259 @@
...
+                    string name = "";
+                    while (reader.Read())
+                    {
+                        switch (reader.NodeType)
+                        {
+                            case XmlNodeType.Element:
+                                if (reader.Name == "setting")
+                                {
+                                    name = reader.GetAttribute("name");
+                                    string serializeAs = reader.GetAttribute("serializeAs");
```
FantasyTeddy, line 39: This variable is not used.

SourceCode/AgLibrary/XmlSettingsHandler.cs
```diff
@@ -0,0 +1,259 @@
...
+                                }
+                                else if (reader.Name == "value")
+                                {
+                                    if (!string.IsNullOrEmpty(name))
+                                    {
+                                        var pinfo = obj.GetType().GetField(name);
+                                        if (pinfo != null)
```
FantasyTeddy, line 46: Can we extract this (and the entire `if` block) into another method? I think this would improve readability.

SourceCode/AgLibrary/XmlSettingsHandler.cs
```diff
@@ -0,0 +1,259 @@
...
+                                                    // Deserialize XML into the custom object
+                                                    var serializer = new XmlSerializer(fieldType);
+                                                    var nestedObj = serializer.Deserialize(reader);
+                                                    pinfo.SetValue(obj, nestedObj);
+                                                }
+                                                else
+                                                {
+                                                    Errors = true;
+                                                    continue;
+                                                    throw new ArgumentException("type not found");
```
FantasyTeddy, line 131: This exception will never be thrown, because of the `continue` before.

SourceCode/AgLibrary/XmlSettingsHandler.cs
```diff
@@ -0,0 +1,259 @@
...
+                    }
+                    reader.Close();
+                }
+
+                return Errors ? LoadResult.Failed : LoadResult.Ok;
+            }
+            catch (Exception)
+            {
+            }
+            return LoadResult.Failed;
```
FantasyTeddy, line 156: I would rearrange the return statements to return the success case last:
```suggestion
            }
            catch (Exception)
            {
                return LoadResult.Failed;
            }
            return Errors ? LoadResult.Failed : LoadResult.Ok;
```

SourceCode/AgLibrary/XmlSettingsHandler.cs
```diff
@@ -0,0 +1,259 @@
...
+                }
+
+                if (File.Exists(filePath))
+                    File.Delete(filePath);
+
+                if (File.Exists(filePath + ".tmp"))
+                    File.Move(filePath + ".tmp", filePath);
+            }
+            catch (Exception)
+            {
```
FantasyTeddy, line 255: I think we should at least have some kind of log in case something goes wrong.

SourceCode/GPS/Properties/RegistrySettings.cs
```diff
@@ -0,0 +1,193 @@
...
+                {
+                    regKey.SetValue(RegKeys.workingDirectory, defaultString);
+                    Log.EventWriter("Registry -> Key workingDirectory was null");
+                }
+                workingDirectory = regKey.GetValue(RegKeys.workingDirectory).ToString();
+
+                //Vehicle File Name Registry Key
+                if (regKey.GetValue(RegKeys.vehicleFileName) == null)
+                {
+                    regKey.SetValue(RegKeys.vehicleFileName, "");
```
FantasyTeddy, line 43: Here again, I think we loose backwards compatibility, because `""` is not `"Default Vehicle"`. On purpose?

SourceCode/GPS/Forms/FormGPS.cs
```diff
@@ -109,7 +109,7 @@ public partial class FormGPS : Form
         /// <summary>
         /// create the scene camera
         /// </summary>
-        public CCamera camera = new CCamera();
+        public CCamera camera;
```
FantasyTeddy, line 112: Why move this initialization into the constructor?

SourceCode/GPS/Forms/Settings/ConfigVehicle.Designer.cs
```diff
@@ -167,27 +150,15 @@ private void btnVehicleSave_Click(object sender, EventArgs e)
...
-                mf.vehicle = new CVehicle(mf);
-                mf.tool = new CTool(mf);
-
-                //reset AOG
-                mf.LoadSettings();
+                RegistrySettings.Save(RegKeys.vehicleFileName, newVehicleName);
+                Settings.Default.Save();
 
-                SectionFeetInchesTotalWidthLabelUpdate();
+                //same anyways so why reload???
```
FantasyTeddy, line 161: Is this comment still useful after the change is done?

SourceCode/AgIO/Source/Properties/RegistrySettings.cs
P00qwerty: the old one does a check if string is "" or null so only thing is that it creates a profile with the name "Default profile" in this version when you start the old one and dont create a profile then come back to this verion.

SourceCode/AgIO/Source/Properties/RegistrySettings.cs
P00qwerty: you cant change it anywhere. in my opinion we should use the AgOpenGPS registery key and use the folder and Language keys from AgOpenGPS.

SourceCode/AgIO/Source/Properties/RegistrySettings.cs
P00qwerty: added a messagebox to tell the user

SourceCode/AgLibrary/XmlSettingsHandler.cs
P00qwerty: used it to see if i catched all types, changed it so it only trows when debugger is attached

SourceCode/GPS/Forms/FormGPS.cs
P00qwerty: fisrt the loading was done inside AgOpenGPS load function but then the CCamera was already initialized

SourceCode/GPS/Properties/RegistrySettings.cs
P00qwerty: same as the other one

SourceCode/AgLibrary/XmlSettingsHandler.cs
P00qwerty: definitely

SourceCode/GPS/Forms/FormGPS.cs
P00qwerty: dont realy want to change it back because now its the same as all the other classes

richardklasens:
Is it ok if can try the PR and loading different xml files with different cultures on a brand new VM (so that is no possible history from older versions)?

SourceCode/AgIO/Source/Properties/RegistrySettings.cs
FantasyTeddy: Ah okay, so it just creates a superfluous "Default Profile", but it will essentially still be able to migrate.

SourceCode/AgIO/Source/Properties/RegistrySettings.cs
FantasyTeddy: I think @farmerbriantee is currently adding this in the new `PatchToMaster` branch.

SourceCode/AgIO/Source/Forms/FormKeyboard.cs
```diff
@@ -22,7 +23,14 @@ private void FormKeyboard_Load(object sender, EventArgs e)
...
 
-            this.Height = 500;
+            if (Thread.CurrentThread.CurrentCulture.Name == "fr")
+            {
+                this.Height = 575;
+            }
+            else
+            {
+                this.Height = 500;
+            }
```
FantasyTeddy, line 33: I would not do this change here. @farmerbriantee will be changing this in `PatchToMaster` anyway.

We need to focus our PRs more towards a single "thing", without doing too many things at once.
This helps reviewers to understand the context better and it also makes the intention of the PR clearer.

FantasyTeddy (commented):
There are still two open comments from me. Otherwise this looks good and I'm looking forward to (finally) merge this.

SourceCode/AgLibrary/XmlSettingsHandler.cs
FantasyTeddy: Thank you for moving the file. Can we also change the namespace?

FantasyTeddy (commented):
One last comment unresolved.

Please do not make changes inside a merge commit. This makes it very hard to distinguish the merge itself from actual changes made.

SourceCode/AgLibrary/XmlSettingsHandler.cs
P00qwerty: not sure why you would want that? as now we add 'using AgLibrary._____' everywhere?

FantasyTeddy: Yes this is exactly the intention. By having the `using` everywhere the class is being used, we can clearly see what the dependencies are.
It is also best practice to have the namespaces follow the folder structure.

FantasyTeddy (approved):

PR #585, opened 2025-01-15
Refactor/simplify pgn calling code

J005t67:
Simplify the code that uses PGN classes.

### DIFF
--- a/SourceCode/GPS/Forms/Position.designer.cs
+++ b/SourceCode/GPS/Forms/Position.designer.cs
@@ -860,6 +860,7 @@ namespace AgOpenGPS
             //if the whole path driving driving process is green
             if (recPath.isDrivingRecordedPath) recPath.UpdatePosition();
 
+            Pgn254AutoSteerData pgn254 = PgnAll.Pgn254;
             // If Drive button off - normal autosteer 
             if (!vehicle.isInFreeDriveMode)
             {
--- a/SourceCode/GPS/Forms/Position.designer.cs
+++ b/SourceCode/GPS/Forms/Position.designer.cs
@@ -1011,8 +1011,7 @@ namespace AgOpenGPS
             }
 
             //out serial to autosteer module  //indivdual classes load the distance and heading deltas
-            pgn254.FindRefactoringBugs(p_254.pgn);
-            SendPgnToLoop(p_254.pgn);
+            SendPgnToLoop(p_254.pgn, pgn254);
 
             //for average cross track error
             if (guidanceLineDistanceOff < 29000)
--- a/SourceCode/GPS/Forms/Position.designer.cs
+++ b/SourceCode/GPS/Forms/Position.designer.cs
@@ -866,6 +866,7 @@ namespace AgOpenGPS
                 //fill up0 the appropriate arrays with new values
                 p_254.pgn[p_254.speedHi] = unchecked((byte)((int)(Math.Abs(avgSpeed) * 10.0) >> 8));
                 p_254.pgn[p_254.speedLo] = unchecked((byte)((int)(Math.Abs(avgSpeed) * 10.0)));
+                pgn254.SetSpeedInKmh(avgSpeed);
                 //mc.machineControlData[mc.cnSpeed] = mc.autoSteerData[mc.sdSpeed];
 
                 //save distance for display
@@ -876,10 +877,14 @@ namespace AgOpenGPS
                     guidanceLineDistanceOff = 32020;
                     p_254.pgn[p_254.status] = 0;
                 }
-
                 else p_254.pgn[p_254.status] = 1;
+                pgn254.SetStatus(isBtnAutoSteerOn);
 
-                if (recPath.isDrivingRecordedPath || recPath.isFollowingDubinsToPath) p_254.pgn[p_254.status] = 1;
+                if (recPath.isDrivingRecordedPath || recPath.isFollowingDubinsToPath)
+                {
+                    p_254.pgn[p_254.status] = 1;
+                    pgn254.SetStatus(true);
+                }
 
                 //mc.autoSteerData[7] = unchecked((byte)(guidanceLineDistanceOff >> 8));
                 //mc.autoSteerData[8] = unchecked((byte)(guidanceLineDistanceOff));
@@ -899,6 +904,7 @@ namespace AgOpenGPS
                 }
 
                 p_254.pgn[p_254.lineDistance] = unchecked((byte)distanceX2);
+                pgn254.SetDist(guidanceLineDistanceOff == 32020 || guidanceLineDistanceOff == 32000, guidanceLineDistanceOff);
 
                 if (!timerSim.Enabled)
                 {
@@ -950,13 +956,17 @@ namespace AgOpenGPS
                 //setAngVel = glm.toDegrees(setAngVel);
 
                 if (isChangingDirection && ahrs.imuHeading == 99999)
-                    p_254.pgn[p_254.status] = 0;
+                {
+                    pgn254.SetStatus(false);
+
+                }
 
                 //for now if backing up, turn off autosteer
                 if (!isSteerInReverse)
                 {
                     if (isReverse) p_254.pgn[p_254.status] = 0;
-                }                
+                    if (isReverse) pgn254.SetStatus(false);
+                }
 
                 // delay on dead zone.
                 if (p_254.pgn[p_254.status] == 1 && !isReverse
@@ -977,6 +987,7 @@ namespace AgOpenGPS
                 {
                     p_254.pgn[p_254.steerAngleHi] = unchecked((byte)(guidanceLineSteerAngle >> 8));
                     p_254.pgn[p_254.steerAngleLo] = unchecked((byte)(guidanceLineSteerAngle));
+                    pgn254.SetGuidanceLineSteerAngle(guidanceLineSteerAngle);
                 }
             }
 
@@ -998,7 +1009,8 @@ namespace AgOpenGPS
 
             }
 
-            //out serial to autosteer module  //indivdual classes load the distance and heading deltas 
+            //out serial to autosteer module  //indivdual classes load the distance and heading deltas
+            pgn254.AssertEqual(p_254);
             SendPgnToLoop(p_254.pgn);
 
             //for average cross track error
--- a/SourceCode/GPS/Forms/Position.designer.cs
+++ b/SourceCode/GPS/Forms/Position.designer.cs
@@ -1010,7 +1010,7 @@ namespace AgOpenGPS
             }
 
             //out serial to autosteer module  //indivdual classes load the distance and heading deltas
-            pgn254.AssertEqual(p_254);
+            pgn254.FindRefactoringBugs(p_254.pgn);
             SendPgnToLoop(p_254.pgn);
 
             //for average cross track error
--- a/SourceCode/GPS/Forms/Sections.Designer.cs
+++ b/SourceCode/GPS/Forms/Sections.Designer.cs
@@ -526,7 +526,7 @@ namespace AgOpenGPS
                         sectionControl1to16 += ((UInt16)(1 << j));
                     }
                 }
-                pgn254.SetSectionControl1to16(sectionControl1to16);
+                PgnAll.Pgn254.SetSectionControl1to16(sectionControl1to16);
 
 
                 //machine pgn
--- a/SourceCode/GPS/Forms/Sections.Designer.cs
+++ b/SourceCode/GPS/Forms/Sections.Designer.cs
@@ -517,6 +517,18 @@ namespace AgOpenGPS
                 }
                 p_254.pgn[p_254.sc9to16] = unchecked((byte)number);
 
+
+                UInt16 sectionControl1to16 = 0;
+                for (int j = 0; j < 16; j++)
+                {
+                    if (section[j].isSectionOn)
+                    {
+                        sectionControl1to16 += ((UInt16)(1 << j));
+                    }
+                }
+                pgn254.SetSectionControl1to16(sectionControl1to16);
+
+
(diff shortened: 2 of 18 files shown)


SourceCode/GPS/Forms/Sections.Designer.cs
```diff
@@ -517,6 +517,18 @@ private void BuildMachineByte()
                 }
                 p_254.pgn[p_254.sc9to16] = unchecked((byte)number);
 
+
+                UInt16 sectionControl1to16 = 0;
```
FantasyTeddy, line 521: What is the guideline for using `ushort` vs. `UInt16`? I get that it might be easier to understand that `UInt16` refers to a "16-bit unsigned int", but Visual Studio recommends to use `ushort` instead.

SourceCode/GPS/Pgn/Pgn252AutoSteerSettings.cs
```diff
@@ -0,0 +1,67 @@
...
+        private const int _countsPerDegreeIndex = 9;
+        private const int _wasOffsetLoIndex = 10;
+        private const int _wasOffsetHiIndex = 11;
+        private const int _ackermanIndex = 12;
+
+        public Pgn252AutoSteerSettings(IPgnErrorPresenter errorPresenter) : base(252, 8, errorPresenter)
+        {
+        }
+
+        public byte SetProportionalGain(int proportionalGain)
```
FantasyTeddy, line 19: It is very confusing that a `Set...` method has a return value... maybe we need a better name for these kinds of methods?

SourceCode/GPS/Pgn/WinFormsPgnErrorPresenter.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+
+namespace AgOpenGPS
+{
+    public class WinFormsPgnErrorPresenter : IPgnErrorPresenter
+    {
+        private void TimedMessageBox(int timeout, string title, string message)
+        {
+            var form = new FormTimedMessage(timeout, title, message);
```
FantasyTeddy, line 9: If we are using this class to find bugs in the new PGN implementation, shouldn't we be more "drastic" about reporting an error than just displaying a timed message box? Why not just throw an exception and crash the application? That way you would immediately notice, that something is wrong.

SourceCode/GPS/Pgn/PgnBase.cs
```diff
@@ -0,0 +1,146 @@
...
+        private readonly IPgnErrorPresenter _errorPresenter;
+        protected readonly byte[] _message;
+
+        public PgnBase(byte id, byte contentSize, IPgnErrorPresenter errorPresenter)
+        {
+            _errorPresenter = errorPresenter;
+            _message = new byte[5 + contentSize + 1];
+            _message[0] = 0x80;
+            _message[1] = 0x81;
+            _message[2] = 0x7f;
```
FantasyTeddy, line 16: Is the source always `0x7f`? What if we want to use the same mechanic in AgIO?

SourceCode/GPS/Pgn/PgnBase.cs
```diff
@@ -0,0 +1,146 @@
+﻿﻿using System;
+
+namespace AgOpenGPS
+{
+    public class PgnBase
+    {
+        private readonly IPgnErrorPresenter _errorPresenter;
+        protected readonly byte[] _message;
+
+        public PgnBase(byte id, byte contentSize, IPgnErrorPresenter errorPresenter)
```
FantasyTeddy, line 10: If we need to know the `contentSize` at the time of creation of the PGN, we might have a problem when we want to encode variable-size PGNs (like 221 which is used to display a text message).

SourceCode/GPS/Pgn/PgnBase.cs
```diff
@@ -0,0 +1,146 @@
...
+            _message = new byte[5 + contentSize + 1];
+            _message[0] = 0x80;
+            _message[1] = 0x81;
+            _message[2] = 0x7f;
+            _message[3] = id;
+            _message[4] = contentSize;
+            for (int i = 5; i < 5 + contentSize; i++)
+            {
+                _message[i] = 0;
+            }
```
FantasyTeddy, line 22: Is it really necessary to set all values to `0`? Isn't this the default for empty arrays?

SourceCode/GPS/Pgn/PgnBase.cs
```diff
@@ -0,0 +1,146 @@
...
+            {
+                ComputeCrc();
+                return _message;
+            }
+        }
+
+        public void FindRefactoringBugs(byte[] oldMessage)
+        {
+            ComputeCrc();
+            for (int i = 0; i < _message.Length; i++)
```
FantasyTeddy, line 40: Maybe a first check should be if the length of `_message` and `oldMessage` match.

SourceCode/GPS/Pgn/PgnBase.cs
```diff
@@ -0,0 +1,146 @@
...
+            _message[loIndex + 1] = unchecked((byte)(int16Value >> 8));
+        }
+
+        protected void SetUInt16LoHi(byte loIndex, UInt16 uint16Value)
+        {
+            _message[loIndex] = (byte)(uint16Value & byte.MaxValue);
+            _message[loIndex + 1] = (byte)(uint16Value >> 8);
+        }
+
+        protected byte SetInt(byte index, int inputValue)
```
FantasyTeddy, line 82: Another way of dealing with the overflow problem could be to just use `byte` throughout the application if it eventually needs to be put into a PGN. That way the compiler could help us in finding the problematic places where we need to be careful when manipulating it.

SourceCode/GPS/Pgn/PgnBase.cs
```diff
@@ -0,0 +1,146 @@
...
+            if (inputValue < byte.MinValue)
+            {
+                OverflowErrorIntToByte(index, inputValue, b);
+                b = byte.MinValue;
+            }
+            _message[index] = b;
+            return b;
+        }
+
+        public void SetDoubleLoHi(byte loIndex, double inputValue)
```
FantasyTeddy, line 99: This is another place where I find it quite confusing that I can set a `double` which is essentially being truncated to an `int`... do we really need this functionality?

SourceCode/GPS/Pgn/Pgn235SectionDimensions.cs
```diff
@@ -0,0 +1,31 @@
...
+        }
+
+        public void SetSectionWidth(int section, double sectionWidthInMeters)
+        {
+            Debug.Assert(0 <= section);
+            Debug.Assert(section < _maxSections);
+
+            int sectionLoIndex = _section0LoIndex + 2 * section;
+            SetDoubleLoHi((byte)sectionLoIndex, 100.0 * sectionWidthInMeters);
+        }
```
FantasyTeddy, line 24: Functions like this have multiple jobs:
- Convert the input parameters into the proper units that will be transmitted (`100.0 * sectionWidthInMeters`)
- Set the underlying message format to the correct values (the actual PGN format)

I'm not sure if it is the correct way to do both in the PGN representation. The second one is the actual PGN logic, the first one is additional (business)-logic that maybe better placed somewhere else.

SourceCode/GPS/Pgn/Pgn236RelayConfig.cs
```diff
@@ -0,0 +1,27 @@
...
+        private const int _pin0Index = 5;
+
+        public Pgn236RelayConfig(IPgnErrorPresenter errorPresenter) : base(236, _contentSize, errorPresenter)
+        {
+        }
+
+        public void SetPin(int pinNumber, string word)
+        {
+            Debug.Assert(0 <= pinNumber);
+            Debug.Assert(pinNumber < _contentSize);
```
FantasyTeddy, line 18: Important note about `Debug.Assert`: if this is used in a production environment (built using the `Release` configuration), it will do nothing. This might lead to some strange issues if the values are actually out of range.

SourceCode/GPS/Pgn/Pgn252AutoSteerSettings.cs
```diff
@@ -0,0 +1,67 @@
...
+            return SetInt(_ackermanIndex, ackerman);
+        }
+
+        // Wheel Angle Sensor
+        public void SetWasOffset(int wasOffset)
+        {
+            SetIntLoHi(_wasOffsetLoIndex, wasOffset);
+        }
+
+        public void SetFromDefaultSettings()
```
FantasyTeddy, line 55: This is also something I would not expect the PGN logic to handle. Reading from settings should be done elsewhere.

SourceCode/GPS/Pgn/PgnAll.cs
```diff
@@ -0,0 +1,25 @@
+﻿using AgOpenGPS.Pgn;
+
+namespace AgOpenGPS
+{
+    public class PgnAll
```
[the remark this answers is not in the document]
J005t67: What I did was completely in line with other AgOpenGps code. The lifetime of almost every object equals the lifetime of the application. ;-) But, I agree. I will remove PgnAll

FantasyTeddy:
Something else to consider: this solution only provides a way to do "serialization", i.e. converting values into PGNs (and then `byte[]` arrays). What about the "deserialization" of PGNs, where a `byte[]` is the input and a PGN should be returned?

SourceCode/GPS/Pgn/Pgn235SectionDimensions.cs
J005t67: The multiplication by 100 is not business logic. It is Pgn's way to squeze a double in two bytes without loosing too much accuracy.

SourceCode/GPS/Pgn/Pgn236RelayConfig.cs
J005t67: Yes, I know, but it helps to make the intentions clear. (More accurately than a comment.)
And obviously it will helps to catch bugs in debug mode during development.
To me, this code is according the intended use of Debug.Asserts

SourceCode/GPS/Pgn/PgnBase.cs
J005t67: Yes, we need it and it is the responsibility of the Pgn classes

SourceCode/GPS/Pgn/PgnBase.cs
J005t67: No, I don't want byte variables all over the place. That would be very annoying during all kind of calculations. The problem that the values need to be squeezed in one ore more bytes, is a problem of Pgn. We should not polute the rest of the application with this problem.

SourceCode/GPS/Forms/Sections.Designer.cs
J005t67: I'm more familiar with C++. In C++, the size of a short and int is not always 16 and 32 bytes. For my understanding I checked it for C#, and indeed the sizes are always 16 and 32. Nevertheless I prefer UInt16 in the exceptional case of Pgn classes.

SourceCode/GPS/Forms/Position.designer.cs
```diff
@@ -950,13 +957,17 @@ public void UpdateFixPosition()
                 //setAngVel = glm.toDegrees(setAngVel);
 
                 if (isChangingDirection && ahrs.imuHeading == 99999)
-                    p_254.pgn[p_254.status] = 0;
```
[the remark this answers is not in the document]
J005t67: Certainly, thanks! We must keep it untill all old pgn code will be removed.

SourceCode/GPS/Pgn/PgnBase.cs
FantasyTeddy: I disagree. If I can set a `double` in a PGN, I expect it to be converted into a binary representation that properly represents it (including all of its digits after the decimal point). This function truncates the value into an integer and discards all the digits after the decimal point, which is most likely not what I expect when calling this method.

SourceCode/GPS/Pgn/Pgn252AutoSteerSettings.cs
J005t67: Yes, confusing indeed. I will change it to regular properties with get and set methods.

SourceCode/GPS/Pgn/Pgn254AutoSteerData.cs
```diff
@@ -0,0 +1,74 @@
...
+                int distanceX2 = (int)(guidanceLineDistanceOffInCm * 0.05);
+
+                distanceX2 = Math.Max(distanceX2, -127);
+                distanceX2 = Math.Min(distanceX2, 127);
+                distanceX2 += 127;
+                _message[_lineDistanceIndex] = unchecked((byte)distanceX2);
+            }
+        }
+
+        public void SetSpeedInKmh(double speed)
```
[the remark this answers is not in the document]
J005t67: Yes, changed.

SourceCode/GPS/Pgn/PgnBase.cs
J005t67: Oops, that method must not be public. Changed it to protected. 

It is used by methods like Pgn235SectionDimensions.SetSectionWidth that do proper scaling before the double is truncated in PgnBase.

FantasyTeddy: Even if it is only used by subclasses, I still think that this is not ideal. If the subclass needs to store the integer value of a `double` it should perform the truncation itself and call something like `SetInt`.

SourceCode/GPS/Pgn/PgnAll.cs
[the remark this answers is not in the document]
FantasyTeddy: Why is it back?

PR #593, opened 2025-01-18
Extract Keypad & Keyboard display into extension methods

FantasyTeddy:
This removes the dependency to the main forms (`FormLoop` & `FormGPS`).


J005t67 (approved):
It is always good in my opinion, to move code (that does not belong there) outside the designer files, so I will aprrove. But the new extension methods are duplicated, and I would prefer to move them to a dll that can be shared by both WinFormApps later.

FantasyTeddy:
> [...] But the new extension methods are duplicated, and I would prefer to move them to a dll that can be shared by both WinFormApps later.

I totally agree. What do you think, is the `AgLibrary` project suitable for that or should we have a different WinForms specific project to decouple backend from frontend logic?

PR #600, opened 2025-01-20
Add new `AgOpenGPS.Core` project

FantasyTeddy:
The new `AgOpenGPS.Core` project will contain any future code that is specific to AgOpenGPS (not AgIO) but does not depend on WinForms.


J005t67 (approved):
Thank you for creating this SDK-style project for me Matthias. It has a nice descriptive name. It is in the right folder. You even added the correct dependency already.   ===> Approve

PR #602, opened 2025-01-20
Take over refactored base classes from AOG2 repository

J005t67:
We need a group of more versatile base classes, like GeoPath and GeoPolygon, GeoCoord, GeoDir, GeoDelta
(The currently used List<vec2> or List<vec3> are much to general and don't do a good job.)

The new classes serve multiple purposes:
- provide logical places to implement methods like Distance and IsPointInsidePoygon, 
- GeoPath is very helpfull in the context of RecordedPath, Ab-line, ReferenceLine etc
- GeoPolygon is very helpfull in the context of Boudaries, HeadLand, etc

The base class GeoCoord, GeoDelta and GeoDirection allow us to write a lot of computations efficiently without bothering about easting and northing components individually and the corresponding sines and cosines.

Note that I prefixed all classes with 'Geo' for reasons'
- it helps to avoid name classes with already existing classes.
- all classes are use Northing and Easting. it will help to distinguish them from other classes that use X and Y components.

Also note that the classes are already use intensively in the AOG2 repository.

The next series of pull requests will transfer refactored classes for Flags, Boundaries, TramLines, RecordedPath and corresponding view models.


SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoBoundingBox.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using System.Diagnostics;
+
+namespace AgOpenGPS.Core
```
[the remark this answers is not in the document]
FantasyTeddy: I'm not too sure that I like this decision.
Following the 1 folder = 1 namespace rule strictly may not always be the best choice, but the other extreme of having only 1 namespace for the entire project seems to be not ideal as well.
Namespaces exist for several reasons:
- Prevent naming collisions
- Grouping/organizing related code
- Searching files can be easier

I tend to think that it also helps in identifying problematic areas: if there are files/classes that have a lot of different `using` statements, the class most likely has too many responsibilities and should be refactored.

There is also the argument, that [IDE0130](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0130) exists for a reason...

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoBoundingBox.cs
```diff
@@ -0,0 +1,47 @@
...
+{
+    public struct GeoBoundingBox
```
FantasyTeddy, line 6: I have never had a very good grasp of `class` vs. `struct`. I know the technical differences, but I could never really make out good reasons to use a `struct`. Can you explain why you chose `struct` here?

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoCoord.cs
```diff
@@ -0,0 +1,52 @@
...
+        private double _easting;
+
+        public GeoCoord(double northing, double easting)
+        {
+            _northing = northing;
+            _easting = easting;
+        }
+
+        public double Northing => _northing;
+        public double Easting => _easting;
```
FantasyTeddy, line 17: Maybe use auto properties here:
```suggestion
        public double Northing { get; }
        public double Easting { get; }
```

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoDelta.cs
```diff
@@ -0,0 +1,28 @@
...
+        }
+
+        public GeoDelta(double northingDelta, double eastingDelta)
+        {
+            _northingDelta = northingDelta;
+            _eastingDelta = eastingDelta;
+        }
+
+        public double NorthingDelta => _northingDelta;
+        public double EastingDelta => _eastingDelta;
```
FantasyTeddy, line 23:
```suggestion
        public double NorthingDelta { get; }
        public double EastingDelta { get; }
```

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoDelta.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    public class GeoDelta
+    {
+        private double _northingDelta;
+        private double _eastingDelta;
+
+        public GeoDelta(GeoCoord fromCoord, GeoCoord toCoord)
+        {
+            _northingDelta = toCoord.Northing - fromCoord.Northing;
+            _eastingDelta = toCoord.Easting - fromCoord.Easting;
```
FantasyTeddy, line 13: I think this calculation should be moved into the `GeoCoord.DistanceSquared` method. This is the only place that this will be calculated and it is where I would search for it instead of a constructor.

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoCoord.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        public double DistanceSquared(GeoCoord coord2)
+        {
+            return new GeoDelta(this, coord2).LengthSquared;
+        }
+
+        public double Distance(GeoCoord coord2)
+        {
+            return Math.Sqrt(DistanceSquared(coord2));
```
FantasyTeddy, line 26: Maybe something like this would be more obvious:
```suggestion
            return (coord2 - this).Length;
```
?

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoDir.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            _angle = angle;
+        }
+
+        public GeoDir(GeoDelta delta)
+        {
+            _angle = Math.Atan2(delta.EastingDelta, delta.NorthingDelta);
+        }
+
+        public double Angle => _angle;
```
FantasyTeddy, line 18:
```suggestion
        public double Angle { get; }
```

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoDir.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+    public struct GeoDir
+    {
+        private double _angle;
+        public GeoDir(double angle)
+        {
+            _angle = angle;
+        }
+
+        public GeoDir(GeoDelta delta)
```
FantasyTeddy, line 13: Is this constructor used anywhere? Because here I would argue as well, that the calculation should probably be done before constructing the `GeoDir`.

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoDir.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            _angle = Math.Atan2(delta.EastingDelta, delta.NorthingDelta);
+        }
+
+        public double Angle => _angle;
+        public double Cos => Math.Cos(_angle);
+        public double Sin => Math.Sin(_angle);
```
FantasyTeddy, line 20: Should this really be exposed like this? If there is a real-world meaning to the sine/cosine, I think the properties should reflect that, but just as the mathematical calculation it seems a bit superfluous.

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoDir.cs
```diff
@@ -0,0 +1,49 @@
...
+        public GeoDir PerpendicularLeft => new GeoDir(_angle - 0.5 * Math.PI);
+
+        public GeoDir PerpendicularRight => new GeoDir(_angle + 0.5 * Math.PI);
+        public GeoDir Inverted => new GeoDir(_angle + Math.PI);
+
+
+        public void Invert()
+        {
+            _angle += Math.PI;
+        }
```
FantasyTeddy, line 31: I really like the idea of these primitives being immutable. Does `GeoDir` really need to be inverted, instead of using a new instance from the `Inverted` property?

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoCoordDir.cs
```diff
@@ -0,0 +1,17 @@
...
+        private GeoDir _direction;
+
+        public GeoCoordDir(GeoCoord coord, GeoDir direction)
+        {
+            _direction = direction;
+            _coord = coord;
+        }
+
+        public GeoCoord Coord => _coord;
+        public GeoDir Direction => _direction;
```
FantasyTeddy, line 15:
```suggestion
        public GeoCoord Coord { get; }
        public GeoDir Direction { get; }
```

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoBoundingBox.cs
```diff
@@ -0,0 +1,47 @@
...
+    {
+        private GeoCoord _minCoord;
+        private GeoCoord _maxCoord;
+        private bool _isEmpty;
+
+        static public GeoBoundingBox CreateBoundingBox()
```
FantasyTeddy, line 12: Maybe use a more descriptive name. E.g.: `CreateEmpty` (and remove the `...BoundingBox` because that is obvious when being called like this: `GeoBoundingBox.CreateEmpty()`

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoBoundingBox.cs
```diff
@@ -0,0 +1,47 @@
...
+            _minCoord = minCoord;
+            _maxCoord = maxCoord;
+            _isEmpty = true;
+        }
+
+        public bool IsEmpty => _isEmpty;
+        public double MinNorthing => _minCoord.Northing;
+        public double MaxNorthing => _maxCoord.Northing;
+        public double MinEasting => _minCoord.Easting;
+        public double MaxEasting => _maxCoord.Easting;
```
FantasyTeddy, line 30: Why not expose the `_minCoord` & `_maxCoord` as properties?

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoBoundingBox.cs
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        public GeoBoundingBox(GeoCoord minCoord, GeoCoord maxCoord)
+        {
+            _minCoord = minCoord;
+            _maxCoord = maxCoord;
+            _isEmpty = true;
+        }
+
+        public bool IsEmpty => _isEmpty;
```
FantasyTeddy, line 26: I don't quite understand the use of `IsEmpty`... why is it `true` when creating a new bounding box and `false` when adding another coordinate?

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoPaths/GeoPathBase.cs
```diff
@@ -0,0 +1,27 @@
+﻿namespace AgOpenGPS.Core
+{
+    public abstract class GeoPathBase
```
FantasyTeddy, line 3: Can you quickly explain why the two inheritance graphs are useful:
- `GeoPath` : `GeoPathBase`
- `GeoPolygon` : `GeoPolygonBase` : `GeoPathBase`

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoPaths/GeoPathWithHeading.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            _headings = new List<GeoDir>();
+        }
+
+        public override GeoCoord this[int i] => throw new NotImplementedException();
+
+        public override void Add(GeoCoord coord)
+        {
+            throw new NotImplementedException();
+        }
```
FantasyTeddy, line 19: Why are these "not implemented"?

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/Wsg84.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace AgOpenGPS.Core
+{
+    public class Wgs84
```
FantasyTeddy, line 3: I think classes with acronyms like this should at least have a comment with the full name and maybe even some informative link (to Wikipedia for example).

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/Wsg84.cs
```diff
@@ -0,0 +1,17 @@
...
+        private double _longitude;
+
+        public Wgs84(double latitude, double longitude)
+        {
+            _latitude = latitude;
+            _longitude = longitude;
+        }
+
+        public double Latitude => _latitude;
+        public double Longitude => _longitude;
```
FantasyTeddy, line 15:
```suggestion
        public double Latitude { get; }
        public double Longitude { get; }
```

SourceCode/AgOpenGPS.Core/ApplicationModels/DebugModels/DebugAsserts.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System.Diagnostics;
+
+namespace AgOpenGPS.Core
+{
+    public class DebugAsserts
+    {
+        public static void AssertEqual(double a, double b, double epsilon = 0.0001)
```
FantasyTeddy, line 7: I think calling these methods looks a bit clunky: `DebugAsserts.AssertEqual`
Wouldn't it be easier to read if it were something like this: `DebugAssert.Equal`?

FantasyTeddy (commented):
Sorry for the many comments.
I love these additions and I want to make sure that we build on a solid foundation. Much easier to change things now (if needed) than later when these are used everywhere.

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoBoundingBox.cs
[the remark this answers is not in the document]
J005t67: I used to work in large code-bases (I try to avoid the word project here), with many projects. Unwanted dependencies, especially the ones that required new dependencies in the project files, always were closely examined in a review. Therefore, lengthy usings did not add much. I also think that my naming chocices do a good job, taking the three bullets into account. But anyway, the number of files in AgOpenGps.Core will keep growing, so just one namespace will not help. The folder structure that I have in mind will keep the various architecture layers appart on the top level. I propose to include one node more (the toplevel) in the name spaces and not go to a 1-to-1 mapping of folders and namespaces.
Do you agree?

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoCoord.cs
J005t67: Yes, why not.

SourceCode/AgOpenGPS.Core/ApplicationModels/DebugModels/DebugAsserts.cs
J005t67: It looks a bit clunky indeed, but it expresses quite well what it does. Just 'Equal' would make me think that it just returns a boolean.

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoBoundingBox.cs
J005t67: The general advise is to use struct for small and easy concepts and classes for bigger and larger concepts.
That is exactly what I ---thought-- I did.  I just checked, but I made a few mistakes.  I will change GeoCoord, GeoDelta, and Wsg84 to a struct. For the GeoBoundingBox in particular: I think it fits the general rule

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoPaths/GeoPathBase.cs
J005t67: Mmm, this needs improvement. I introduced the abstract class GeoPathBase because I do not like the idea that every polygon and path 'carries a bucket of coordinates'. I would like to use more mathimatical description for paths and paths segments ,if that is more appropriate  (For example the fragments that make up a DubinsPath are circle segments and straight lines.) The base class GeoPath was meant to be base class for the mathimatical paths too, but I will remove it now, to simplify the code.

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoPaths/GeoPathWithHeading.cs
J005t67: If you only 'Add' a GeoCoord, the corresponding 'Heading' is missing, and the GeoPathWithHeading is in a corrupt state. 
Will change it to InvalidOperationException. which is more appropriate.

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoPaths/GeoPathBase.cs
J005t67: Simplified it by removing GeoPolygonBase

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoBoundingBox.cs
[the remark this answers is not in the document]
FantasyTeddy: I currently work in a code-base that also consists of many projects and we follow the one-to-one rule strictly, without any noticeable drawback.

I think adding at least one level more in the namespace is a good compromise.

Speaking of: if this first level is the top-most folder (in this case `ApplicationModels`) it may be worth discussing about these names. If I understand correctly, the `...Models` suffix comes from the MVVM background? If so, I would not include that and simply go with `Application`, because the "Model" in MVVM is such a broad term that we would have to append it to almost anything backend-related.

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoBoundingBox.cs
FantasyTeddy: Okay, thank you for clarifying.

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoPaths/GeoPathWithHeading.cs
FantasyTeddy: I was just pointing this out, because in general, having to "remove" functions that are inherited, usually indicate that the abstractions are not fitting.
Maybe a `GeoPathWithHeading` should be a composition of a `GeoPath` and its heading, instead of an inheritance?

SourceCode/AgOpenGPS.Core/ApplicationModels/DebugModels/DebugAsserts.cs
FantasyTeddy: Okay, but isn't this given by the "Assert" in `DebugAssert` already? You cannot use the `Equal` method standalone, you always have to prefix it with the class name.
Would the method name `AreEqual` be more appropriate? (Like used in many testing frameworks for assertions)

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoBoundingBox.cs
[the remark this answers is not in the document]
J005t67: My plan is to use 'Models' suffix for the directories only not for the classes. The Folder ApplicationModels will contain folders like BaseModels, FieldModels, VehicleModels, GuidanceModels. And classes like RecordedPath, PathRecorder, DubinsPath, DubinsPathGenerator.  The folder ApplicationViewModels (not part of this review)  will contain, folders like FieldViewModels, VehicleViewModel, and GuidanceViewModels. I want to prevent that directory names like 'Field' or 'Vehicle' occur many times.

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoPaths/GeoPathWithHeading.cs
J005t67: Yes, this is a code smell and it does not feel right. I tried to come up with an alternative for the old List<vec3>, but this one is unnecessary complex. For now, I will just remove GeoPathWithHeading and GeoPolygonWithHeading from this review and solve it properly later in a context where I can test it. I lalso noticed in the code that many times the coords and the headings are computed in a different for loop. So forcing people to add them simulataniously will not work well.

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoBoundingBox.cs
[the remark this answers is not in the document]
FantasyTeddy: Why not just start with a base folder `Model` and then add subdirectories `Field`, `Vehicle`, etc.? If the folders are in different parent directories the usage (if the namespaces also align) read very nicely: `AgOpenGPS.Core.Model.Field` and `AgOpenGPS.Core.ViewModel.Field`.

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoDir.cs
J005t67: We do not really need it, because we can work around it., but is convenient in cases that the direction must be inverted conditionally.

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoDir.cs
J005t67: Removed

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoDir.cs
J005t67: I'm already using it 7 times in AOG2. It is also essential not to mix up the order of easting and northing to get the correct result.

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoDir.cs
J005t67: Changed to:
        public double Angle { get; private set; }

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoCoord.cs
J005t67: Changed to
            return new GeoDelta(this, coord2).Length;
to be more in line with method above.

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoBoundingBox.cs
J005t67: Yes, much better.

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoBoundingBox.cs
J005t67: Some functions will sometimes return empty boundingboxes and this is the most convenient way for the caller to check this. Otherwise the caller needs to know the internal details like the intial values double.MaxValue and double.MinValue

SourceCode/AgOpenGPS.Core/ApplicationModels/DebugModels/DebugAsserts.cs
J005t67: Changed it to AreEqual, but to be honest, I does not look any better.

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoBoundingBox.cs
FantasyTeddy: Okay, but doesn't the current implementation require that a `GeoBoundingBox` is initialized with `double.MaxValue` and `double.MinValue`? If any other values are passed to the constructor, `IsEmpty` will still be `true`, even if it isn't.

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoBoundingBox.cs
FantasyTeddy: You still have the suffix `...BoundingBox`...?

SourceCode/AgOpenGPS.Core/ApplicationModels/DebugModels/DebugAsserts.cs
FantasyTeddy: I think it does. But if you want to use `AssertEqual`, feel free to change it back.

SourceCode/AgOpenGPS.Core/ApplicationModels/DebugModels/DebugAsserts.cs
```diff
@@ -0,0 +1,49 @@
...
+        public static void AreEqual(GeoCoord a, GeoCoord b)
+        {
+            AreEqual(a.Northing, b.Northing);
+            AreEqual(a.Easting, b.Easting);
+        }
+
+        public static void AreEqual(GeoDir a, GeoDir b)
+        {
+            AreEqual(Math.Cos(a.Angle), Math.Cos(b.Angle));
+            AreEqual(Math.Sin(a.Angle), Math.Sin(b.Angle));
```
FantasyTeddy, line 28: Wouldn't it be sufficient to just assert `Angle`?

J005t67: No, the current code contains a lot of lines that keep angles in the range [-PI, PI] or [0, 2 * PI]. I don't want to do the same with  the angle in GeoDir (and make assumptions about  which range is best.) Keeping the angles within a certain range is only needed for displaying the values.

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoBoundingBox.cs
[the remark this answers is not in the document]
J005t67: I don't want to use the singular form of Models and ViewModels. it is confusing to me. Also I don't favor the folder names Field and Vehicle, just because they will appear more than once and also because the singular form is confusing to me. To me it seems that you clinch so much to the long namespaces that you want to sacrifice clear foldernames for it.

SourceCode/AgOpenGPS.Core/ApplicationModels/DebugModels/DebugAsserts.cs
FantasyTeddy: Ah, didn't consider that 👍

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoBoundingBox.cs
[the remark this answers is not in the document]
FantasyTeddy: I'm fine with using the plural form for Models/ViewModels, I was just thinking out loud.
Regarding the folder structure, I just think that it does not make sense to append a suffix everywhere, if the same could be achieved by simply having a folder with the same name. You never look at a folder in isolation, it always has a hierarchy it belongs to. The folder `GeoPaths` is also contained within `BaseModels` and you don't call it `GeoPathBaseModels`. Context matters.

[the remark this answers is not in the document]
J005t67: I changed the top level folder name from ApplicationModel to Models, and folder BaseModels to Base.
I did not rename DebugModels to Debug, because everything inside Debug folders is ignored in versioning.

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoBoundingBox.cs
J005t67: Indeed a bug. Corrected.

SourceCode/AgOpenGPS.Core/ApplicationModels/BaseModels/GeoBoundingBox.cs
J005t67: Changed

FantasyTeddy (approved):
Thank you for your patience with all my remarks. I think it paid off and we even found some potential bugs.

PR #603, opened 2025-01-20
Autoswitch Dual to Fix speed based

chrisriteco:
Driving with Dual ensures that you know your driving direction quickly (almost) from a standstill. Perfect for starting from a standstill, or for turning well into the planting beds on small plots. The disadvantage for us is that with certain soil cultivation and driving on loose ploughed land, the tractor starts driving crooked and becomes restless at high speed. The direction that the Dual transmits is then no longer the actual driving direction. Here is a proposal to combine the best of both worlds. AG to automatically switch from Dual to Fix based on speed. For a quickly available driving direction, start with Dual and when you are driving above 2 km/h, switch to Fix for the correct driving direction.

### DIFF
--- a/SourceCode/GPS/Classes/CAHRS.cs
+++ b/SourceCode/GPS/Classes/CAHRS.cs
@@ -16,7 +16,11 @@
         public double rollFilter;
 
         //is the auto steer in auto turn on mode or not
-        public bool isAutoSteerAuto, isRollInvert, autoswitchDualFix2FixOn, isDualAsIMU, isReverseOn;
+        public bool isAutoSteerAuto, isRollInvert, isDualAsIMU, isReverseOn;
+
+        // AutoswitchDualFix 
+        public bool autoSwitchDualFixOn;
+        public double autoSwitchDualFixSpeed;
 
         //the factor for fusion of GPS and IMU
         public double forwardComp, reverseComp, fusionWeight;
@@ -41,7 +45,9 @@
 
             isReverseOn = Properties.Settings.Default.setIMU_isReverseOn;
 
-            autoswitchDualFix2FixOn = Properties.Settings.Default.setAutoswitchDualFix2FixOn;
+            autoSwitchDualFixOn = Properties.Settings.Default.setAutoSwitchDualFixOn;
+
+            autoSwitchDualFixSpeed = Properties.Settings.Default.setAutoSwitchDualFixSpeed;
         }
     }
 }
--- a/SourceCode/GPS/Classes/CAHRS.cs
+++ b/SourceCode/GPS/Classes/CAHRS.cs
@@ -16,7 +16,7 @@
         public double rollFilter;
 
         //is the auto steer in auto turn on mode or not
-        public bool isAutoSteerAuto, isRollInvert, isReverseOn;
+        public bool isAutoSteerAuto, isRollInvert, autoswitchDualFix2FixOn, isDualAsIMU, isReverseOn;
 
         //the factor for fusion of GPS and IMU
         public double forwardComp, reverseComp, fusionWeight;
@@ -40,6 +40,8 @@
             isRollInvert = Properties.Settings.Default.setIMU_invertRoll;
 
             isReverseOn = Properties.Settings.Default.setIMU_isReverseOn;
+
+            autoswitchDualFix2FixOn = Properties.Settings.Default.setAutoswitchDualFix2FixOn;
         }
     }
 }
--- a/SourceCode/GPS/Forms/Position.designer.cs
+++ b/SourceCode/GPS/Forms/Position.designer.cs
@@ -150,10 +150,9 @@ namespace AgOpenGPS
             pn.speed = pn.vtgSpeed;
             pn.AverageTheSpeed();
 
-            // FOR AUTOSWITCH DUALFIX START
-            if (Properties.Settings.Default.setGPS_headingFromWhichSource == "Dual" && ahrs.autoSwitchDualFixOn == true)
+            if (Properties.Settings.Default.setGPS_headingFromWhichSource == "Dual" && ahrs.autoSwitchDualFixOn)
             {
-                if (pn.speed > ahrs.autoSwitchDualFixSpeed || pn.speed < (ahrs.autoSwitchDualFixSpeed * -1))
+                if (Math.Abs(pn.speed) > ahrs.autoSwitchDualFixSpeed)
                 {
                     headingFromSource = "Fix";
                     ahrs.isDualAsIMU = true;
@@ -165,7 +164,6 @@ namespace AgOpenGPS
                     ahrs.imuHeading = 99999;
                 }
             }
-            // FOR AUTOSWITCH DUALFIX END
 
             #region Heading
             switch (headingFromSource)
@@ -175,12 +173,10 @@ namespace AgOpenGPS
                     {
                         #region Start
 
-                        // FOR AUTOSWITCH DUALFIX START
-                        if (Properties.Settings.Default.setGPS_headingFromWhichSource == "Dual" && ahrs.autoSwitchDualFixOn == true)
+                        if (Properties.Settings.Default.setGPS_headingFromWhichSource == "Dual" && ahrs.autoSwitchDualFixOn)
                         {
                             lblSpeed.ForeColor = System.Drawing.Color.Red;
                         }
-                        // FOR AUTOSWITCH DUALFIX END
 
                         distanceCurrentStepFixDisplay = glm.Distance(prevDistFix, pn.fix);
                         distanceCurrentStepFixDisplay *= 100;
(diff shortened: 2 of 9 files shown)


chrisriteco:
@FantasyTeddy 

1. The name 'Fix2Fix' was indeed created by us, better to change this to 'Dual to Fix' in terms of text and variables for uniformity.

2. By default, this new option should indeed be disabled.

3. In theory, both the Dual and Fix settings should be active when the 'Autoswitch Dual <> Fix' option is active. Both are relevant. Do you agree?

4. I have retrieved the variable 'isDualAsIMU' and made it active when the new option is enabled, so that in Fix mode the value of the Dual is assigned to 'ahrs.imuHeading'. This ensures that: 1. The direction of travel from the Fix is ​​Fused/corrected. 2. When the Fix logic is switched to the byPass section, the direction of travel automatically falls back to the value of 'ahrs.imuHeading'. This allowed me to leave as much logic intact as possible.

What is useful now, can I implement the discussed changes in my Branch and report it again in the PR? Thanks for your review.

FantasyTeddy:
> 1. The name 'Fix2Fix' was indeed created by us, better to change this to 'Dual to Fix' in terms of text and variables for uniformity.

Yes I think that makes it more in-line with the rest of the UI (and is even shorter).

> 2. By default, this new option should indeed be disabled.

👍 

> 3. In theory, both the Dual and Fix settings should be active when the 'Autoswitch Dual <> Fix' option is active. Both are relevant. Do you agree?

Yes I do. @richardklasens what do you think?

> 4. I have retrieved the variable 'isDualAsIMU' and made it active when the new option is enabled, so that in Fix mode the value of the Dual is assigned to 'ahrs.imuHeading'. This ensures that: 1. The direction of travel from the Fix is ​​Fused/corrected. 2. When the Fix logic is switched to the byPass section, the direction of travel automatically falls back to the value of 'ahrs.imuHeading'. This allowed me to leave as much logic intact as possible.

Okay I cannot follow entirely, but I guess it is fine then.

> What is useful now, can I implement the discussed changes in my Branch and report it again in the PR?

Yes you can just add commits to your branch and the PR will automatically update.

> Thanks for your review.

Thank you for the contribution!

FantasyTeddy:
One more thing: do you think it could be useful to make the threshold of 2 km/h configurable?

richardklasens:
> One more thing: do you think it could be useful to make the threshold of 2 km/h configurable?

I would prefer that with a minimum of 1 and a max of 10 for a start.

richardklasens:
> > 3. In theory, both the Dual and Fix settings should be active when the 'Autoswitch Dual <> Fix' option is active. Both are relevant. Do you agree?
> 
> Yes I do. @richardklasens what do you think?

when you select auto switch function it should also disable the single checkbox. Because it’s a very special mode.

chrisriteco:
- Names of variables and texts adjusted. Function is now called 'Auto Dual <> Fix'.
- Function is disabled by default.
- Both Dual and Fix settings are now active when the new function is enabled.
- Checkbox for Fix mode is disabled when the new option is enabled.
- Speed ​​is now adjustable from 1 to 10 km/h.

SourceCode/GPS/Forms/Position.designer.cs
```diff
@@ -150,10 +150,10 @@ public void UpdateFixPosition()
             pn.speed = pn.vtgSpeed;
             pn.AverageTheSpeed();
 
-            // FOR AUTOSWITCH DUALFIX2FIX START
-            if (Properties.Settings.Default.setGPS_headingFromWhichSource == "Dual" && ahrs.autoswitchDualFix2FixOn == true)
+            // FOR AUTOSWITCH DUALFIX START
+            if (Properties.Settings.Default.setGPS_headingFromWhichSource == "Dual" && ahrs.autoSwitchDualFixOn == true)
```
J005t67, line 154: Please leave out the '== true'

SourceCode/GPS/Forms/Position.designer.cs
```diff
@@ -150,10 +150,10 @@ public void UpdateFixPosition()
...
             {
-                if (pn.speed > 2 || pn.speed < -2)
+                if (pn.speed > ahrs.autoSwitchDualFixSpeed || pn.speed < (ahrs.autoSwitchDualFixSpeed * -1))
```
J005t67, line 156: Use Math.Abs(ahrs.autoSwitchDualFixSpeed) to simplify expression

SourceCode/GPS/Forms/Position.designer.cs
```diff
@@ -175,12 +175,12 @@ public void UpdateFixPosition()
                     {
                         #region Start
 
-                        // FOR AUTOSWITCH DUALFIX2FIX START
-                        if (Properties.Settings.Default.setGPS_headingFromWhichSource == "Dual" && ahrs.autoswitchDualFix2FixOn == true)
+                        // FOR AUTOSWITCH DUALFIX START
+                        if (Properties.Settings.Default.setGPS_headingFromWhichSource == "Dual" && ahrs.autoSwitchDualFixOn == true)
```
J005t67, line 179: Please remove the part '== true'

SourceCode/GPS/Forms/Settings/ConfigData.Designer.cs
```diff
@@ -188,6 +210,30 @@ private void hsbarFusion_ValueChanged(object sender, EventArgs e)
             mf.ahrs.fusionWeight = (double)hsbarFusion.Value * 0.002;
         }
 
+        private void cboxIsAutoSwitchDualFixOn_CheckedChanged(object sender, EventArgs e)
+        {
+            if (cboxIsAutoSwitchDualFixOn.Checked == true)
+            {
```
J005t67, line 216: This is duplicate code( it is the same as lines 129 to 140). Please move it to a separate function and reuse it.

SourceCode/GPS/Forms/Position.designer.cs
```diff
@@ -150,10 +150,10 @@ public void UpdateFixPosition()
             pn.speed = pn.vtgSpeed;
             pn.AverageTheSpeed();
 
-            // FOR AUTOSWITCH DUALFIX2FIX START
-            if (Properties.Settings.Default.setGPS_headingFromWhichSource == "Dual" && ahrs.autoswitchDualFix2FixOn == true)
+            // FOR AUTOSWITCH DUALFIX START
```
FantasyTeddy, line 153: Do we need these comments? If I imagine that we would add comments like this for every feature that we have, it would be a huge mess... better to use methods with descriptive names to encapsulate certain functionality.

SourceCode/GPS/Properties/Settings.cs
```diff
@@ -274,7 +274,8 @@ public static Settings Default
         public bool setDisplay_isSectionLinesOn = true;
         public bool setDisplay_isLineSmooth = false;
         public Size setWindow_tramLineSize = new Size(921, 676);
-        public bool setAutoswitchDualFix2FixOn = true;
+        public bool setAutoSwitchDualFixOn = false;
+        public int setAutoSwitchDualFixSpeed = 2;
```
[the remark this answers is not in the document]
FantasyTeddy: Yes, I think that should be a `double`.

SourceCode/GPS/Classes/CAHRS.cs
```diff
@@ -41,7 +45,9 @@ public CAHRS()
 
             isReverseOn = Properties.Settings.Default.setIMU_isReverseOn;
 
-            autoswitchDualFix2FixOn = Properties.Settings.Default.setAutoswitchDualFix2FixOn;
+            autoSwitchDualFixOn = Properties.Settings.Default.setAutoSwitchDualFixOn;
```
J005t67, line 48: 'Somewhere in the basement of AOG' we have a struct called vecFix2Fix. Do you know if it has anything to do with this PR? If so, can it be removed?

FantasyTeddy:
> when you select auto switch function it should also disable the single checkbox. Because it’s a very special mode.

@richardklasens: I'm not sure why this is really necessary. Why shouldn't you be allowed to switch to "Fix" if this mode is enabled? (It makes it really hard to figure out why the button cannot be clicked...)

SourceCode/GPS/Forms/Settings/ConfigData.Designer.cs
```diff
@@ -126,18 +124,7 @@ private void rbtnHeadingFix_CheckedChanged(object sender, EventArgs e)
...
-                    labelGboxSingle.Enabled = true;
-                    labelGboxDual.Enabled = true;
-                }
-                else
-                {
-                    rbtnHeadingFix.Enabled = true;
-                    labelGboxSingle.Enabled = false;
-                    labelGboxDual.Enabled = true;
-                }
+                this.setAutoSwitchDualFixPanelOptions();
```
FantasyTeddy, line 127: Please do not prefix with `this` and capitalize the first letter: `SetAutoSwitchDualFixPanelOptions`

SourceCode/GPS/Classes/CAHRS.cs
chrisriteco: The struct called vecFix2Fix does not belong to this PR. The name may suggest otherwise, do not use anything with the name Fix2Fix anymore.

FantasyTeddy (dismissed):

FantasyTeddy (approved):

PR #607, opened 2025-01-22
Add ISOBUS integration

GwnDaan:
Hi all, I'm thrilled to share the first integration of the ISOBUS TC (task controller): [GwnDaan/AOG-TaskController] (https://github.com/GwnDaan/AOG-TaskController) with AgOpenGPS. It communicates with AOG via UDP packets, and can be started via AgIO after it is installed. 

For those who want to try it out, the latest installer is always built via the pipeline on GitHub. (There is no release yet for as long as the TODO list is unfinished).

As of 22-01-2025 11:32 CET: [download the latest installer here](https://github.com/GwnDaan/AOG-TaskController/actions/runs/12918260572/artifacts/2471238200)

I will explain the main changes to AOG below with comments.

There is still a small TODO list I want to finish before merging this:
- [ ] Automatically start ISOBUS if it was active when AgIO was shut down.
- [ ] Add support for other known CAN adapters. (Currently, only PCAN can be configured).
- [ ] Add a way to “bridge” can-messages to a “virtual” canbus, so that the windows ISOBUS [Virtual Terminal](https://github.com/Open-Agriculture/AgIsoVirtualTerminal) can also run in parallel with AOG using the same can-adapter.

For those who want to know more about ISOBUS in general, I recommend reading https://www.csselectronics.com/pages/isobus-introduction-tutorial-iso-11783.

Bare with me with this PR, as I don't really understand the codebase (nor C# 😅), so any suggestions are greatly appreciated.


SourceCode/GPS/Classes/CISOBUS.cs
```diff
@@ -0,0 +1,157 @@
...
+            // Send the request
+            byte[] data = new byte[7];
+            data[0] = 0x80; // standard AIO header
+            data[1] = 0x81; // PGN header
+            data[2] = 0x7F; // SRC address
+            data[3] = 0xF1; // PGN
+            data[4] = 0x02; // Length
+            data[5] = (byte)(enabled ? 0x01 : 0x00); // Section control enabled request
+            mf.SendPgnToLoop(data);
+        }
```
GwnDaan, line 45: Here we have created a new PGN that notifies ISOBUS whether or not AOG is ready/wants to control the sections of the ISOBUS implement. It corresponds with the following button:
image

SourceCode/GPS/Classes/CISOBUS.cs
```diff
@@ -0,0 +1,157 @@
...
+            }
+            else if (!wasAlive && isAlive)
+            {
+                // Reconnected
+                mf.btnIsobusSC.Visible = true;
+            }
+
+            wasAlive = isAlive;
+            return isAlive;
+        }
```
GwnDaan, line 89: We determine if an ISOBUS implement is connected by checking if we received an heartbeat within the last second. This heartbeat contains all the information AOG needs to correctly display the current state of the TC.

SourceCode/GPS/Forms/OpenGL.Designer.cs
```diff
@@ -1228,6 +1228,18 @@ private void oglBack_Paint(object sender, PaintEventArgs e)
...
+            if (isobus.IsAlive())
+            {
+                for (int j = 0; j < tool.numOfSections; j++)
+                {
+                    if (isobus.IsSectionOn(j))
+                        section[j].isMappingOn = true;
+                    else
+                        section[j].isMappingOn = false;
+                }
+            }
```
GwnDaan, line 1241: If we have the ISOBUS implement connected, we override the mapping of all sections in AOG with their "actual" state. It might happen that AOG wants a section to be on, but the ISOBUS implement cannot fulfill that request. With this piece of code the mapping still represents the actual worked area.

SourceCode/GPS/Forms/Position.designer.cs
```diff
@@ -871,12 +871,7 @@ public void UpdateFixPosition()
...
                 lightbarDistance = guidanceLineDistanceOff;
 
-                if (!isBtnAutoSteerOn) //32020 means auto steer is off
-                {
-                    guidanceLineDistanceOff = 32020;
-                    p_254.pgn[p_254.status] = 0;
-                }
-
+                if (!isBtnAutoSteerOn) p_254.pgn[p_254.status] = 0;
                 else p_254.pgn[p_254.status] = 1;
```
GwnDaan, line 876: Currently the XTE / guidanceLineDistanceOff doesn't reflect the actual value when auto steer is disengaged. I think that's wrong, as you still have a valid distance to the line even when not using auto steer

SourceCode/GPS/Classes/CISOBUS.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        public void RequestSectionControlEnabled(bool enabled)
+        {
+            // Send the request
+            byte[] data = new byte[7];
+            data[0] = 0x80; // standard AIO header
+            data[1] = 0x81; // PGN header
+            data[2] = 0x7F; // SRC address
+            data[3] = 0xF1; // PGN
+            data[4] = 0x02; // Length
```
FantasyTeddy, line 42: Shouldn't the length only be 1?

SourceCode/GPS/Classes/CISOBUS.cs
FantasyTeddy: Should we maybe document this PGN here as well: https://github.com/AgOpenGPS-Official/Boards/blob/main/PGN.md ?

SourceCode/GPS/Classes/CISOBUS.cs
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        public bool IsSectionControlEnabled()
+        {
+            return sectionControlEnabled;
+        }
+
+        private void SetSectionControlEnabled(bool enabled)
+        {
+            if (sectionControlEnabled != enabled)
```
FantasyTeddy, line 54: Here we could again use an early return:
```suggestion
            if (sectionControlEnabled == enabled)
                return;
```

SourceCode/GPS/Classes/CISOBUS.cs
```diff
@@ -0,0 +1,157 @@
...
+            data[3] = 0xF1; // PGN
+            data[4] = 0x02; // Length
+            data[5] = (byte)(enabled ? 0x01 : 0x00); // Section control enabled request
+            mf.SendPgnToLoop(data);
+        }
+
+        public bool IsSectionControlEnabled()
+        {
+            return sectionControlEnabled;
+        }
```
FantasyTeddy, line 50: Instead of having accessor methods (`IsSectionControlEnabled` & `SetSectionControlEnabled`), could we convert this into a property:
```suggestion
        public bool SectionControlEnabled
        {
            get => sectionControlEnabled;
            private set
            {
                ...
            }
        }
```
?

SourceCode/GPS/Classes/CISOBUS.cs
```diff
@@ -0,0 +1,157 @@
...
+using System.Threading.Tasks;
+
+namespace AgOpenGPS
+{
+    public class CISOBUS
+    {
+        private readonly FormGPS mf;
+
+        private bool wasAlive;
+        private long timestamp;
```
FantasyTeddy, line 14: Can we store the timestamp as a `DateTimeOffset`? That way you could compare the values easier in `IsAlive`.

SourceCode/GPS/Classes/CISOBUS.cs
```diff
@@ -0,0 +1,157 @@
...
+                // Reconnected
+                mf.btnIsobusSC.Visible = true;
+            }
+
+            wasAlive = isAlive;
+            return isAlive;
+        }
+
+
+        private static bool ReadBit(byte[] data, int bitIndex)
```
FantasyTeddy, line 92: To me this method seems overly complicated. I would just pass the correct `byte` as the first argument and then the `bitIndex`. That way the calls do not require any math.

SourceCode/GPS/Classes/CISOBUS.cs
```diff
@@ -0,0 +1,157 @@
...
+            int byteIndex = bitIndex / 8; // Find the byte
+            if (data.Length <= byteIndex)
+            {
+                return false;
+            }
+            int bitPosition = bitIndex % 8; // Find the bit within the byte
+            return (data[byteIndex] & (1 << bitPosition)) != 0;
+        }
+
+        private static void WriteBit(ref byte[] data, int bitIndex, bool value)
```
FantasyTeddy, line 103: This method isn't used. Can we remove it?

SourceCode/GPS/Classes/CISOBUS.cs
```diff
@@ -0,0 +1,157 @@
...
+            {
+                // Make sure we have enough data to read all the section states
+                return false;
+            }
+
+            for (int i = 0; i < numberOfSections; i++)
+            {
+                bool sectionState = ReadBit(data, 16 + i);
+                if (this.actualSectionStates.Count <= i)
+                    this.actualSectionStates.Add(sectionState);
```
FantasyTeddy, line 143: Maybe we could convert `actualSectionStates` into an array (`bool[]`) and then simply re-assign it before the `for` loop to `new bool[numberOfSections]`.
Then we can simply set the values inside and don't need to check for count or remove afterwards.

SourceCode/GPS/Forms/FormGPS.Designer.cs
```diff
@@ -199,6 +199,7 @@ private void InitializeComponent()
             this.btnAutoYouTurn = new System.Windows.Forms.Button();
             this.btnSectionMasterAuto = new System.Windows.Forms.Button();
             this.btnSectionMasterManual = new System.Windows.Forms.Button();
+            this.btnIsobusSC = new System.Windows.Forms.Button();
```
FantasyTeddy, line 202: I'm not sure if we should abbreviate "section control" here. It took me a second to figure out what "SC" meant.

FantasyTeddy (commented):
Really cool addition and well done!

SourceCode/GPS/Classes/CISOBUS.cs
J005t67:
```suggestion
        public bool IsAlive()
        {
            // Check if the timestamp is not older than 1 second
            bool isAlive = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds() - timestamp < 1000;

            mf.btnIsobusSC.Visible = isAlive;

            wasAlive = isAlive;
            return isAlive;
        }
```

J005t67: public bool IsAlive()
        {
            // Check if the timestamp is not older than 1 second
            bool isAlive = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds() - timestamp < 1000;
```suggestion
           mf.btnIsobusSC.Visible = isAlive;
```
            wasAlive = isAlive;
            return isAlive;
        }

J005t67: Seems that visibility can be set like this
mf.btnIsobusSC.Visible = isAlive

without doing any tests

SourceCode/AgIO/Source/Forms/FormLoop.cs
```diff
@@ -85,6 +85,15 @@ private void FormLoop_Load(object sender, EventArgs e)
...
                 LoadUDPNetwork();
                 Log.EventWriter("UDP Network Is On");
+
+                // Start ISOBUS if needed
+                if (Settings.Default.isobus_isOn)
+                {
+                    // Little hack to load ISOBUS form without showing it
+                    isobusForm.Show();
+                    isobusForm.Hide();
+                    isobusForm.StartAogTaskController();
```
FantasyTeddy, line 95: It feels wrong to do that. Maybe we should think about separating the form-logic from the actual task-controller-logic. If you introduce a new class that handles all the task-controller stuff (process logic, data buffer, ...) you can then simply display the relevant information in the form once it is opened. This would also allow us to only create the form when it is needed and not have it globally available and hiding it in most cases.

SourceCode/AgIO/Source/Forms/FormISOBUS.cs
```diff
@@ -0,0 +1,317 @@
...
+                    string adapter = cboxRadioAdapter.SelectedItem.ToString();
+                    if (adapter == "PEAK-PCAN")
+                    {
+                        flowLayoutChannel.Visible = true;
+                        cboxRadioChannel.Items.Clear();
+                        for (int i = 1; i <= 16; i++)
+                        {
+                            cboxRadioChannel.Items.Add(i);
+                        }
+                    }
```
FantasyTeddy, line 247: Could we maybe simplify this `if else` chain by using a lookup table? For example, define a `Dictionary<string, int> adapterChannels` that maps the adapter to the number of channels.

SourceCode/GPS/Classes/CISOBUS.cs
```diff
@@ -39,86 +38,49 @@ public void RequestSectionControlEnabled(bool enabled)
...
+                    mf.btnIsobusSectionControl.Image = Properties.Resources.IsobusSectionControlOff;
                 }
             }
         }
 
         public bool IsAlive()
         {
             // Check if the timestamp is not older than 1 second
-            bool isAlive = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds() - timestamp < 1000;
+            bool isAlive = (timestamp != default && DateTimeOffset.Now - timestamp < TimeSpan.FromSeconds(1));
```
FantasyTeddy, line 71: Did you change this to `DateTimeOffset.Now` instead of `UtcNow` on purpose?

SourceCode/AgIO/Source/Forms/FormLoop.cs
GwnDaan: Yes, definitely. That would be the optimal approach, but with still a lot of things to be added down the line, I feel like it might be okay to postpone this for now? I'm not sure yet how the other things will turn out

SourceCode/GPS/Classes/CISOBUS.cs
GwnDaan: Yeah, no copilot chose `DateTimeOffset.Now` this time around. I think it's fine since all occurrences with this timestamp have been changed to the `.Now`?

SourceCode/AgIO/Source/Forms/FormLoop.cs
FantasyTeddy: Can you see any reason why separating it would cause problems down the line? If you design it like that early on, I think you can avoid having to redesign it later.

SourceCode/GPS/Classes/CISOBUS.cs
FantasyTeddy: Yes, I think it shouldn't be an issue. I was just remarking it, because usually `Now` is used whenever you want to display the date & time, whereas `UtcNow` is used for thinks like this where it shouldn't change depending on the timezone (which it will not anyways).

SourceCode/GPS/Forms/Position.designer.cs
farmerbriantee: Some of the old steer boards used 32020 and 32000 as status, before status existed.

gunicsba:
I've tested this on a couple implements and it seems to be working well. Peak PCAN always nicely reconnects in the background, etc.

richardklasens:
@GwnDaan is this still actual and somewhat ready for merge?

SourceCode/AgIO/Source/Forms/FormLoop.cs
GwnDaan: No I don't see any issues that it'll cause. It's just that I don't have the spare time available to make the changes :/

GwnDaan:
> @GwnDaan is this still actual and somewhat ready for merge?

I’ve updated the code to align with the latest _develop_ branch. Unfortunately, I don’t currently have the time to address all the requested review changes, but based on testing by several people, it seems stable and ready to ship.

If anyone is interested in continuing the development of ISOBUS in AOG, that would be greatly appreciated, my spare time is quite limited at the moment. I’m happy to provide support, explanations, and documentation to help anyone who wants to pick it up.

SourceCode/AgIO/Source/Forms/FormLoop.cs
GwnDaan: Oh I forgot to add: ultimately one day everything inside the [AOG-TaskController.exe](https://github.com/GwnDaan/AOG-TaskController)https://github.com/[GwnDaan/AOG-TaskController](https://github.com/GwnDaan/AOG-TaskController) runs on the Teensy. That way there is no need to start any process at all and the logic above can be removed.

SourceCode/AgIO/Source/Forms/FormLoop.Designer.cs
```diff
@@ -1244,6 +1255,7 @@ private void InitializeComponent()
         private ToolStripMenuItem toolStripLogViewer;
         private ToolStripMenuItem toolStripAgDiag;
         private ToolStripMenuItem toolStripSettings;
+        private ToolStripMenuItem deviceManagerToolStripMenuItem;
```
FantasyTeddy, line 1258: @GwnDaan: If you delete this line, the build will be successful.
For some reason I am not able to add this change to the branch myself...

FantasyTeddy (approved):

PR #610, opened 2025-01-24
Take over Field aspect models from AOG2 repository

J005t67:
This pull request contains the following (a.o.)

1) New Field Aspect Models like FlagList, Boundary, TramLines, and RecordedPath. These classes are built on top of the new base classes, like GeoPolygon and GeoPath and require less code than the current implementations, and keep things pretty organized. (The current implementations are scattered  over the large *.Designer.cs files). The list is not complete.  Aspects like Contour, WorkedArea and the background image wil follow later.

2) The class Field that will contain all 'Field Aspect Models'

3) Streaming classes for reading and writing the field aspect models from/to file. This will replace the very lengthy code in SaveOpen.Designer.cs later. The various 'field aspect streamers' use the new classes GeoStreamReader and GeoStreamWriter to achieve a large code reduction. 

4) A number of interfaces that help to keep the 'Core' code separated from the Winforms or WPF specific parts.

Note 1) : After finishing this PR, the current code is still in place and can be used to check the correctness of the refactored code. (For example by comparing the results after reading or writing.)

Note 2) After finisinh this PR, the refactored code is still not used or connected to the original code. I will first do another PR for the new refactored 'Drawing code'  to draw the new Field aspect models. After that I will start implementing new ViewModels and start using the refactored code.


SourceCode/AgOpenGPS.Core/Models/Field/Flag.cs
```diff
@@ -0,0 +1,41 @@
+﻿namespace AgOpenGPS.Core.Models
+{
+    public enum FlagColor { Red = 0, Green = 1, Yellow = 2}
+
+    public class Flag
+    {
+        public Flag(Wgs84 wgs84, GeoCoord geoCoord, GeoDir heading, FlagColor flagColor, int uniqueNumber, string notes)
+        {
+            Wgs84 = new Wgs84(wgs84.Latitude, wgs84.Longitude);
```
FantasyTeddy, line 9: Isn't `Wgs84` a `struct` and therefore copied anyways when passed to this constructor?

SourceCode/AgOpenGPS.Core/Models/Field/Flag.cs
```diff
@@ -0,0 +1,41 @@
...
+        public GeoDir Heading => GeoCoordDir.Direction;
+        public double Northing => GeoCoord.Northing;
+        public double Easting => GeoCoord.Easting;
+
+        public FlagColor FlagColor { get; }
+        public int UniqueNumber { get; set; }
+
+        public string Notes { get; set;}
+
+        public string ColorAsXbgrHexString()
```
FantasyTeddy, line 31: Shouldn't this maybe be an extension method for `FlagColor` itself?

SourceCode/AgOpenGPS.Core/Models/Field/FlagList.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.Collections.Generic;
+
+namespace AgOpenGPS.Core.Models
+{
+    public class FlagList
```
FantasyTeddy, line 5: Is a class like this really necessary? I don't see any benefit from using `List<Flag>` directly.

SourceCode/AgOpenGPS.Core/Models/Field/TramLines.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        public TramLines()
+        {
+        }
+
+        public GeoPolygon OuterTrack
+        {
+            get { return _outerTrack; }
+            set { _outerTrack = value; }
+        }
```
FantasyTeddy, line 19:
```suggestion
        public GeoPolygon OuterTrack { get; set; }
```

SourceCode/AgOpenGPS.Core/Streamers/Field/FieldAspectStreamer.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public FieldAspectStreamer(
+            ILogger logger,
+            string defaultFileName)
+        {
+            _logger = logger;
+            _defaultFileName = defaultFileName;
+        }
+
+        public void SetPresenter(IFieldStreamerPresenter presenter)
```
FantasyTeddy, line 21: Why does this need to be set separately and not passed to the constructor?

SourceCode/AgOpenGPS.Core/Streamers/Field/FieldAspectStreamer.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            _presenter = presenter;
+        }
+
+        public string FullPath(string fieldPath, string fileName = null)
+        {
+            string fn = fileName ?? _defaultFileName;
+            return fieldPath + "\\" + fn;
```
FantasyTeddy, line 29:
```suggestion
            fileName = fileName ?? _defaultFileName;
            return Path.Combine(fieldPath, fileName);
```

SourceCode/AgOpenGPS.Core/Streamers/Field/FieldAspectStreamer.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        public void CreateDirectory(string fieldPath)
+        {
+            string dirField = fieldPath + "\\";
+
+            string directoryName = Path.GetDirectoryName(dirField);
```
FantasyTeddy, line 36: Isn't this the same as just using `fieldPath` as the directory name?

```suggestion
            string directoryName = fieldPath ;
```

SourceCode/AgOpenGPS.Core/Streamers/Field/BoundaryStreamer.cs
```diff
@@ -0,0 +1,105 @@
...
+    public class BoundaryStreamer : FieldAspectStreamer
+    {
+        public BoundaryStreamer(
+            ILogger logger
+        ) :
+            base(logger, "Boundary.txt")
+        {
+        }
+
+        public Boundary TryRead(string fieldPath)
```
FantasyTeddy, line 17: Usually, `Try...` methods have a slightly different signature:
```suggestion
        public bool TryRead(string fieldPath, out Boundary boundary)
```
That way you can check as a caller if the read was successful or not.

SourceCode/AgOpenGPS.Core/Streamers/Field/BoundaryStreamer.cs
```diff
@@ -0,0 +1,105 @@
...
+                _presenter.PresentBoundaryFileMissing();
+            }
+            else
+            {
+                try
+                {
+                    boundary = Read(fieldPath);
+                }
+
+                catch (Exception e)
```
FantasyTeddy, line 32: It is generally considered bad practice to catch the general `Exception` class. In this case it would probably make sense to catch `IOException`.

SourceCode/AgOpenGPS.Core/Streamers/Field/BoundaryStreamer.cs
```diff
@@ -0,0 +1,105 @@
...
+            return boundary;
+        }
+
+        private BoundaryPolygon ReadBoundaryPolygon(GeoStreamReader reader)
+        {
+            if (reader.EndOfStream) return null;
+
+            BoundaryPolygon polygon = new BoundaryPolygon();
+            bool peekedBool = false;
+            if (reader.PeekReadBool(ref peekedBool))
```
FantasyTeddy, line 64: Maybe using an `out` parameter here would be better suitable?

SourceCode/AgOpenGPS.Core/Streamers/Field/BoundaryStreamer.cs
```diff
@@ -0,0 +1,105 @@
...
+                }
+            }
+        }
+
+        public void CreateFile(string fieldPath)
+        {
+            CreateDirectory(fieldPath);
+            using (GeoStreamWriter writer = new GeoStreamWriter(FullPath(fieldPath)))
+            {
+            }
```
FantasyTeddy, line 101:
```suggestion
            File.Create(FullPath(fieldPath));
```

SourceCode/AgOpenGPS.Core/Streamers/Field/FieldStreamer.cs
```diff
@@ -0,0 +1,90 @@
...
+            _presenter = presenter;
+            _flagsStreamer.SetPresenter(presenter);
+            _recordedPathStreamer.SetPresenter(presenter);
+            _tramLinesStreamer.SetPresenter(presenter);
+        }
+
+        public string CurrentFieldPath
+        {
+            set { _currentFieldPath = value; }
+        }
```
FantasyTeddy, line 38: Why can this property only be set?

SourceCode/AgOpenGPS.Core/Streamers/Field/GeoStreamReader.cs
```diff
@@ -0,0 +1,111 @@
...
+    public class GeoStreamReader : StreamReader
+    {
+        public GeoStreamReader(string fullPath) : base(fullPath)
+        {
+        }
+
+        public bool ParseBool(string boolWord)
+        {
+            return bool.Parse(boolWord);
+        }
```
FantasyTeddy, line 17: Does it really make sense to expose this as a member method on the reader? All the `Parse...` methods in this class have no dependency on the reader itself and I think it would be better to have a separate helper(?) class that could offer these kinds of methods if they are needed at all.

SourceCode/AgOpenGPS.Core/Streamers/Field/GeoStreamWriter.cs
```diff
@@ -0,0 +1,113 @@
...
+    public class GeoStreamWriter : StreamWriter
+    {
+        public GeoStreamWriter(string fullPath) : base(fullPath)
+        {
+        }
+
+        public string BoolString(bool boolValue)
+        {
+            return boolValue.ToString(CultureInfo.InvariantCulture);
+        }
```
FantasyTeddy, line 17: Here the same argument as for the `Parse...` methods in `GeoStreamReader`; I think this should be in a different class if it is needed at all.

SourceCode/AgOpenGPS.Core/Streamers/Field/GeoStreamWriter.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+            return intValue.ToString(CultureInfo.InvariantCulture);
+        }
+
+        public string DoubleString(double value, string formatString)
+        {
+            return value.ToString(formatString, CultureInfo.InvariantCulture);
+        }
+
+        public string Wgs84String(Wgs84 wgs84, string formatString = "N7")
```
FantasyTeddy, line 29: This would for example be an excellent place to have an extension method for `Wgs84` called `ToString`. That way we could mimic the behavior of the other `ToString` methods.

SourceCode/AgOpenGPS.Core/Streamers/Field/RecordedPathStreamer.cs
```diff
@@ -0,0 +1,97 @@
...
+                _logger.LogError("Load Recorded Path" + e.ToString());
+            }
+            return recordedPath;
+        }
+
+        public void TryWrite(RecordedPath recordedPath, string fieldPath, string fileName)
+        {
+            CreateDirectory(fieldPath);
+            Write(recordedPath, fieldPath, fileName);
+        }
```
FantasyTeddy, line 37: If this is `TryWrite`, shouldn't there be some kind of `try`-`catch` somewhere?

SourceCode/AgOpenGPS.Core/Streamers/Field/RecordedPathStreamer.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        private RecordedPath Read(string fieldPath, string fileName)
+        {
+            RecordedPath recordedPath = new RecordedPath();
+            string fn = fileName != null ? fileName : _defaultFileName;
+            string fileAndDirectory = fieldPath + "\\" + fn;
```
FantasyTeddy, line 43: Why not use the `FullPath` method?

SourceCode/AgOpenGPS.Core/Streamers/Field/RecordedPathStreamer.cs
```diff
@@ -0,0 +1,97 @@
...
+            if (File.Exists(fileAndDirectory))
```
FantasyTeddy, line 44: Invert the `if` and do an early return.
```suggestion
            if (!File.Exists(fileAndDirectory))
                return recordedPath;
```

SourceCode/AgOpenGPS.Core/Streamers/Field/TramLinesStreamer.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            TramLines tramLines = new TramLines();
+            string fullPath = FullPath(fieldPath);
+            if (File.Exists(fullPath))
+            {
+                using (GeoStreamReader reader = new GeoStreamReader(fullPath))
+                {
+                    reader.ReadLine(); // skip header: $Tram
+
+                    if (-1 != reader.Peek())
```
FantasyTeddy, line 41: It feels very weird to me to do a comparison that way. Is there a particular reason, why you have the constant (`-1`) on the left side of the comparison?

FantasyTeddy:
Something else to consider: can we add tests for reading/writing these files? I think we need to keep testability in mind, because that's what is currently lacking and makes a lot of refactorings scary because nobody knows if it still works (except if we do a lot of manual testing).

SourceCode/AgOpenGPS.Core/Models/Base/GeoPolygons/GeoPolygonWithHeading.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        public int Count => _polygon.Count;
+        public double Area => _polygon.Area;
+
+        public GeoCoord this[int index]
+        {
+            get { return _polygon[index]; }
+        }
+
+        public void Add(GeoCoord coordDir, GeoDir geoDir)
```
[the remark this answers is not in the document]
J005t67: Yes, that will be more accurate. I will change the first one 'coordDir' to, because that really does not make sense.

SourceCode/AgOpenGPS.Core/Streamers/Field/TramLinesStreamer.cs
J005t67: I did it for years like that, but maybe it is time now to stop. The idea is that once in a while people introduce bugs by writing things like

if (a =5)... or if (path = null)
I will check if the IDE helps to prevent those mistakes. If so, I will change it.

J005t67:
> Something else to consider: can we add tests for reading/writing these files? I think we need to keep testability in mind, because that's what is currently lacking and makes a lot of refactorings scary because nobody knows if it still works (except if we do a lot of manual testing).

Yes, in general tests will help a big deal. In this particular case: it would be nice to check that reading and writing back a certain field aspect will not change the file content. (Or writing a cetain aspect to file and reading back  will give the same result.)
During refactoring I will compare the results of the new code with the results of the old code, but that will of course not longer work when I remove the old code.

SourceCode/AgOpenGPS.Core/Models/Field/Flag.cs
J005t67: Yes, I will correct this. (I moved back and forth a bit in AOG2, changing it from struct to class and back again in AOG2)

SourceCode/AgOpenGPS.Core/Streamers/Field/GeoStreamWriter.cs
FantasyTeddy: Okay, I get your point. But I think the methods should be `protected` then, such that only the overloads can access them.

SourceCode/AgOpenGPS.Core/Streamers/Field/FieldAspectStreamer.cs
J005t67: Typically, Presenters need to know a few ViewModels and ViewModels need to know one or more Presenters. So passing everything in constructors is impossible. That would be a cyclic dependency in terms of which order the objects must be created.

SourceCode/AgOpenGPS.Core/Streamers/Field/FieldAspectStreamer.cs
J005t67: Yes, two improvements.

SourceCode/AgOpenGPS.Core/Streamers/Field/TramLinesStreamer.cs
FantasyTeddy: Yes that is certainly a potential issue in other languages (like C or C++), but in C# the assignment operation is not convertible into a boolean, so it cannot be used wrong in that scenario.
(There is a possibility to use a `bool` assignment instead of a comparison, but the compiler will issue compiler warning [CS0665](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0665) in that case.)

SourceCode/AgOpenGPS.Core/Interfaces/ILogger.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace AgOpenGPS.Core.Interfaces
+{
+    public interface ILogger
```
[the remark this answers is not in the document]
J005t67: I have a small implementation in AOG2. That is simply the Logger that was present in develop a few weeks ago, before you moved it to the AgLibrary. I agree that  we want to migrate to a more mature Logging framework, but let's do that after I completed this PR.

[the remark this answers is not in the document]
FantasyTeddy: Okay, sounds good 👍

SourceCode/AgOpenGPS.Core/Models/Field/Flag.cs
J005t67: Don't know yet. Removed it for now and reconsider later.

SourceCode/AgOpenGPS.Core/Streamers/Field/BoundaryStreamer.cs
J005t67: I was not aware of the more common signature. Anyway the caller will ignore the return value. Besides that, some of the AspectsStreamers can also return null as a valid result and that does not work very well with an out parameter.

SourceCode/AgOpenGPS.Core/Streamers/Field/BoundaryStreamer.cs
J005t67: During parsing also Overflow exceptions or IllegalFormat exceptions can happen and I don't want to check them all individually and probably forget 1 or 2. Also, I handle all exceptions gracefully by returning null. I don't think the caller is interested in any exceptions caused by the streamer code, so why not cath them all.

SourceCode/AgOpenGPS.Core/Streamers/Field/BoundaryStreamer.cs
J005t67: Yes, why not.

SourceCode/AgOpenGPS.Core/Streamers/Field/FieldStreamer.cs
J005t67: I will implement the getter, but nobody will propbably ever use it.

SourceCode/AgOpenGPS.Core/Streamers/Field/RecordedPathStreamer.cs
J005t67: Some of the original Read/Write functions in SaveOpen.Designer.cs use try catch blocks, some others don't.
I only implemented the TryRead and TryWrite if the original code used a try ctach block.  
==> Will remove TryWrite from RecordedPathStreamer.

SourceCode/AgOpenGPS.Core/Streamers/Field/RecordedPathStreamer.cs
J005t67: Yes, much better.

SourceCode/AgOpenGPS.Core/Streamers/Field/BoundaryStreamer.cs
FantasyTeddy: Okay, different question: is it really necessary, that theses classes even provide the `Try...` methods? For me the `try`-`catch` and the error handling seems to belong to the caller of those methods.
This would have several advantages:
- The AspectStreamers get simpler
- They do not need to know about logging or the presenter

SourceCode/AgOpenGPS.Core/Streamers/Field/BoundaryStreamer.cs
FantasyTeddy: I understand, but there are very valid reasons to consider catching the known exceptions: do you want to catch an `ArgumentException` or a `OutOfMemoryException`?
The former is a programming error that needs to be fixed and should not be hidden. The later is an exception that you probably cannot handle in any way and the user should know about it.

If you put yourself into the users perspective, what does he see when one of those unexpected exceptions occur? You tell him that the file is corrupt. That isn't even close to the real problem when it is not one of your expected exceptions.

We can leave the generic `catch (Exception)` (especially because it has been done already throughout the application) but I'm just saying that it is bad practice for a reason.

SourceCode/AgOpenGPS.Core/Streamers/Field/GeoStreamReader.cs
```diff
@@ -0,0 +1,112 @@
...
+        public bool PeekReadBool(out bool boolValue)
+        {
+            boolValue = false;
+            bool success = false;
+            int asciiCode = Peek();
+            if ('A' <= asciiCode && asciiCode <= 'Z')
+            {
+                boolValue = bool.Parse(ReadLine());
+            }
+            return success;
```
FantasyTeddy, line 109:
```suggestion
            int asciiCode = Peek();
            if ('A' <= asciiCode && asciiCode <= 'Z')
            {
                boolValue = bool.Parse(ReadLine());
                return true;
            }
            boolValue = false;
            return false;
```

SourceCode/AgOpenGPS.Core/Streamers/Field/RecordedPathStreamer.cs
FantasyTeddy: I think it is the wrong way around now, isn't it?

SourceCode/AgOpenGPS.Core/Streamers/Field/TramLinesStreamer.cs
```diff
@@ -37,8 +37,7 @@ public TramLines Read(string fieldPath)
                 using (GeoStreamReader reader = new GeoStreamReader(fullPath))
                 {
                     reader.ReadLine(); // skip header: $Tram
-
-                    if (-1 != reader.Peek())
+                    if (reader.Peek() == -1)
```
FantasyTeddy, line 40: Isn't the comparison wrong now?

SourceCode/AgOpenGPS.Core/Streamers/Field/RecordedPathStreamer.cs
J005t67: Yes, a bug. Will change TramLines similarly.

SourceCode/AgOpenGPS.Core/Streamers/Field/GeoStreamReader.cs
J005t67: OK.

SourceCode/AgOpenGPS.Core/Streamers/Field/TramLinesStreamer.cs
J005t67: Sorry

SourceCode/AgOpenGPS.Core/Streamers/Field/BoundaryStreamer.cs
J005t67: It is not really necessary of course. A different approach would be possible, but I don't want to bother the caller with various error states, like missing or corrupt files. So I prefer to do the logging here. Also this code will not only be used during openening and closing a field, but also in during editing of the Flags or recording paths, generation of tram lines and also during field selection or copying of fields, and I don't want to bother all these callers either.

SourceCode/AgOpenGPS.Core/Streamers/Field/GeoStreamReader.cs
J005t67: To me it makes sense. Exposing methods like ParseBool and ParseDouble and ParseInt helps to keep the formatting issues (like using CultureInfo.InvariantCulture) centralized. This
helps to avoid bugs caused by code that forgets this. Althouhg ParseBool does not need a Culture, it fits perfectly here and it is also used in for example RecordedPathStreamer

J005t67: The Parse... methods are so closele related to the Read.. methods that I want to keep them in the same file. I also want the Write and Formatting methods in the same file: GeoStreamWriter. If we want to reconsider this, we can better wait until we will implement the classes for KML-export and KML-import and see what approach fits best.

SourceCode/AgOpenGPS.Core/Streamers/Field/GeoStreamWriter.cs
J005t67: Sorry, I do not agree.

SourceCode/AgOpenGPS.Core/Streamers/Field/GeoStreamReader.cs
FantasyTeddy: Okay, I can see your point. However, can we consider making these methods `protected`? To me these are only internal methods that should not be provided to the user of this class or any sub-class.

SourceCode/AgOpenGPS.Core/Streamers/Field/GeoStreamWriter.cs
FantasyTeddy: Same discussion as for `Parse...`, so no need to apologize, I get why you want to do it that way.

SourceCode/AgOpenGPS.Core/Streamers/Field/RecordedPathStreamer.cs
```diff
@@ -0,0 +1,92 @@
...
+                            string[] words = line.Split(',');
+                            RecordedPoint point = new RecordedPoint(
+                                reader.ParseGeoCoordDir(words[0], words[1], words[2]),
+                                reader.ParseDouble(words[3]),
+                                reader.ParseBool(words[4]));
+                            recordedPath.Add(point);
+                        }
+                    }
+                }
+            }
```
FantasyTeddy, line 60:
```suggestion
            using (GeoStreamReader reader = new GeoStreamReader(fileAndDirectory))
            {
                reader.ReadLine(); // skip header
                int numPoints = reader.ReadInt();

                while (!reader.EndOfStream)
                {
                    for (int v = 0; v < numPoints; v++)
                    {
                        string line = reader.ReadLine();
                        string[] words = line.Split(',');
                        RecordedPoint point = new RecordedPoint(
                            reader.ParseGeoCoordDir(words[0], words[1], words[2]),
                            reader.ParseDouble(words[3]),
                            reader.ParseBool(words[4]));
                        recordedPath.Add(point);
                    }
                }
            }
```

J005t67: Will change

SourceCode/AgOpenGPS.Core/Streamers/Field/GeoStreamWriter.cs
J005t67: All four methods BoolString, IntString, DoubleString, and Wgs84String are used by the various *StreamWriters classes.
It is ok if these methods are public. The various *StreamWriter classes can not corrupt anything by calling these methods

FantasyTeddy: Oh my bad. Let's keep it that way then.

FantasyTeddy (approved):

PR #618, opened 2025-01-30
Add test project for `AgOpenGPS.Core` and a WPF example application

FantasyTeddy:
The test project uses the following suggested conventions:
- Test class follows the same folder structure as the tested class
- "Arrange", "Act", "Assert" sections (denoted by comments)
- Test method name: `<Tested Method>_Should<Description>`

The added WPF projects have the following intention:
- `AgOpenGPS.WpfApp`: used as a test application for WPF standalone
- `AgOpenGPS.WpfViews`: will contain the individual views that are converted from the WinForms application and is referenced by the test app and later by AgOpenGPS itself


J005t67 (approved):
Yes, exactly what we need now. I will replace the automatically generated content of the WPF projects later.

PR #623, opened 2025-02-01
New views and main layout for new WpfApp

J005t67:
This pull request contains the first views for the WPF Test app. The WPF test app will be a 'cheap' application that does not bother about translation, icons or perfect layout. The purpose of the app is to support and validate the refactoring towards the MVVM pattern. The user-interface will only give acess to the part of code/funtionality that has been refactored. As such it can also be used to show the progress of the refactoring process.

For people not familiar with WPF:
The layout of the MainWindow makes optimal use of the various  'layout containers' of WPF. Basically everything will layout itself without the need for additional code. For example the 'viewportGrid' is the last child of a DockPanel and will take all remaining space afte the buttonstrips have been placed and will update automtically when the visibility of one of the buttonstrip changes.
The 'viewportOverlay' inside the 'viewportGrid' also follows the automatic layout changes and no additional code is required to put the SimulatorBar in the correct position. 

Not in scope of this review:
- proper styling
- databinding
- all other panels and windows.
- all connections to other projects or other parts of the code.

### DIFF
--- a/SourceCode/AgOpenGPS.WpfApp/App.xaml.cs
+++ b/SourceCode/AgOpenGPS.WpfApp/App.xaml.cs
@@ -1,6 +1,6 @@
 ﻿using System.Windows;
 
-namespace AogWpf
+namespace AgOpenGPS.WpfApp
 {
     public partial class App : Application
     {
--- a/SourceCode/AgOpenGPS.WpfApp/App.xaml.cs
+++ b/SourceCode/AgOpenGPS.WpfApp/App.xaml.cs
@@ -1,14 +1,8 @@
-﻿using System.Configuration;
-using System.Data;
-using System.Windows;
+﻿using System.Windows;
 
-namespace AgOpenGPS.WpfApp
+namespace AogWpf
 {
-    /// <summary>
-    /// Interaction logic for App.xaml
-    /// </summary>
     public partial class App : Application
     {
     }
-
 }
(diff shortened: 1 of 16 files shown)


SourceCode/AgOpenGPS.WpfApp/App.xaml.cs
```diff
@@ -1,14 +1,8 @@
-﻿using System.Configuration;
-using System.Data;
-using System.Windows;
+﻿using System.Windows;
 
-namespace AgOpenGPS.WpfApp
+namespace AogWpf
```
FantasyTeddy, line 3: Do we really want to use a different namespace here? The other views still have the `AgOpenGPS.WpfApp` namespace (which I would prefer).

SourceCode/AgOpenGPS.WpfApp/ViewportViews/SimulatorBar.xaml
```diff
@@ -0,0 +1,19 @@
+﻿<UserControl x:Class="AgOpenGPS.WpfApp.ViewportViews.SimulatorBar"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008" mc:Ignorable="d"
+
+    d:DesignHeight="30" d:DesignWidth="800">
```
FantasyTeddy, line 7: I think for UserControls like this, the `DesignHeight` & `DesignWidth` can/should be removed, such that the content can define the size it needs.

SourceCode/AgOpenGPS.WpfApp/App.xaml.cs
J005t67: No, we don't want that different namespce. It is the old namespace that I used in old repository.

SourceCode/AgOpenGPS.WpfApp/ViewportViews/SimulatorBar.xaml
J005t67: Yes, and removed a few more.

FantasyTeddy (approved):

PR #624, opened 2025-02-01
Create or add multiple flags from text file

Rob2029:
Ive added a button in Load flags by lat long that loads a txt or csv file containing latitude and longitude values.
The format of the text file has to be "latitude,longitude,flagColor,flagName" for example:

lat,lon,color,name
53.4370564,-111.160047,0,pointA
53.4380564,-111.160047,0,
53.4390564,-111.160047,0,pointC

The flagName column is optional, but it must be present in the header, otherwise it will not create the flags
The header will trigger a warning message about that line being invalid
I think it needs a method to skip the header if there is one
Also in the future it can have a method to read the existing flags and avoid saving new flagss on top of old flags


SourceCode/GPS/Forms/Field/FormEnterFlag.cs
```diff
@@ -101,5 +103,63 @@ private void btnRed_Click(object sender, EventArgs e)
             Close();
 
         }
+        // This loads flags of a txt or CSV file with this format: "latitude,longitude,flagColor,flagName"
+        private void btnLoad_Click(object sender, EventArgs e)
+        {
```
J005t67, line 108: Unfortunately, the Designer files in AgOpenGPS contain a lot of functionality, that is too closely connected to the user-interface. This really harms maintainability and testability Also we started refactoring to (slowly) introduce the MVVM-pattern. (Please look it up if you like). All code that is currently in the designer files needs to be moved and redesigned. This is needed to move away from WinFroms to WPF, or Avalonia. So adding more code to Designer files is really not what we want. Sorry, if this feels unfair. You are probably the first person that receives this comment, but we have to stop doing this. Another problem at the moment is that we even don't have code in AgOpenGps, that shows how it can be done properly. My advice is to wait a month or so, unitl all Flags functionality has been refactored and make your code fit in the new MVVM-pattern

J005t67 (changes requested):

SourceCode/GPS/Forms/Field/FormEnterFlag.cs
Rob2029: No worries thanks for the explanation, I'll read about MVVM

FantasyTeddy: @J005t67: I don't think that it is fair to expect from contributors, that they adhere to a new pattern that isn't yet established in the project (even though we are slowly getting there).
I would suggest to fix the other issues and then we merge this PR, since it isn't worse than most of the other code in AgOpenGPS.

Rob2029: Thanks for the suggestion, there is a problem with this that I did not see when doing the PR, it only loads the flags properly when the language is english
So when I have the time I[ll make another PR with the feature working properly

FantasyTeddy: Okay, I will close this PR.

PR #625, opened 2025-02-02
doneforreal

BabtaiRTK:


richardklasens:
Hello, 

Thanks for your contribution.
please describe what your PR is doing,

also, there are mergers conflicts that need to be solved. Thank you

Pat-I:
Hi BabtaiRTK
I think you should make at least 3 PR?
I see one for the new JCB color?(Which I would like to have)
One for DadMode (simplified interface)
One for the  smaller GUI with space on the right side

What is the development team opinion about this?
Also I think all the coding should be done in English?

FantasyTeddy:
> What is the development team opinion about this? Also I think all the coding should be done in English?

Yes, I think splitting these changes up into separate PRs would make it easier to discuss the various things separately.

However, there are a few issues that need to be addressed regardless:
- Disabling the initial `Form_First` with the terms and conditions is an absolute no-no.
- Making the main viewport smaller without any option to disable this is also not allowable.
- Why is the `DrawBeta()` method disabled?

I am also not too much into the "Dad Mode" idea, because even though I agree that the user interface could be improved in terms of accessibility, I don't like the idea of hiding settings behind some kind of switch. @AgOpenGPS-Official/aog-coders: What do you think about that?

richardklasens:
I’m sorry but this can’t be in the official release, disabling the terms accept display and removing the beta string is strictly against the community guidelines. I will close this PR and ask you to redo your changes in separate PR’s without the unnecessary parts like mentioned here above. 
@AgOpenGPS-Official/aog-coders PR Closed

PR #634, opened 2025-02-06
Correct typo in filename Wgs84.cs and little bit more.

J005t67:


SourceCode/AgOpenGPS.Core/Models/Base/Wgs84.cs
```diff
@@ -0,0 +1,34 @@
...
+            Latitude = latitude;
+            Longitude = longitude;
+        }
+
+        public double Latitude { get; }
+        public double Longitude { get; }
+
+        public double Distance(Wgs84 b) // In meters
+        {
+            const double EarthRadius = 6376500.0;
```
FantasyTeddy, line 19: Is this approximation good enough? [Wikipedia](https://en.wikipedia.org/wiki/World_Geodetic_System#WGS84) has different numbers (however, I guess the full algorithm is more complex than assuming the earth is a sphere).

SourceCode/AgOpenGPS.Core.Tests/Models/Base/Wgs84Tests.cs
```diff
@@ -0,0 +1,44 @@
...
+            Wgs84 amsterdam = new Wgs84(52.377956, 4.897070);
+            Wgs84 denDungen = new Wgs84(51.665, 5.37222);
+
+            // Act
+            double distance = amsterdam.Distance(denDungen);
+            double distanceBack = denDungen.Distance(amsterdam);
+
+            // Assert
+            Assert.That(distance, Is.GreaterThan(80000));
+            Assert.That(distance, Is.LessThan(90000));
```
FantasyTeddy, line 21: Is this the precision that the algorithm provides? +/- 10 km ?

SourceCode/AgOpenGPS.Core/Models/Base/Wgs84.cs
J005t67: I took the algorithm and the number from the existing code. I will make the number more accurate, but leave the algorithm unchanged. Also added a test cases that is closer to the actual use. (Local distances)

SourceCode/AgOpenGPS.Core.Tests/Models/Base/Wgs84Tests.cs
J005t67: No, this was the precision that I provided, the algorithm is more accurate. ;-) Made the test more accurate.

SourceCode/AgOpenGPS.Core.Tests/Models/Base/Wgs84Tests.cs
```diff
@@ -0,0 +1,69 @@
...
+
+            Assert.That(distance, Is.EqualTo(distanceBack));
+        }
+
+        [Test]
+        public void Test_LongDistance()
+        {
+            // Arrange
+            Wgs84 amsterdam = new Wgs84(52.377956, 4.897070);
+            Wgs84 opposite = new Wgs84(-52.377956, 4.897070 + 180);
```
FantasyTeddy, line 32: Is this a valid value to be used? Shouldn't the longitude be between -180 and 180 ?

FantasyTeddy (approved):

SourceCode/AgOpenGPS.Core.Tests/Models/Base/Wgs84Tests.cs
J005t67: Changed

FantasyTeddy (approved):

PR #647, opened 2025-02-10
Add `VehicleConfig` model

FantasyTeddy:
This adds a `VehicleConfig` model that contains (or will contain) all information related to a vehicles configuration/settings.

In contrast to the existing `CVehicle` class, the new `VehicleConfig` class does not contain any particular logic and most importantly it does not contain a reference to the `FormGPS` class.


SourceCode/GPS/Classes/CVehicle.cs
```diff
@@ -314,7 +319,7 @@ public void DrawVehicle()
                     GL.Disable(EnableCap.Texture2D);
                     //GL.Disable(EnableCap.Blend);
                 }
-                else if (vehicleType == 1) //Harvestor
+                else if (Vehicle.Type == VehicleType.Harvestor)
```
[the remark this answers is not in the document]
FantasyTeddy: Thank you, I just copied the name from the comment, but of course I will rename it to `Harvester`.

Yes you are absolutely right, "4WD" looks way better, but unfortunately it is not a valid identifier in C# (it cannot start with a number).
What do you think if we would rename it to `Articulated`? Because according to my (limited) research, that is what actually distinguishes that kind of tractor from the rest (not that it has 4-wheel-drive).

lansalot:
Yes articulated is better as that's the type of machine, rather than it
being 4wd (or course many conventional tractors are 4wd...)

FantasyTeddy:
@lansalot: I changed it according to your suggestions. Thanks for the inputs.

(I'm debating now, whether or not I should also change all other occurrences of "4WD" and "FourWD" to "Articulated"...)

lansalot:
Consistency would be good, I'd go with articulated

FantasyTeddy:
Okay, I did the renaming.

SourceCode/GPS/Forms/Settings/ConfigVehicle.Designer.cs
```diff
@@ -632,25 +632,25 @@ private void SetOpacity()
...
-            mf.vehicleOpacity = Math.Max(mf.vehicleOpacity - 0.2, 0.2);
+            mf.vehicle.Vehicle.Opacity = Math.Max(mf.vehicle.Vehicle.Opacity - 0.2, 0.2);
             OpacityChanged();
         }
 
         private void OpacityChanged()
         {
-            lblOpacityPercent.Text = ((int)(mf.vehicleOpacity * 100)).ToString() + "%";
-            Properties.Settings.Default.setDisplay_vehicleOpacity = (int)(mf.vehicleOpacity * 100);
+            lblOpacityPercent.Text = ((int)(mf.vehicle.Vehicle.Opacity * 100)).ToString() + "%";
```
J005t67, line 653: Please ignore this comment, but this is incorrect in Turkish. In Turkish, the %-sign is before the number ;-)

SourceCode/AgOpenGPS.Core/Models/Vehicle/Vehicle.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace AgOpenGPS.Core.Models
+{
+    public enum VehicleType
+    {
+        Tractor = 0,
+        Harvester = 1,
+        Articulated = 2
+    }
+
+    public class Vehicle
```
J005t67, line 10: I have the feeling that we need this class name later for a more  high level concept. Isn't VehicleSettings more appropriate?

SourceCode/AgOpenGPS.Core/Models/Vehicle/Vehicle.cs
```diff
@@ -0,0 +1,14 @@
...
+    {
+        public VehicleType Type { get; set; }
```
J005t67, line 12: I understand that we need this enum type to read and write the settings, but shouldn't we use subclasses of Vehicle to differentiate between the type?

SourceCode/AgOpenGPS.Core/Models/Vehicle/Vehicle.cs
FantasyTeddy: Yes I was debating as well, if this will just be the configuration part of a vehicle. Would `VehicleConfig` also be okay?

J005t67: Yes, I like VehicleConfig even better.

SourceCode/AgOpenGPS.Core/Models/Vehicle/Vehicle.cs
J005t67: I think that the VehicleConfig class is Ok as it is now. The idea for subclasses is more applicable to the more high level Vehicle class, because that will do type specific rendering and guidance. Whatever it will be, it is ok with me to decide later, when we know more.

J005t67 (approved):

PR #651, opened 2025-02-12
Refactor: new class Texture to simplify all Texture operations

J005t67:
This new class Texture helps to simplify all texture related oparations. 

It does not replace all existing texture code yet. That will happen in another pull request
The class Texture needs to be extended with at least the following to replace all existing texture code:
- support textures that are rendered on a Z-coord.
- support rendering of part of textures.
- support updating the bitmap of a texture.

### DIFF
--- a/SourceCode/AgOpenGPS.Core/Drawing/Texture.cs
+++ b/SourceCode/AgOpenGPS.Core/Drawing/Texture.cs
@@ -1,0 +1,69 @@
+﻿using System.Drawing;
+using System.Drawing.Imaging;
+
+using OpenTK.Graphics.OpenGL;
+
+namespace AgOpenGPS.Core.Drawing
+{
+    public class Texture2D
+    {
+
+        private readonly int _textureId;
+        public Texture2D(Bitmap bitmap)
+        {
+            GL.GenTextures(1, out _textureId);
+            if (bitmap != null) SetBitmap(bitmap);
+        }
+
+        public void Bind()
+        {
+            GL.BindTexture(TextureTarget.Texture2D, _textureId);
+        }
+
+        public void Draw(double x0, double x1, double y0, double y1)
+        {
+            GL.Enable(EnableCap.Texture2D);
+
+            Bind();
+            GL.Begin(PrimitiveType.Quads);
+            GL.TexCoord2(0, 0); GL.Vertex2(x0, y0);
+            GL.TexCoord2(1, 0); GL.Vertex2(x1, y0);
+            GL.TexCoord2(1, 1); GL.Vertex2(x1, y1);
+            GL.TexCoord2(0, 1); GL.Vertex2(x0, y1);
+            GL.End();
+            GL.Disable(EnableCap.Texture2D);
+        }
+
+        public void DrawCentered(
+            double halfSizeX,  // Pass minus hafSizeX to flip positive X-axis to negative X-axis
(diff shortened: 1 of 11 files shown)


SourceCode/AgOpenGPS.Core/Drawing/Texture.cs
```diff
@@ -0,0 +1,69 @@
...
+namespace AgOpenGPS.Core.Drawing
+{
+    public class Texture2D
+    {
+
+        private readonly int _textureId;
+        public Texture2D(Bitmap bitmap)
+        {
+            GL.GenTextures(1, out _textureId);
+            if (bitmap != null) SetBitmap(bitmap);
```
FantasyTeddy, line 15: Is it valid to pass `null` as a bitmap?

SourceCode/AgOpenGPS.Core/Drawing/Texture.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        public void Bind()
+        {
+            GL.BindTexture(TextureTarget.Texture2D, _textureId);
+        }
+
+        public void Draw(double x0, double x1, double y0, double y1)
```
FantasyTeddy, line 23: I always find it quite hard to figure out what the arguments mean in these kinds of methods. What do you think if we would use [`Point`](https://learn.microsoft.com/en-us/dotnet/api/system.windows.point) and [`Size`](https://learn.microsoft.com/en-us/dotnet/api/system.windows.size) here?

SourceCode/AgOpenGPS.Core/Drawing/Texture.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        public void DrawCentered(
+            double halfSizeX,  // Pass minus hafSizeX to flip positive X-axis to negative X-axis
+            double halfSizeY)  // Pass minus hafSizeXYto flip positive Y-axis to negative Y-axis
+        {
+            Draw(-halfSizeX, halfSizeX, -halfSizeY, halfSizeY);
+        }
+
+        public void SetBitmap(Bitmap bitmap)
```
FantasyTeddy, line 44:
```suggestion
        private void SetBitmap(Bitmap bitmap)
```

SourceCode/AgOpenGPS.Core/Drawing/Texture.cs
J005t67: Removed.  (It was a remainder of my first attempt to implement 'SetBitmap'

SourceCode/AgOpenGPS.Core/Drawing/Texture.cs
J005t67: I find it rather funny that I wrote a function that I dislike myself and that I need you to find out ;-) I know, too many arguments in an arbitrary order is error prone. Especially if it's all doubles. 

Introduced XyCoord and XyDelta to improve this.  I do not like Point a lot, because it does not show wether it is 2D or 3D. And it adds a dependency more to Windows. Also XyCoord, and XyDelta matches nicely with the already existing GeoCoord and GeoDelta.

SourceCode/AgOpenGPS.Core/Drawing/Texture.cs
J005t67: Made private.

FantasyTeddy (approved):

PR #653, opened 2025-02-15
Add more Field Aspect Streamers

J005t67:


SourceCode/AgOpenGPS.Core/Streamers/Field/GeoStreamWriter.cs
```diff
@@ -53,6 +53,11 @@ public string GeoCoordDirStringENH(GeoCoord coord, GeoDir heading, string coordF
                 HeadingStringH(heading, headingFormatString);
         }
 
+        public void WriteDateTime()
+        {
+            WriteLine(DateTime.Now.ToString("yyyy-MMMM-dd hh:mm:ss tt", CultureInfo.InvariantCulture));
```
FantasyTeddy, line 58: With testability in mind, we should maybe think of a way to inject the current date/time somehow, instead of using `DateTime.Now` directly?

J005t67: I'm more worried about the total absence of test code for the streamers so far. I don't care so much about the date time field in the Overview in the files 'Field.txt'. I think it is not even used in the code.

I could for example write tests that write a certain aspect to file and read it back and check that nothing was changed.

FantasyTeddy: Yes, I think a roundtrip test would be great.
However, I'm not sure if actually checking the output (Field.txt) would not also make sense.

J005t67: I prefer the idea of 'write and read back roundtrip' because it does not need any additional data files for testing. Adding more test will not add much coverage I think.

PR #657, opened 2025-02-20
Use DirectoryInfo in stead of plain strings

J005t67:


SourceCode/AgOpenGPS.Core/Streamers/Field/FieldAspectStreamer.cs
```diff
@@ -20,17 +20,14 @@ public void SetPresenter(IFieldStreamerPresenter presenter)
...
         }
 
-        public void CreateDirectory(string fieldPath)
+        public void CreateDirectory(DirectoryInfo fieldDirectory)
         {
-            if (0 < fieldPath.Length && !Directory.Exists(fieldPath))
-            {
-                Directory.CreateDirectory(fieldPath);
-            }
+            fieldDirectory.Create();
```
FantasyTeddy, line 30: Is it even necessary to have an extra method for that if it is now only one line?

FantasyTeddy (commented):
By the way: there is also the concept of [`FileInfo`](https://learn.microsoft.com/en-us/dotnet/api/system.io.fileinfo) if you want to go a step further.

SourceCode/AgOpenGPS.Core/Streamers/Field/FieldAspectStreamer.cs
J005t67: Removed.

FantasyTeddy (approved):

PR #658, opened 2025-02-20
Refactor/improve cnmea

J005t67:


SourceCode/GPS/Classes/CFlag.cs
```diff
@@ -33,5 +34,7 @@ public CFlag(double _lati, double _longi, double _easting, double _northing, dou
             ID = _ID;
             notes = _notes;
         }
+
+        public GeoCoord AsGeoCoord => new GeoCoord(northing, easting);
```
FantasyTeddy, line 38: Microsoft makes a clear distinction between `As...` methods and `To...` methods:
- `To...` methods create a new type (allocate memory) -> example: [`ToList`](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.tolist)
- `As...` methods act more like down-casts (no allocation) -> example: [`AsEnumerable`](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.asenumerable)

This property (which should maybe be a method?) is of the first category and should probably be called `ToGeoCoord`.

Alternatively, if you plan for the `GeoCoord` to replace the current `northing` and `easting` fields, the name `GeoCoord` (without prefix) could make sense, to avoid future renamings.

SourceCode/GPS/Classes/CNMEA.cs
```diff
@@ -82,17 +87,24 @@ public void ConvertWGS84ToLocal(double Lat, double Lon, out double Northing, out
             //Easting += mf.RandomNumber(-0.02, 0.02);
         }
 
-        public void ConvertLocalToWGS84(double Northing, double Easting, out double Lat, out double Lon)
+        public Wgs84 ConvertGeoCoordToWgs84(GeoCoord geoCoord)
         {
-            Lat = ((Northing + fixOffset.northing) / mPerDegreeLat) + latStart;
-            mPerDegreeLon = 111412.84 * Math.Cos(Lat * 0.01745329251994329576923690766743) - 93.5 * Math.Cos(3.0 * Lat * 0.01745329251994329576923690766743) + 0.118 * Math.Cos(5.0 * Lat * 0.01745329251994329576923690766743);
-            Lon = ((Easting + fixOffset.easting) / mPerDegreeLon) + lonStart;
+            latitude = ((geoCoord.Northing + fixOffset.northing) / mPerDegreeLat) + latStart;
```
FantasyTeddy, line 92: Why are these values stored in a class field?

J005t67: Oops, a bug

SourceCode/GPS/Classes/CFlag.cs
J005t67: In this particular case, I replaced northing and easting with GeoCoord.

FantasyTeddy: Okay, then maybe `GeoCoord` would be a better name for the property?

What about the other occurrences of `AsGeoCoord`?

FantasyTeddy: Nice. Could we make `ToGeoCoord` a method instead of a property? It makes it more obvious, that it is not just something that gets returned, but there is something happening (allocation).

FantasyTeddy (approved):

PR #662, opened 2025-02-21
Variable textures for Vehicle

J005t67:


SourceCode/GPS/Classes/VehicleTextures.cs
```diff
@@ -14,6 +19,42 @@ public VehicleTextures()
...
+                if (_tractor == null) _tractor = new Texture2D(null);
+                return _tractor;
+            }
+        }
+
+        public Texture2D Harvester
+        {
+            get
+            {
+                if (_harvester == null) _harvester = new Texture2D(Resources.z_HarvesterAoG);
```
FantasyTeddy, line 35: Why does the harvester have a default texture and the others don't?

J005t67: Sorry, this was a left over from a debug session.

FantasyTeddy (approved):

PR #663, opened 2025-02-21
New Feature: Skipping already worked Tracks (AB Lines/Curves) (new SkipMode)

steran82:
https://discourse.agopengps.com/t/youturn-skipping-worked-lines/17743


SourceCode/GPS/Classes/CYouTurn.cs
```diff
@@ -25,6 +27,7 @@ public class CYouTurn
         public int rowSkipsWidth = 1, uTurnSmoothing;
 
         public bool alternateSkips = false, previousBigSkip = true;
```
FantasyTeddy, line 29: Shouldn't it be possible to remove `alternateSkips` now?

SourceCode/GPS/Classes/CYouTurn.cs
```diff
@@ -25,6 +27,7 @@ public class CYouTurn
...
+        public SkipMode skip_mode = SkipMode.Normal;
```
FantasyTeddy, line 30:
```suggestion
        public SkipMode skipMode = SkipMode.Normal;
```

SourceCode/GPS/Classes/CYouTurn.cs
```diff
@@ -102,9 +105,56 @@ public CYouTurn(FormGPS _f)
             uTurnSmoothing = Properties.Settings.Default.setAS_uTurnSmoothing;
         }
 
+        //find next not worked lane after the defined lanes to skip
+        private int GetNextNotWorkedTrack(bool isTurnLeft, int rowSkipsWidth, bool isAB)
+        {
+            double goalLane;
+
+            if (isAB)
```
FantasyTeddy, line 113: Can we do one of these options:
1. Either use two different methods for "AB" and "not AB"? (Because the `if`-`else` are distinct)
2. Or rewrite this method to share the common code (the only difference is the use of `mf.ABLine` vs. `mf.curve`, which should be possible to rewrite with a temporary variable)

SourceCode/GPS/Forms/Controls.Designer.cs
```diff
@@ -1753,25 +1753,52 @@ private void btnChangeMappingColor_Click(object sender, EventArgs e)
...
-                    cboxpRowWidth.Text = "1";
-                }
-                yt.Set_Alternate_skips();
-                yt.ResetCreatedYouTurn();
-                if (!yt.isYouTurnBtnOn) btnAutoYouTurn.PerformClick();
-            }
-            else
+
+            yt.skip_mode++;
+            if ((int)yt.skip_mode > 2) yt.skip_mode = SkipMode.Normal;
```
FantasyTeddy, line 1758: Can we do this with a switch statement that explicitly converts from one mode to the next?

```suggestion
            switch (yt.skip_mode)
            {
                case SkipMode.Normal:
                    yt.skip_mode = SkipMode.Alternative;
                    break;
                case SkipMode.Alternative:
                    yt.skip_mode = SkipMode.ignoreWorkedTracks;
                    break;
                case SkipMode.ignoreWorkedTracks:
                    yt.skip_mode = SkipMode.Normal;
                    break;
            }
```

Even though this is more code, I think this would be more readable. (Maybe move it into a separate method `NextSkipMode`, that way it would become even more understandable.)

SourceCode/GPS/Classes/CYouTurn.cs
```diff
@@ -102,9 +105,56 @@ public CYouTurn(FormGPS _f)
...
+            {
+
+                if ((isTurnLeft && !mf.ABLine.isHeadingSameWay) || (!isTurnLeft && mf.ABLine.isHeadingSameWay))
```
J005t67, line 116: One line if and else blocks are error prone if no curly braces are present. One option is to add curly braces. (You need 4 pair for the if-blocks and 4 pair for the else-blocks.) Unfortunately, that would add another 16 lines of code to the code that is already duplicate. (And believe me, duplicate code takes time to read and maintain.)
I would prefer to test this condition oly once, and write one (a bit larger) if-block and one (a bit larger) else-block

SourceCode/GPS/Forms/Controls.Designer.cs
```diff
@@ -1939,6 +1966,12 @@ private void toolStripAreYouSure_Click(object sender, EventArgs e)
                         }
                         patchSaveList?.Clear();
 
+                        //delete all worked Lanes too
+                        foreach (CTrk trkitm in trk.gArr)
```
J005t67, line 1970: Please use CamelCase. Please do not abbreviate. => trackItem

SourceCode/GPS/Properties/Resources.resx
```diff
@@ -1429,4 +1429,7 @@
   <data name="ABPivotCircle" type="System.Resources.ResXFileRef, System.Windows.Forms">
     <value>..\btnImages\ABPivotCircle.png;System.Drawing.Bitmap, System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a</value>
   </data>
+  <data name="YouSkipWT" type="System.Resources.ResXFileRef, System.Windows.Forms">
```
J005t67, line 1432: Do not use abbreviations in the file name. (It took me a minute before I figured out that WT means WorkedT.)

J005t67, line 1432: Why is it spelled 'You' if the word directly refers to the shape of an 'U'? Ah, OK, I see, most of the others also use 'YouTurn'. In that case I prefer the prefix 'YouTurn'

SourceCode/GPS/Classes/CYouTurn.cs
steran82: i would prefer the second option - but they don't have a common interface, so it's not possible to assign it dynamically, isn't it?

SourceCode/GPS/Properties/Resources.Designer.cs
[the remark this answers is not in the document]
steran82: i use visual studio 2022; is there a setting to avoid this in future?

SourceCode/GPS/Classes/CYouTurn.cs
steran82: yeah - now we can use current skipMode; i'll adjust  if (skipMode == SkipMode.Alternative && rowSkipsWidth2 > 1) and delete the bool alternateSkips

SourceCode/GPS/Classes/CYouTurn.cs
steran82: pls see https://github.com/AgOpenGPS-Official/AgOpenGPS/pull/663#discussion_r1966454194
currently i have no good idea to prevent duplicate code without bigger refactoring (common interface for ABLine and Curve Classes

SourceCode/GPS/Forms/Controls.Designer.cs
steran82: Thanks - will do it

SourceCode/GPS/Properties/Resources.resx
steran82: You are right - i'll adjust it

SourceCode/GPS/Forms/Controls.Designer.cs
steran82: Thanks - i'll adjust and create a new suggestion

SourceCode/GPS/Resources/YouSkipWT.png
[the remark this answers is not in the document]
steran82: Thanks - you are right. VS Resource Manager and me are not friends so far :-P
I deleted it

SourceCode/GPS/Classes/CYouTurn.cs
J005t67: OK, better not start the bigger refactoring now.

SourceCode/GPS/Classes/CYouTurn.cs
FantasyTeddy: The field is still there. Are you going to remove it?

FantasyTeddy (approved):

FantasyTeddy:
Thank you for the contribution ❤️

PR #666, opened 2025-02-23
Refactor/geotexture

J005t67:
Replaced the last textures.

Could someone help me? And test the BingGrid (background map) functionality. I was not able to do that.


FantasyTeddy (approved):
I quickly looked if the Bing map works, and it seems fine to me.

PR #667, opened 2025-02-24
Refactor Brand related bitmap resources

J005t67:
Refactoring goes best bottom up, so I will start moving resources from project AgOpenGPS to project AgOpenGPS.Core.
This will make the resources that are currently only accessible by the Winform application also accessible for others.

This first 'batch' is only about the brand specific bitmaps. As you can see, my intention is to split the very large Resource.resx file
into more manageble chunks.


SourceCode/AgOpenGPS.Core/Resources/BrandBitmaps.resx
FantasyTeddy, line 1: Do you have a particular reason, why you named this "Bitmaps"? Wouldn't "Images" be better suited here?

SourceCode/AgOpenGPS.Core/Models/Vehicle/Brands.cs
```diff
@@ -0,0 +1,108 @@
...
+
+namespace AgOpenGPS.Core.Models
+{
+    public enum TractorBrand { AGOpenGPS, Case, Claas, Deutz, Fendt, JohnDeere, Kubota, Massey, NewHolland, Same, Steyr, Ursus, Valtra, JCB }
+    public enum HarvesterBrand { AgOpenGPS, Case, Claas, JohnDeere, NewHolland }
+    public enum ArticulatedBrand { AgOpenGPS, Case, Challenger, JohnDeere, NewHolland, Holder }
+
+    public static class TractorBitmaps
+    {
+        public static Bitmap GetBitmap(TractorBrand brand)
```
FantasyTeddy, line 12: Maybe we could even make these extension methods?

SourceCode/AgOpenGPS.Core/Resources/Brands/Brand/BrandTriangleVehicle.png
FantasyTeddy, line 1: Is this really in the correct place now? Even though the triangle is used in the selection form, it is not really a brand itself and should probably be located somewhere else (maybe for the time being just leave it where it was?)

SourceCode/AgOpenGPS.Core/Models/Vehicle/Brands.cs
J005t67: One day, this code fragment wil become part of TractorSelectionViewModel. That does not need to take too long.
Once the top level ApplicationViewModel is in place, we can add a ConfigViewModel to ApplicationViewModel, and ConfigVehicleViewModel to ConfigViewModel and  put it there.

SourceCode/AgOpenGPS.Core/Resources/BrandBitmaps.resx
J005t67: Wel , my reason was that we put the Images in a Bitmap after reading, but I agree BrandImages is more appropriate.

SourceCode/AgOpenGPS.Core/Resources/Brands/Brand/BrandTriangleVehicle.png
J005t67: I agree that this one is the odd one out and it should go somewhere else one day. But for now, I just moved over the whole directory structure with brand related png's, (and the original directories will not longer exist after this PR.)

FantasyTeddy (dismissed):

FantasyTeddy (approved):

SourceCode/AgOpenGPS.Core/AgOpenGPS.Core.csproj
```diff
@@ -4,10 +4,12 @@
     <TargetFramework>net48</TargetFramework>
     <OutputType>Library</OutputType>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+    <GenerateResourceUsePreserializedResources>true</GenerateResourceUsePreserializedResources>
```
J005t67, line 7: Thanks a lot for fixing the build! Although I read about your solution on StackTrace, I was not sure how to do it, because I didn't know how to reproduce the problem. I still wonder, why do the other porjects build nicely, without the settings you added?

FantasyTeddy: They actually don't when building with `dotnet build`.
I think this should be done next, to use the .NET Cli to build instead of MSBuild.

PR #668, opened 2025-02-24
Patches

richardklasens:
- ISOXML fixes for Fendt filing (Andy)
- text in Tramlinebuilder when swapping lines
- reorder text alignments and positions for Config and Steer Forms
- Disable icons when No GPS Sign is spinning

### DIFF
--- a/SourceCode/GPS/Forms/FormGPS.cs
+++ b/SourceCode/GPS/Forms/FormGPS.cs
@@ -493,11 +493,6 @@ namespace AgOpenGPS
             }
         }
 
-        private void toolStripDropDownButton1_Click(object sender, EventArgs e)
-        {
-
-        }
-
         private void FormGPS_FormClosing(object sender, FormClosingEventArgs e)
         {
             Form f = Application.OpenForms["FormGPSData"];
--- a/SourceCode/GPS/Forms/FormGPS.cs
+++ b/SourceCode/GPS/Forms/FormGPS.cs
@@ -493,6 +493,10 @@ namespace AgOpenGPS
             }
         }
 
+        private void toolStripDropDownButton1_Click(object sender, EventArgs e)
+        {
+
+        }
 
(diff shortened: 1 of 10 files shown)


SourceCode/GPS/Forms/FormGPS.cs
```diff
@@ -493,6 +493,10 @@ private void FormGPS_Load(object sender, EventArgs e)
             }
         }
 
+        private void toolStripDropDownButton1_Click(object sender, EventArgs e)
```
FantasyTeddy, line 496: Is this method needed?

SourceCode/GPS/Forms/GUI.Designer.cs
```diff
@@ -104,8 +104,19 @@ private void tmrWatchdog_tick(object sender, EventArgs e)
             if (++sentenceCounter > 20)
             {
                 ShowNoGPSWarning();
+                //make sure settings and others can't be openend, the program is in standby
```
FantasyTeddy, line 107: This seems to be a very drastic solution for a relatively small issue. Is there really no other way to solve this?

richardklasens: Yes but it won’t harm anyone as well. Although what are the disadvantages? Load or save settings won’t be good either

SourceCode/GPS/Forms/FormGPS.cs
richardklasens: No I fix it

SourceCode/GPS/Forms/GUI.Designer.cs
FantasyTeddy: It just seems like a quite random requirement to have GPS (or Simulation) in order to be able to change literally anything.
I fear that this needs to be explained a lot (without a satisfying answer).

FantasyTeddy (approved):

PR #674, opened 2025-02-28
Let CNMEA use baseclasses more

J005t67:
I'am working towards the first ViewModels for opening Fields. The tables that show the fields contain a column with the distance.
To compute these distances, I need the current location'. The current location is stored in the two separate variables 'latitiude' and 'longitude'
This PR only converts the two separate values for latitude and logitude, to a more maintainable and meaningfull variable named CurrentLocation
The next step will be that I move 'CurrentLocation' to the Core project, to make it accessible for future non-winform applications.

### DIFF
--- a/SourceCode/AgOpenGPS.Core/Models/Base/Wgs84.cs
+++ b/SourceCode/AgOpenGPS.Core/Models/Base/Wgs84.cs
@@ -5,6 +5,7 @@ namespace AgOpenGPS.Core.Models
     // Represents a coordinate in the World Geodetic System 1984
     public struct Wgs84
     {
+        const double EarthRadiusInMeters = 6371 * 1000.0;
         public Wgs84(double latitude, double longitude)
         {
             Latitude = latitude;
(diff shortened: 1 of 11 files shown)


SourceCode/AgOpenGPS.Core/Models/Base/Wgs84.cs
```diff
@@ -34,6 +34,22 @@ public double DistanceInKiloMeters(Wgs84 b)
         {
             return 0.001 * DistanceInMeters(b);
         }
+
+        public Wgs84 CalculateNewPostionFromBearingDistance(double bearing, double distanceInMeters)
+        {
+            const double degreesToRadians = Math.PI / 180.0;
+            const double radiansToDegrees = 1.0 / degreesToRadians;
```
FantasyTeddy, line 41: I feel like these deserve their own method somewhere, what do you think?

FantasyTeddy (commented):
I really like how everything starts to look :)

SourceCode/AgOpenGPS.Core/Models/Base/Wgs84.cs
```diff
@@ -34,6 +34,22 @@ public double DistanceInKiloMeters(Wgs84 b)
...
+            double latRadians = Latitude * degreesToRadians;
+            double lonRadians = Longitude * degreesToRadians;
+
+            double R = distanceInMeters / (6371.0 * 1000);
```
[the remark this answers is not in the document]
J005t67: Yes, indeed. I have put the new constant in this file, because probably nobody else needs it.

SourceCode/AgOpenGPS.Core/Models/Base/Wgs84.cs
J005t67: A method is more explicit and leaves absolutely no doubt how to use it. I will change it to methods. For the moment I will keep it in the same file, because I don't know yet what the best place is.

FantasyTeddy (approved):

PR #680, opened 2025-03-02
Move currentLocation to the Core

J005t67:


SourceCode/AgOpenGPS.Core/Models/ApplicationModel.cs
```diff
@@ -28,5 +28,13 @@ public void SetPresenter(IApplicationPresenter applicationPresenter)
 
         public FieldStreamer FieldStreamer => Fields.FieldStreamer;
 
+        public Wgs84 CurrentLatLon { get; set; }
+        public double latitude => CurrentLatLon.Latitude;
+        public double longitude => CurrentLatLon.Longitude;
```
FantasyTeddy, line 33: If we expose these as properties, shouldn't they follow the same naming convention: `Latitude` & `Longitude`?

SourceCode/AgOpenGPS.Core/Models/ApplicationModel.cs
```diff
@@ -28,5 +28,13 @@ public void SetPresenter(IApplicationPresenter applicationPresenter)
...
+
+        //local plane geometry
+        public Wgs84 StartLatLon { get; set; }
+        public double latStart => StartLatLon.Latitude;
+        public double lonStart => StartLatLon.Longitude;
```
FantasyTeddy, line 38: I think it would be better to turn these around and don't abbreviate: `StartLatitude` & `StartLongitude`.

SourceCode/GPS/Forms/Field/FormMap.cs
```diff
@@ -150,41 +141,24 @@ private void btnGo_Click(object sender, EventArgs e)
...
-                    mapControl.Center = new GeoPoint((float)mf.pn.longitude, (float)mf.pn.latitude);
-
-                    UpdateWindowTitle();
-                    mapControl.Invalidate();
-                }
+                mapControl.Markers.Clear();
             }
-            else
+            mapControl.Center = new GeoPoint((float)mf.AppModel.longitude, (float)mf.AppModel.latitude);
+            if (mapControl.Markers.Count == 0)
```
FantasyTeddy, line 147: Doesn't this `if` also need to check for `bingLine.Count == 0` ?

SourceCode/AgOpenGPS.Core/Models/ApplicationModel.cs
J005t67: I wanted to keep the old names working to reduce the number of changes needed. Now I realize that it does not add much, (except confusion), I removed them.

SourceCode/AgOpenGPS.Core/Models/ApplicationModel.cs
J005t67: I wanted to keep the old names working to reduce the number of changes needed. Now I realize that it does not add much, (except confusion), I removed them.

FantasyTeddy (approved):

PR #681, opened 2025-03-03
Move conversion part of CNMEA to AppModel.LocalPlane

J005t67:


SourceCode/AgOpenGPS.Core/Models/Base/LocalPlane.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System;
+
+namespace AgOpenGPS.Core.Models
+{
+    public class LocalPlane
```
FantasyTeddy, line 5: I don't really understand what "LocalPlane" stands for... can you explain it in a few words?

SourceCode/AgOpenGPS.Core/Models/Base/LocalPlane.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        const double degreesToRadians = 2.0 * Math.PI / 360.0;
```
FantasyTeddy, line 7: Here we have the conversion again... maybe it is time to introduce some static math class?

SourceCode/AgOpenGPS.Core/Models/Base/LocalPlane.cs
```diff
@@ -0,0 +1,50 @@
...
+        //used to offset the antenna position to compensate for drift
+        static public GeoDelta FixDelta = new GeoDelta(0, 0);
```
FantasyTeddy, line 9: Why is this `static`?

SourceCode/AgOpenGPS.Core/Models/Base/LocalPlane.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        private double _mPerDegreeLat;
+        private double _mPerDegreeLon;
```
FantasyTeddy, line 12: It took me a while to figure out that the "m" stands for "meters". Maybe we should write it out?

SourceCode/AgOpenGPS.Core/Models/Base/LocalPlane.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            _mPerDegreeLat = 111132.92
+                - 559.82 * Math.Cos(2.0 * Origin.Latitude * degreesToRadians)
+                + 1.175 * Math.Cos(4.0 * Origin.Latitude * degreesToRadians)
+                - 0.0023 * Math.Cos(6.0 * Origin.Latitude * degreesToRadians);
+
+            _mPerDegreeLon =
+                111412.84 * Math.Cos(Origin.Latitude * degreesToRadians)
+                - 93.5 * Math.Cos(3.0 * Origin.Latitude * degreesToRadians)
+                + 0.118 * Math.Cos(5.0 * Origin.Latitude * degreesToRadians);
```
FantasyTeddy, line 46: That is quite a lot of magic numbers here...

SourceCode/AgOpenGPS.Core/Models/Base/LocalPlane.cs
J005t67: Yes, added a few comment lines to the code.

SourceCode/AgOpenGPS.Core/Models/Base/LocalPlane.cs
J005t67: Done

SourceCode/AgOpenGPS.Core/Models/Base/LocalPlane.cs
J005t67: It replaces  'vec2 fixOffset' from CNMEA. Just like this fixOffset, it is not specific for  one field, but applicable and (depending on the settings) also shared for all fields.

SourceCode/AgOpenGPS.Core/Models/Base/LocalPlane.cs
J005t67: Yes, better do it now, instead of delaying it forever. I choose to put it in a new file Units.  I plan to move other conversions between metric and imperial there too.

SourceCode/AgOpenGPS.Core/Models/Base/LocalPlane.cs
J005t67: Yes, a lot of magic numbers. I have added a reference to a wiki page that shows the formula.

SourceCode/AgOpenGPS.Core/Models/Base/LocalPlane.cs
FantasyTeddy: Okay, but does it belong in this class then?
Using mutable static member variables can be a source for issues, especially once testing is considered. I would strongly advise against doing it like that.

SourceCode/AgOpenGPS.Core/Models/Base/LocalPlane.cs
FantasyTeddy: Perfect!

SourceCode/AgOpenGPS.Core/Models/Base/LocalPlane.cs
J005t67: I was preparing a lengthy reply to explain why this was correct in my opinion, when I discovered that I forgot to apply the drift compensation in ConvertGeoCoordToWgs84. This ofcourse made it hard to review and very confusing.
I also improved the variable name to make clear that it applies to more than one field.
I agree with you that this static variable is error prone and complex, but now you can see that the code does not add any complexity without a reason. It merely shows the  complexity of the problem it takes care of.

FantasyTeddy: I am convinced, that this will make problems later (good luck writing tests for this).
How did you try to remove the static? I think it should be pretty easy if you add a new class like `MultiFieldProperies` with this value inside, that you then pass to the constructor of `LocalPlane`.

J005t67: I agree with your comment and changed the code accordingly. The dependency is now very explicit and that is a good thing. In the future, when field selection runs on the refactored code, I will make Field a subclass of LocalPlane and move the SharedFieldProperties to Fields.

FantasyTeddy: Awesome, thank you!

FantasyTeddy (approved):

PR #684, opened 2025-03-06
Ensure AgIO restart is done properly

FantasyTeddy:
When AgIO is restarting itself, there are a few things that can go wrong (as discovered by @lansalot):
- The new application instance is being started before the current application can release its `Mutex`.
  - In that case AgIO will just appear to not restart (even though it tried, but thought it is still running)
- The `Mutex` is not being properly disposed, because of the `Environment.Exit` call.
  - The new AgIO instance will then fail with an `AbandonedMutexException` (can be observed in the Event Viewer)

### DIFF
--- a/SourceCode/AgIO/Source/Program.cs
+++ b/SourceCode/AgIO/Source/Program.cs
@@ -1,10 +1,6 @@
-﻿using AgIO.Properties;
-using AgLibrary.Logging;
-using Microsoft.Win32;
+﻿using AgLibrary.Logging;
 using System;
-using System.Configuration;
 using System.Globalization;
-using System.IO;
 using System.Reflection;
 using System.Threading;
 using System.Windows.Forms;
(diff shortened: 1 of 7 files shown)


richardklasens (commented):
Is it an option to call the restart function on the places where it needs a restart instead of doing the same code everytime?

FantasyTeddy:
> Is it an option to call the restart function on the places where it needs a restart instead of doing the same code everytime?

Isn't that what I'm doing by providing the `Program.Restart()` method? Or do you mean something else?

SourceCode/AgIO/Source/Program.cs
```diff
@@ -23,31 +19,39 @@ internal static class Program
...
+                }
+            }
+        }
+
+        // If the application needs to be restarted, it should be done via this method.
+        // Manually disposing the Mutex before calling Application.Restart() ensures
+        // that the Mutex is not still owned by the current application instance while
+        // the new application instance is being started.
+        // See: https://stackoverflow.com/a/9456822
+        public static void Restart()
```
richardklasens, line 46: I mean call this on the places where it needs a restart.

FantasyTeddy: Am I not doing that? Did I miss some other places where a restart is needed?

richardklasens (approved):
You’re right.. misread it

PR #685, opened 2025-03-06
Refactor/field selection

J005t67:
Small changes to prepare field selection.


SourceCode/AgOpenGPS.Core/Models/Field/FieldDescription.cs
```diff
@@ -1,20 +1,42 @@
...
     {
-        public FieldDescription(string name)
+        private readonly DirectoryInfo _fieldDirectory;
+        public FieldDescription(DirectoryInfo fieldDirectory)
         {
-            Name = name;
+            _fieldDirectory = fieldDirectory;
+            try
+            {
+                var overview = new OverviewStreamer().Read(fieldDirectory);
```
FantasyTeddy, line 19: There are a few drawbacks with doing extensive initialization work (like reading a file) inside a constructor.
One of these reasons that comes to mind is: what if we want the file read to be asynchronous? We could not do that within the constructor.
Instead I would propose to create a `static` factory method inside this class that does the file reading and returns an instance of `FieldDescription`.

J005t67: Done

FantasyTeddy (approved):

PR #686, opened 2025-03-07
Use Avalonia instead of WPF for sample application

FantasyTeddy:
Since replacing one Windows-only framework with another one (WinForms -> WPF) is not really what we are aiming for, the previously introduced WPF sample application (`AgOpenGPS.WpfApp`) is now replaced with an application based on [Avalonia UI](https://avaloniaui.net/) (`AgOpenGPS.Avalonia.App`).


J005t67:
It's great to include the Avalonia app, but to me it is not a sample app. It is THE application that one day will replace the current winforms-app. So, from this starting point we must work towards a version that does a proper job regarding icons, translation, layout, styles etc. On the other hand, when I'm working on the refactoring towards MVVM I need a simple test application that uses databinding to test my new viewmodels. And that is exactly what the current WPF app is for. The WPF app will be very efficient for me because I will do not need to spent time on icons, translations, layout styling etc. So, do not remove the WPF-app. It will slow down my refactoring.

SourceCode/AgOpenGPS.Avalonia.App/App.axaml
```diff
@@ -0,0 +1,15 @@
+<Application xmlns="https://github.com/avaloniaui"
+             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+             x:Class="AgOpenGPS.Avalonia.App.App"
```
J005t67, line 3: Put class name on first line

SourceCode/AgOpenGPS.Avalonia.App/App.axaml
```diff
@@ -0,0 +1,15 @@
...
+             xmlns:local="using:AgOpenGPS.Avalonia.App"
```
J005t67, line 4: Use normal indendation of 4 spaces

SourceCode/AgOpenGPS.Avalonia.App/MainViews/BottomButtonStrip.axaml
```diff
@@ -0,0 +1,23 @@
...
+             xmlns="https://github.com/avaloniaui"
+             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+             xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+             mc:Ignorable="d"
+
+             d:DesignWidth="800">
+
+    <Grid x:Name="bottomButtonStripGrid"
+          HorizontalAlignment="Stretch">
```
J005t67, line 11: Use indent of 4

SourceCode/AgOpenGPS.Avalonia.App/MainWindow.axaml
```diff
@@ -0,0 +1,26 @@
+<Window x:Class="AgOpenGPS.Avalonia.App.MainWindow"
+        xmlns="https://github.com/avaloniaui"
+        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"        mc:Ignorable="d"
```
J005t67, line 5: This line seems to contain a strange character after 2006"

SourceCode/AgOpenGPS.Avalonia.App/ViewLocator.cs
```diff
@@ -0,0 +1,34 @@
+using System;
+using System.ComponentModel;
+using Avalonia.Controls;
+using Avalonia.Controls.Templates;
+
+namespace AgOpenGPS.Avalonia.App
+{
+    public class ViewLocator : IDataTemplate
```
J005t67, line 8: I'm not sure what this class tries to do, but it seems that I tries to automatically find a corresponding view for a viewModel. I don't think we want or need something like that in our application.

SourceCode/AgOpenGPS.WpfApp/App.xaml
```diff
@@ -1,10 +0,0 @@
-﻿<Application x:Class="AgOpenGPS.WpfApp.App"
```
J005t67, line 1: I need the WPF app for efficiency. It should not be removed.

J005t67 (changes requested):
Great to have the Avalonia app!

Please also correct the indentations I did not mention explicitly.

FantasyTeddy:
Okay, so there is not really a need for the Avalonia app then. I was under the impression that we would benefit from not making a detour via WPF and going straight for Avalonia. It would have also been a good learning opportunity for all of us, to get familiar with the framework, before actually trying to bootstrap the entire "real" application.

But I close this PR and we add the "real" app when you are ready.

PR #689, opened 2025-03-07
Bugfix: index out of bounds when clicking the green or yellow symbol for marking

steran82:
Bugfix: when opening a field and clicking one of the 2 mark-buttons (green or yellow) before you have selected a track an error occured in Sections.Designer.cs Method markAsWorkedTrack() - its about the new feature which allows skipping already worked tracks at youturn.

### DIFF
--- a/SourceCode/GPS/Forms/Sections.Designer.cs
+++ b/SourceCode/GPS/Forms/Sections.Designer.cs
@@ -17,24 +17,31 @@ namespace AgOpenGPS
         //Off, Manual, and Auto, 3 states possible
         public btnStates manualBtnState = btnStates.Off;
         public btnStates autoBtnState = btnStates.Off;
-
+        
         private void markAsWorkedTrack()
         {
-            //is it an AB Line or a Curve?
-            if (trk.gArr[this.trk.idx].mode == TrackMode.AB)
+            
+            //only mark if there was a track selected?
+            if (this.trk.idx >= 0)
             {
-                //AB Line
-                if (!this.trk.gArr[this.trk.idx].workedTracks.Contains(this.ABLine.howManyPathsAway))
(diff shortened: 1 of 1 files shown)


SourceCode/GPS/Forms/Sections.Designer.cs
```diff
@@ -17,24 +17,31 @@ public partial class FormGPS
         //Off, Manual, and Auto, 3 states possible
         public btnStates manualBtnState = btnStates.Off;
         public btnStates autoBtnState = btnStates.Off;
-
+        
         private void markAsWorkedTrack()
```
J005t67, line 21: Start method names with a capital

SourceCode/GPS/Forms/Sections.Designer.cs
```diff
@@ -17,24 +17,31 @@ public partial class FormGPS
...
+                {
+                    //AB Line
+                    if (!this.trk.gArr[this.trk.idx].workedTracks.Contains(this.ABLine.howManyPathsAway))
+                        this.trk.gArr[this.trk.idx].workedTracks.Add(this.ABLine.howManyPathsAway);
 
+                }
+                else if (trk.gArr[this.trk.idx].mode == TrackMode.Curve)
+                {
+                    //Curve
+                    if (!this.trk.gArr[this.trk.idx].workedTracks.Contains(this.curve.howManyPathsAway))
```
J005t67, line 39: workedTracks is a List<double>, so the method workedTracks.Contains() will compare doubles, to check if the value is already present.
Comparing doubles is always tricky. If two double are only very slightly different (for example only the least significant bit is different), they are considered unequal. In short: the method Contain() will probably return false many times in cases where you expect true;

J005t67 (changes requested):
I really think that the large amounts of duplicate code in AgOpenGPS are slowing us down.
Sorry, for being a bit harsh on that topic..

PR #690, opened 2025-03-07
Introduce color lib

J005t67:
I started a color library for the following reasons.
- color names are more easy to read and remember than hardcoded colors
- color names are more easy to search for, especially when they contain both the functional name, (so yo can search for TramLine/TramLineDot etc) and a physical name (so you can search for Black/Blue etc.)
- currently the code contains about 1600 direct dependencies to opengl toolkit. (Search for 'GL.') For portability it is better to start moving all the colors to the colorlib.

### DIFF
--- a/SourceCode/AgOpenGPS.Core/Models/Base/ColorRgb.cs
+++ b/SourceCode/AgOpenGPS.Core/Models/Base/ColorRgb.cs
@@ -1,4 +1,6 @@
-﻿namespace AgOpenGPS.Core.Models
+﻿using OpenTK.Graphics.OpenGL;
+
+namespace AgOpenGPS.Core.Models
 {
     public struct ColorRgb
     {
@@ -9,9 +11,22 @@
             Blue = blue;
         }
 
-        public byte Red { get; }
-        public byte Green { get; }
-        public byte Blue { get; }
+        public ColorRgb(float red, float green, float blue) :
+        this(
+            ColorRgb.FloatToByte(red),
+            ColorRgb.FloatToByte(green),
+            ColorRgb.FloatToByte(blue))
+        {
+        }
+
+        public byte Red { get; set; }
+        public byte Green { get; set; }
+        public byte Blue { get; set; }
+
+        public void SetColor()
+        {
+            GL.Color3(Red, Green, Blue);
+        }
 
         public static explicit operator System.Drawing.Color(ColorRgb color)
         {
@@ -22,5 +37,11 @@
         {
             return new ColorRgb(color.R, color.G, color.B);
         }
+
+        static private byte FloatToByte(float fraction)
+        {
+            return (byte)(255 * fraction);
+        }
     }
+
 }
--- a/SourceCode/AgOpenGPS.Core/Models/Base/ColorRgb.cs
+++ b/SourceCode/AgOpenGPS.Core/Models/Base/ColorRgb.cs
@@ -18,6 +18,9 @@ namespace AgOpenGPS.Core.Models
                 ColorRgb.FloatToByte(green),
                 ColorRgb.FloatToByte(blue))
         {
+            if (red < 0.0f || 255.0 < red) throw new ArgumentOutOfRangeException("red", "Argument out of range");
+            if (green < 0.0f || 255.0 < green) throw new ArgumentOutOfRangeException("green", "Argument out of range");
+            if (blue < 0.0f || 255.0 < blue) throw new ArgumentOutOfRangeException("blue", "Argument out of range");
         }
 
         public byte Red { get; set; }
@@ -36,8 +39,6 @@ namespace AgOpenGPS.Core.Models
 
         static private byte FloatToByte(float fraction)
         {
-            fraction = Math.Max(fraction, 0.0f);
-            fraction = Math.Min(fraction, 1.0f);
             return (byte)(255 * fraction);
         }
     }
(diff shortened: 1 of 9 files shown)


SourceCode/AgOpenGPS.Core/Models/Base/ColorRgb.cs
```diff
@@ -9,9 +11,22 @@ public ColorRgb(byte red, byte green, byte blue)
             Blue = blue;
         }
 
-        public byte Red { get; }
-        public byte Green { get; }
-        public byte Blue { get; }
+        public ColorRgb(float red, float green, float blue) :
```
FantasyTeddy, line 14: Since this constructor only works correctly with values from 0.0 - 1.0, should we maybe check that this is true?

SourceCode/AgOpenGPS.Core/Models/Base/ColorRgb.cs
```diff
@@ -9,9 +11,22 @@ public ColorRgb(byte red, byte green, byte blue)
...
+        {
+        }
+
+        public byte Red { get; set; }
+        public byte Green { get; set; }
+        public byte Blue { get; set; }
+
+        public void SetColor()
+        {
+            GL.Color3(Red, Green, Blue);
```
FantasyTeddy, line 28: Is it really a good idea to mix the responsibility of storing a color with applying it in OpenGL?

SourceCode/GPS/Classes/CTool.cs
```diff
@@ -148,25 +174,8 @@ public void DrawTool()
...
-                GL.LineWidth(1);
-                //draw the rigid hitch
-                GL.Color3(0.765f, 0.76f, 0.32f);
-                GL.Begin(PrimitiveType.LineLoop);
-                GL.Vertex3(-0.57, trailingTank, 0);
-                GL.Vertex3(0, 0, 0);
-                GL.Vertex3(0.57, trailingTank, 0);
-                GL.End();
+                DrawHitchLayer(true, trailingTank);
+                DrawHitchLayer(false, trailingTank);
```
FantasyTeddy, line 178: Wouldn't it make more sense, if `DrawHitchLayer` was only called once, and did the drawing of the background/foreground internally? (maybe in a small loop)

SourceCode/GPS/Forms/OpenGL.Designer.cs
```diff
@@ -1922,33 +1916,27 @@ private void DrawFlags()
         {
             try
             {
-                int flagCnt = flagPts.Count;
-                for (int f = 0; f < flagCnt; f++)
+                foreach (CFlag flag in flagPts)
                 {
                     GL.PointSize(8.0f);
                     GL.Begin(PrimitiveType.Points);
+                    ColorRgb flagColorRgb = Colors.FlagRedColor;
```
FantasyTeddy, line 1923: I think this color is only ever assigned. Shouldn't you also apply it?

SourceCode/GPS/Classes/CTool.cs
J005t67: Yes, looks better now.

SourceCode/AgOpenGPS.Core/Models/Base/ColorRgb.cs
J005t67: No, obviously not. ;-) In the context of configuration for example, we don't want the dependency to the OpenGL toolkit.

SourceCode/GPS/Forms/OpenGL.Designer.cs
J005t67: Oops, a bug indeed.

SourceCode/AgOpenGPS.Core/Models/Base/ColorRgb.cs
FantasyTeddy: You solved this by clamping the values to 0.0 - 1.0, so it is essentially allowed to pass values lower than 0.0 and bigger than 1.0 to this constructor. Is that really what we want?

SourceCode/GPS/Classes/CTool.cs
FantasyTeddy: Very neat solution with the `LineStyle` 👍

SourceCode/AgOpenGPS.Core/Models/Base/ColorRgb.cs
J005t67: Made Color constructors more strict.

SourceCode/AgOpenGPS.Core/Models/Base/ColorRgb.cs
```diff
@@ -18,6 +18,9 @@ public ColorRgb(float red, float green, float blue) :
                 ColorRgb.FloatToByte(green),
                 ColorRgb.FloatToByte(blue))
         {
+            if (red < 0.0f || 255.0 < red) throw new ArgumentOutOfRangeException("red", "Argument out of range");
```
FantasyTeddy, line 21:
```suggestion
            if (red < 0.0f || 255.0 < red) throw new ArgumentOutOfRangeException(nameof(red), "Argument out of range");
```

J005t67: Changed.

FantasyTeddy (approved):

PR #693, opened 2025-03-08
Refactor AgDiag

FantasyTeddy:
- Functionality is extracted from the main `FormLoop`
- The [`UdpClient`](https://learn.microsoft.com/en-us/dotnet/api/system.net.sockets.udpclient) class is used instead of raw `Socket` 
- Instead of using the Asynchronous Programming Model (APM), the more modern Task-based Asynchronous Pattern (TAP) is being used

### DIFF
--- a/SourceCode/AgDiag/Forms/FormLoop.cs
+++ b/SourceCode/AgDiag/Forms/FormLoop.cs
@@ -44,7 +44,7 @@ namespace AgDiag
             lblSetSteerAngle.Text = (_pgns.asData.Bytes[_pgns.asData.steerAngleHi] << 8 | _pgns.asData.Bytes[_pgns.asData.steerAngleLo]).ToString();
             lblStatus.Text = _pgns.asData.Bytes[_pgns.asData.status].ToString();
 
-            lblSteerDataPGN.Text = BitConverter.ToString(_pgns.asData.Bytes);
+            lblSteerDataPGN.Text = _pgns.asData.ToHexString();
 
             //from autosteer  module
             lblSteerAngleActual.Text = ((Int16)((_pgns.asModule.Bytes[_pgns.asModule.actualHi] << 8)
@@ -66,10 +66,10 @@ namespace AgDiag
                 lblSteerSwitch.BackColor = Color.Red;
             else lblSteerSwitch.BackColor = Color.Green;
 
-            lblPGNFromAutosteerModule.Text = BitConverter.ToString(_pgns.asModule.Bytes);
+            lblPGNFromAutosteerModule.Text = _pgns.asModule.ToHexString();
 
             //Autosteer settings
-            lblPGNSteerSettings.Text = BitConverter.ToString(_pgns.asSet.Bytes);
+            lblPGNSteerSettings.Text = _pgns.asSet.ToHexString();
             lblP.Text = _pgns.asSet.Bytes[_pgns.asSet.gainProportional].ToString();
             lblHiPWM.Text = _pgns.asSet.Bytes[_pgns.asSet.highPWM].ToString();
             lblLoPWM.Text = _pgns.asSet.Bytes[_pgns.asSet.lowPWM].ToString();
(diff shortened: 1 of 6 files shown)


FantasyTeddy:
@AgOpenGPS-Official/aog-coders: Should I split up this PR into smaller ones to make it easier to review?

SourceCode/AgDiag/Forms/FormLoop.cs
```diff
@@ -29,90 +37,88 @@ private static string ByteArrayToHex(byte[] barray)
 
         private void timer1_Tick(object sender, EventArgs e)
         {
-            if ((asData.pgn[asData.sc1to8] & 1) == 1) lblSection1.BackColor = Color.Green;
```
J005t67, line 32: Or use a small function
private Color SectionBackColor(bool isOn)
{
    return isOn ? Color.Green : ColorRed;
}

And use it like:
lblSection2.BackColor = SectionBackColor(_pgns.asData.pgn[_pgns.asData.sc1to8] & 2 > 0)

SourceCode/AgDiag/Forms/FormLoop.cs
```diff
@@ -29,90 +37,88 @@ private static string ByteArrayToHex(byte[] barray)
...
+            if ((_pgns.asData.pgn[_pgns.asData.sc1to8] & 64) == 64) lblSection7.BackColor = Color.Green;
             else lblSection7.BackColor = Color.Red;
-            if ((asData.pgn[asData.sc1to8] & 128) == 128) lblSection8.BackColor = Color.Green;
+            if ((_pgns.asData.pgn[_pgns.asData.sc1to8] & 128) == 128) lblSection8.BackColor = Color.Green;
             else lblSection8.BackColor = Color.Red;
 
-            lblSpeed.Text = (asData.pgn[asData.speedHi] << 8 | asData.pgn[asData.speedLo]).ToString();
-            lblSetSteerAngle.Text = (asData.pgn[asData.steerAngleHi] << 8 | asData.pgn[asData.steerAngleLo]).ToString();
-            lblStatus.Text = asData.pgn[asData.status].ToString();
+            lblSpeed.Text = (_pgns.asData.pgn[_pgns.asData.speedHi] << 8 | _pgns.asData.pgn[_pgns.asData.speedLo]).ToString();
```
J005t67: I do not know the precedence of the bitwise or '|'. I would prefer to combine the two bytes using a simple + or to add parentheses.

SourceCode/AgDiag/Protocol/PGNs.cs
```diff
@@ -1,15 +1,15 @@
 ﻿namespace AgDiag.Protocol
 {
+    public abstract class PGN
```
J005t67, line 6: My personal preference would be the name PgnBase

SourceCode/AgDiag/Protocol/PGNs.cs
```diff
@@ -1,15 +1,15 @@
...
+    {
+        public byte[] Bytes { get; set; }
+    }
+
```
J005t67, line 15: Why not move a bit more code to the base class? For example the initialization of the Bytes

J005t67 (dismissed):
I reviewed the commits one by one and while doing so, I discovered that some of my comments were already addressed in later commits. Please ignore those.

FantasyTeddy:
Thank you for the thorough review. I tried to fix all of your comments.

richardklasens (approved):

SourceCode/AgDiag/Protocol/PGNs.cs
```diff
@@ -18,14 +18,20 @@ public class CPGN_FE
...
-        {
             /// <summary>
-            /// From steer module
+            /// autoSteerData FE 254 speedHi=5 speedLo=6  status = 7 free = 8;
+            /// steerAngleHi = 9 steerAngleLo = 10 tramControl = 11 sc1to8 = 12 sc9to16 = 13;
             /// </summary>
-            public byte[] pgn = new byte[] { 0x80, 0x81, 0x7f, 0xFD, 8, 0, 0, 0, 0, 0, 0, 0, 0, 0xCC };
+            public CPGN_FE()
+            {
+                Bytes = new byte[] { 0x80, 0x81, 0x7f, 0xFE, 8, 0, 0, 0, 0, 0, 0, 0, 0, 0xCC };
```
J005t67: It is up to you, but I would move this initialization to the base class too. The base class only needs to know the ID and the message length to do so.

J005t67 (approved):
I added one minor comment.

SourceCode/AgDiag/Protocol/PGNs.cs
FantasyTeddy: There is a lot more that could be done to improve this application. However, I already did way more than I originally wanted to and I think we will visit this problem again, when we continue with #585 for example.

PR #694, opened 2025-03-09
Bugfix: index out of bounds when clicking the green or yellow symbol …

steran82:
Bugfix: when opening a field and clicking one of the 2 mark-buttons (green or yellow) before you have selected a track an error occured in Sections.Designer.cs Method markAsWorkedTrack() - its about the new feature which allows skipping already worked tracks at youturn.

replacement for PR: #689

### DIFF
--- a/SourceCode/GPS/Forms/Sections.Designer.cs
+++ b/SourceCode/GPS/Forms/Sections.Designer.cs
@@ -18,24 +18,33 @@ namespace AgOpenGPS
         public btnStates manualBtnState = btnStates.Off;
         public btnStates autoBtnState = btnStates.Off;
 
-        private void markAsWorkedTrack()
+        private void MarkAsWorkedTrack()
         {
-            //is it an AB Line or a Curve?
-            if (trk.gArr[this.trk.idx].mode == TrackMode.AB)
+            // only mark if there was a track selected
+            if (this.trk.idx >= 0)
             {
-                //AB Line
-                if (!this.trk.gArr[this.trk.idx].workedTracks.Contains(this.ABLine.howManyPathsAway))
-                    this.trk.gArr[this.trk.idx].workedTracks.Add(this.ABLine.howManyPathsAway);
+                var track = this.trk.gArr[this.trk.idx];
 
-            }
-            else if (trk.gArr[this.trk.idx].mode == TrackMode.Curve)
-            {
-                //Curve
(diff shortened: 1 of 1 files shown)


steran82:
I noticed the comment that the check for List (workedTracks.Contains isn' ideal for doubles; but i dont have an idea how to improve this.
Is this really a problem, when it comes from the same source (howManyPathsAway) and it is compared with this?
I thought about changing it to an integer - but then we will eventually run into an overflow when track/field is still active and losing gps or getting illegal gps coordinates...

i tried to refactore too, to avoid duplicate code...

SourceCode/GPS/Forms/Sections.Designer.cs
```diff
@@ -18,26 +18,35 @@ public partial class FormGPS
         public btnStates manualBtnState = btnStates.Off;
         public btnStates autoBtnState = btnStates.Off;
 
-        private void markAsWorkedTrack()
+        private void MarkAsWorkedTrack()
         {
-            //is it an AB Line or a Curve?
-            if (trk.gArr[this.trk.idx].mode == TrackMode.AB)
+            // only mark if there was a track selected
+            if (this.trk.idx >= 0)
```
FantasyTeddy, line 24: Maybe use an early return here:
```suggestion
            if (this.trk.idx < 0) return;
```

FantasyTeddy: And also remove the `{` afterwards. (That's the whole idea of the early return: to get rid of an indentation level 😉)

FantasyTeddy:
Regarding the `double` issue:
I added two commits to your branch:
1. Use `int` instead of `double` for `workedTracks`. I don't quite understand how this should lead to some kind of overflow. Could you explain that?
2. Use a `HashSet` to store the `workedTracks`, because the previously used `List` is actually being treated like a hash set.

@J005t67: What do you think?

steran82:
> Regarding the `double` issue: I added two commits to your branch:
> 
>     1. Use `int` instead of `double` for `workedTracks`. I don't quite understand how this should lead to some kind of overflow. Could you explain that?
I thought about it again and also played with the debugger.
I don't think it can occur. If I manipulate the values in the debugger as much as it could not be in reality, I get a different error message:

Serious Field Origin Error

Field Origin is More Then 20 km from your current GPS
Position Delete this field and create a new one as
Accuracy will be poor

Or you may have a field open and drove far away

So in my humble opinion integer should be enough.

Translated with DeepL.com (free version)

PR #695, opened 2025-03-10
Fixes/6.7

richardklasens:


richardklasens:
Since there was a lot of discussion about the New Vehicle option, i removed it for now, and the code is still in there, but commented, i would leave it there for now if there is any one missing the feature, its easy to put back in. But can probably be removed completely

SourceCode/GPS/Forms/Settings/ConfigVehicle.Designer.cs
```diff
@@ -184,74 +184,6 @@ private void tboxCreateNewVehicle_TextChanged(object sender, EventArgs e)
...
-            //{
-            //    btnVehicleNewSave.Enabled = false;
-            //    btnVehicleNewSave.BackColor = Color.Transparent;
-            //}
-            //else
-            //{
-            //    btnVehicleNewSave.Enabled = true;
-            //    btnVehicleNewSave.BackColor = Color.LimeGreen;
-            //}
-        }
```
FantasyTeddy, line 214: This should stay :)

FantasyTeddy (approved):

PR #698, opened 2025-03-11
First viewmodels for field selection

J005t67:
This PR contains the basic functionality of the first four WPF dialogs and corresponding viewmodels.
Beacuse these are the first viewmodels, this PR also includes some infrastructure, like ApplicationViewModel and PanelPresenter
Not all functionality has been implemented yet. About ten issues still needs attention and are documented in the code using' '// TODO'
Note that the correspondig winform views are not part of this PR. The original Winforms-app can only start using the new viewmodels, when all functionality has been implemented.

### DIFF
--- a/SourceCode/AgOpenGPS.Core/ViewModels/ApplicationViewModel.cs
+++ b/SourceCode/AgOpenGPS.Core/ViewModels/ApplicationViewModel.cs
@@ -1,0 +1,43 @@
+﻿using AgLibrary.ViewModels;
+using AgOpenGPS.Core.Presenters;
+using AgOpenGPS.Core.ViewModels;
+using System.Windows.Input;
+
+namespace AgOpenGPS.Core.ViewModels
+{
+    // TODO implement Day and Night mode
+    // TODO implement Metric and Imperial mode
+    public class ApplicationViewModel : ViewModel
+    {
+        private readonly ApplicationModel _appModel;
+        private ApplicationPresenter _applicationPresenter;
+        private StartNewFieldViewModel _startNewFieldViewModel;
+
+        public ApplicationViewModel(ApplicationModel appModel)
+        {
+            _appModel = appModel;
+            StartNewFieldCommand = new RelayCommand(StartNewField);
+        }
+
+        public void SetPresenter(ApplicationPresenter appPresenter)
+        {
+            _applicationPresenter = appPresenter;
+            StartNewFieldViewModel.PanelPresenter = appPresenter.PanelPresenter;
+        }
+
+        public StartNewFieldViewModel StartNewFieldViewModel
+        {
+            get
+            {
+                if (_startNewFieldViewModel == null) _startNewFieldViewModel = new StartNewFieldViewModel(_appModel);
+                return _startNewFieldViewModel;
+            }
+        }
+
+        public ICommand StartNewFieldCommand { get; }
+        private void StartNewField()
+        {
+            _applicationPresenter.PresentStartNewField(StartNewFieldViewModel);
+        }
+    }
+}
--- a/SourceCode/AgOpenGPS.Core/ViewModels/StartNewFieldViewModels/CreateFromExistingFieldViewModel.cs
+++ b/SourceCode/AgOpenGPS.Core/ViewModels/StartNewFieldViewModels/CreateFromExistingFieldViewModel.cs
@@ -1,0 +1,92 @@
+﻿using AgLibrary.ViewModels;
+using AgOpenGPS.Core.Interfaces;
+using AgOpenGPS.Core.Models;
+using System.Globalization;
+using System;
+using System.Windows.Input;
+
+namespace AgOpenGPS.Core.ViewModels
+{
+    public class CreateFromExistingFieldViewModel : FieldTableViewModel
+    {
+        private readonly IPanelPresenter _panelPresenter;
+        private string _newFieldName = "";
+
+        public CreateFromExistingFieldViewModel(
+            ApplicationModel appModel,
+            IPanelPresenter panelPresenter
+        )
+            : base(appModel)
+        {
+            _panelPresenter = panelPresenter;
+            AddVehicleCommand = new RelayCommand(AddVehicle);
+            AddDateCommand = new RelayCommand(AddDate);
+            AddTimeCommand = new RelayCommand(AddTime);
+            BackSpaceCommand = new RelayCommand(BackSpace);
+        }
+
+        public new FieldDescriptionViewModel LocalSelectedField
+        {
+            get { return _localSelectedField; }
+            set
+            {
+                if (value != _localSelectedField)
+                {
+                    _localSelectedField = value;
+                    NotifyPropertyChanged();
+                    NewFieldName = _localSelectedField.FieldName;
+                }
+            }
+        }
+
+        public string NewFieldName
+        {
+            get { return _newFieldName; }
+            set {
+                if (value != _newFieldName )
+                {
+                    _newFieldName = value;
+                    NotifyPropertyChanged();
+                }
+            }
+        }
+
+        public ICommand AddVehicleCommand { get; }
+        public ICommand AddDateCommand { get; }
+        public ICommand AddTimeCommand { get; }
+        public ICommand BackSpaceCommand { get; }
+
+
+        public bool MustCopyFlags { get; set; }
+        public bool MustCopyMapping { get; set; }
+        public bool MustCopyHeadland { get; set; }
+        public bool MustCopyLines { get; set; }
+
+        private void AddVehicle()
+        {
+            NewFieldName += " " + "Vehicle"; // TODO RegistrySettings.vehicleFileName;
+        }
+
+        private void AddDate()
+        {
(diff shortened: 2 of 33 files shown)


SourceCode/AgOpenGPS.Core/ViewModels/ApplicationViewModel.cs
```diff
@@ -0,0 +1,43 @@
...
+
+namespace AgOpenGPS.Core.ViewModels
+{
+    // TODO implement Day and Night mode
+    // TODO implement Metric and Imperial mode
+    public class ApplicationViewModel : ViewModel
+    {
+        private readonly ApplicationModel _appModel;
+        private ApplicationPresenter _applicationPresenter;
+        private StartNewFieldViewModel _startNewFieldViewModel;
```
FantasyTeddy, line 14: Why is the `StartNewFieldViewModel` owned by the `ApplicationViewModel` and not for example by the `WpfPanelPresenter` (which also owns the corresponding `StartNewFieldDialog`)?

SourceCode/AgOpenGPS.Core/ViewModels/StartNewFieldViewModels/FieldTableViewModel.cs
```diff
@@ -0,0 +1,81 @@
...
+            var descriptions = _appModel.Fields.GetFieldDescriptions();
+            foreach (FieldDescription fd in descriptions)
+            {
+                FieldDescriptionViewModel viewModel = new FieldDescriptionViewModel(
+                    fd.FieldDirectory,
+                    fd.Area.HasValue ? fd.Area.Value.ToString() : "Error",
+                    fd.Wgs84Start.HasValue ? _appModel.CurrentLatLon.DistanceInKiloMeters(fd.Wgs84Start.Value).ToString() : "Error");
+                viewModels.Add(viewModel);
+            }
+            FieldDescriptionViewModels = viewModels;
```
FantasyTeddy, line 62: Isn't that kind of defeating the purpose of an `ObservableCollection`, if you only ever set it as a whole and not add to it dynamically?

SourceCode/AgOpenGPS.Core/ViewModels/StartNewFieldViewModels/FieldTableViewModel.cs
```diff
@@ -0,0 +1,81 @@
...
+                    NotifyPropertyChanged();
+                }
+            }
+        }
+
+        public void UpdateFields()
+        {
+            ObservableCollection<FieldDescriptionViewModel> viewModels = new ObservableCollection<FieldDescriptionViewModel>();
+            var descriptions = _appModel.Fields.GetFieldDescriptions();
+            foreach (FieldDescription fd in descriptions)
```
FantasyTeddy, line 54:
```suggestion
            foreach (FieldDescription description in descriptions)
```

SourceCode/AgOpenGPS.Core/ViewModels/StartNewFieldViewModels/StartNewFieldViewModel.cs
```diff
@@ -0,0 +1,111 @@
...
+            StartSelectFieldCommand = new RelayCommand(StartSelectField);
+            CancelCommand = new RelayCommand(Cancel);
+        }
+
+        public IPanelPresenter PanelPresenter
+        {
+            set {
+                _panelPresenter = value;
+            }
+        }
```
FantasyTeddy, line 34:
```suggestion
        public IPanelPresenter PanelPresenter { private get; set; }
```

SourceCode/AgOpenGPS.Core/ViewModels/StartNewFieldViewModels/CreateFromExistingFieldViewModel.cs
```diff
@@ -0,0 +1,92 @@
+﻿using AgLibrary.ViewModels;
+using AgOpenGPS.Core.Interfaces;
+using AgOpenGPS.Core.Models;
+using System.Globalization;
+using System;
+using System.Windows.Input;
+
+namespace AgOpenGPS.Core.ViewModels
+{
+    public class CreateFromExistingFieldViewModel : FieldTableViewModel
```
FantasyTeddy, line 10: Maybe it would be better to keep the `FieldTableViewModel` as a property, instead of inheriting from it? (Composition over inheritance)
This would also match very well, because `FieldTable` is its own view anyway.

SourceCode/AgOpenGPS.WpfApp/MainViews/LeftButtonStrip.xaml
```diff
@@ -4,11 +4,20 @@
...
+    d:DataContext="{d:DesignInstance Type=corevm:ApplicationViewModel, IsDesignTimeCreatable=False}"
     d:DesignHeight="700">
 
     <Grid x:Name="leftButtonStripGrid" VerticalAlignment="Stretch">
-        <StackPanel VerticalAlignment="Bottom">
+        <StackPanel VerticalAlignment="Top">
+            <Button Height="80" Width="100" />
+            <Button Height="80" Width="100" />
             <Button Height="80" Width="100" />
+            <Button x:Name="startNewFieldButton"
```
FantasyTeddy, line 17: I see that you tend to give every UI element a `x:Name`. Is there a particular reason for that?

J005t67: The messages about DataBinding errors are much better this way. So I consider it good practice to name all UI-element that use DataBinding.

SourceCode/AgOpenGPS.Core/ViewModels/StartNewFieldViewModels/CreateFromExistingFieldViewModel.cs
J005t67: I started that way, but in that case the paths in the Bindings look a bit more difficult. Some of the bindings will need a path that consists of two nodes. Something like "{Binding FieldTable.LocalSelectedField}" This works perfectly with WPF, (without any additional effort) because WPF will observe both changes to FieldTable and LocalSlecteField to come up with the correct result, but on Winforms it would be a lot more difficult. Also the Sort button is 'a bit in between' it belongs to the FieldTableViewModel, but it is not  in the FieldTableView. This requires a bit less code  when using inheritance.

SourceCode/AgOpenGPS.Core/ViewModels/StartNewFieldViewModels/FieldTableViewModel.cs
J005t67: Will add this comment to the code:
            // The Winforms views do not update when elements inside the ObservableCollection are changed.
            // Therefore change the ObservableCollection as a whole.

FantasyTeddy: What if we used the `ObservableCollection` as intended and just add a `NotifyPropertyChanged(nameof(FieldDescriptionViewModels))` at the end? Would that work?

SourceCode/AgOpenGPS.WpfApp/MainViews/LeftButtonStrip.xaml
FantasyTeddy: That's good to know, thank you for the explanation.

SourceCode/AgOpenGPS.Core/ViewModels/ApplicationViewModel.cs
J005t67: I do not want to put too many commands and ViewModels directly in the ApplicationViewModel, but I did not figure out what the best way is to distribute them. I wanted to reconsider the best place for StartNewFieldViewModel later.
The WpfPanelPresenter is not an option because the Winforms app can not reach it. If you agree, I could already introduce LeftButtonStripViewModel and move it there.

FantasyTeddy: Let's keep it as is for the moment, I guess.

SourceCode/AgOpenGPS.Core/ViewModels/StartNewFieldViewModels/StartNewFieldViewModel.cs
J005t67: Changed

SourceCode/AgOpenGPS.Core/ViewModels/StartNewFieldViewModels/FieldTableViewModel.cs
J005t67: Yes, that would work, and it also would be an performance improvement if only a small number of items in the collection change. In our case, all items in the collection change and if we do that on the active collection, the view would 'suffer' from lots of CollectionChanged events. That's why I like to do the changes on the 'inactve' collection and set the collection as a whole.

FantasyTeddy: Okay, fair point. But why even use `ObservableCollection` if none of its features are relevant?

FantasyTeddy (approved):

PR #699, opened 2025-03-13
Refactor: add GLW.DrawPrimitive to reduce number of dependencies to OpenGL toolkit

J005t67:
I introduced functions for drawing OpenGL primitives in GLW (OpenGL Wrapper)
I replaced some code with calls to the new functions.
As a result, the number of direct dependencies to the current ToolKit is lowered from ~1600 to ~1500

### DIFF
--- a/SourceCode/AgOpenGPS.Core/Drawing/Colors.cs
+++ b/SourceCode/AgOpenGPS.Core/Drawing/Colors.cs
@@ -25,7 +25,7 @@ namespace AgOpenGPS.Core.Drawing
 
         static public ColorRgb HitchColor = new ColorRgb(0.765f, 0.76f, 0.32f);
         static public ColorRgb HitchTrailingColor = new ColorRgb(0.7f, 0.4f, 0.2f);
-        static public ColorRgb HitchRigidColor = new ColorRgb(1.237f, 0.037f, 0.0397f);
+        static public ColorRgb HitchRigidColor = new ColorRgb(0.237f, 0.037f, 0.0397f);
 
         static public ColorRgb SvenArrowColor = new ColorRgb(0.95f, 0.95f, 0.10f);
 
--- a/SourceCode/AgOpenGPS.Core/Drawing/Colors.cs
+++ b/SourceCode/AgOpenGPS.Core/Drawing/Colors.cs
@@ -1,4 +1,5 @@
 ﻿using AgOpenGPS.Core.Models;
+
 namespace AgOpenGPS.Core.Drawing
 {
     static public class Colors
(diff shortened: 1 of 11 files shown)


SourceCode/AgOpenGPS.Core/Models/Base/ColorRgba.cs
```diff
@@ -13,6 +13,16 @@ public ColorRgba(byte red, byte green, byte blue, byte alpha)
...
         }
 
+        public ColorRgba(ColorRgb colorRgb, float alpha) :
+            this(
+                colorRgb.Red,
+                colorRgb.Green,
+                colorRgb.Blue,
+                ColorRgba.FloatToByte(alpha))
+        {
+            if (alpha < 0.0f || 255.0 < alpha) throw new ArgumentOutOfRangeException(nameof(alpha), "Argument out of range");
```
FantasyTeddy, line 23: Is this range check correct? Shouldn't it be `alpha < 0.0f || alpha > 1.0f`? (Maybe it would help if we didn't use the base constructor, but set the values directly?)

SourceCode/AgOpenGPS.Core/Drawing/Colors.cs
```diff
@@ -9,19 +10,33 @@ static public class Colors
...
         static public ColorRgba TramDotAutomaticControlBitOffColor = new ColorRgba(0.9f, 0.0f, 0.0f, 0.53f);
         static public ColorRgba TramDotAutomaticControlBitOnColor = new ColorRgba(0.29f, 0.990f, 0.290f, 0.983f);
         static public ColorRgb TramMarkerOnColor = new ColorRgb(0.0f, 0.900f, 0.39630f);
 
-        static public ColorRgb FlagRedColor = Red;
-        static public ColorRgb FlagGreenColor = Green;
-        static public ColorRgb FlagYellowColor = Yellow;
+        static public ColorRgb WorldGridDayColor = Gray012;
+        static public ColorRgb WorldGridNightColor = Gray025;
+        static public ColorRgb WorldGridColor(bool isDay) { return isDay ? WorldGridDayColor : WorldGridNightColor; }
```
FantasyTeddy, line 40: I think it would be cleaner, if we only hold actual color values in this `Colors` class. Any computation based on those colors should be done outside.

J005t67: OK, moved it back to WorldGrid.

SourceCode/AgOpenGPS.Core/Models/Base/ColorRgba.cs
FantasyTeddy: I agree, but the "rule" about the base constructor only really makes sense, if there is any meaningful initialization happening (which I would argue is not the case here).
The problem with your current solution is, that it is much harder to figure out which value (red, green, blue, alpha) is actually out of range.

J005t67: Changed, and changed RgbColor also. It works: it caught a bug in one of the colors, (that was already present before I moved the color to Colors.cs.)

PR #703, opened 2025-03-15
Use separate menu entries/forms for profile creation/loading

FantasyTeddy:
The current way of creating/saving/loading vehicle profiles in the configuration is confusing for several reasons:
- The vehicle configuration entails more settings than just the parameters in the configuration (e.g. steer settings) which is not obvious by having the options to create/save/load vehicles inside the configuration.
- Configuration settings are saved automatically and the options "New" and "Save As" are confusing for many people.
- Creating a new vehicle profile based on an already existing one is a two-step process:
  1. Load the existing vehicle
  2. Save as a new vehicle

This PR moves the entire profile creation/loading mechanism to a new top-level menu entry:
image

The "New Profile" form lists all the possible options in the table:
- Create a new profile based on the default settings (choose "&lt;Empty Profile&gt;")
- Copy the settings of the current profile (choose the profile marked as "(Current)", which is also selected per default)
- Copy the settings of another already existing profile

image

The "Load Profile" form also includes the "Delete" option:

image


lansalot:
I think Current and Existing mean practically the same thing to the farmer tho? I understand what you're saying in terms of it showing a list only if Existing is ticked, but this seems to be a little over-engineered?

There is still a "default profile" which I think we could lock - it's the out-of-the-box setting, so then operationally you have all your vehicles. Icons for + mean "copy this existing profile to new", and if they want a new profile from default, they simply click that locked profile? Or, when they click the "default profile", it immediately prompts them to give it a new name?

farmerbriantee:
One thing was pretty clear in discussions of this "problem" was that loading occurred where the settings are.

FantasyTeddy:
@lansalot
> I think Current and Existing mean practically the same thing to the farmer tho? I understand what you're saying in terms of it showing a list only if Existing is ticked, but this seems to be a little over-engineered?

What is your suggestion?

> There is still a "default profile" which I think we could lock - it's the out-of-the-box setting, so then operationally you have all your vehicles. Icons for + mean "copy this existing profile to new", and if they want a new profile from default, they simply click that locked profile and it immediately prompts them to give it a new name? Locking it means you cannot edit or change it, or even load it. It must be immediately saved as a new name. This is similar to the first-run when registry entries are not complete.

The "Default Vehicle" is an artifact from a previous migration. When starting with a "clean" AgOpenGPS instance, the list will be completely empty.
Can you implement your suggestion?

---

@farmerbriantee

> One thing was pretty clear in discussions of this "problem" was that loading occurred where the settings are.

Are you saying that the profiles should remain in the configuration? What about the fact that not everything that belongs to a profile is in the configuration (like steer settings)?

> Have you considered expanding the current strings to define more of what the action is ? Instead of New, use "New from Default settings". Use more words basically.

I don't think that this will help a lot, since the usability is not very good.

FantasyTeddy:
What if we used the `ListView` in "New Profile" like this:
image
Would that be any better?

FantasyTeddy:
I updated the description of this PR to reflect the current state.

@AgOpenGPS-Official/aog-coders: The PR is now ready for review.

richardklasens (approved):

PR #704, opened 2025-03-15
Fix weird FormConfig resource bug

J005t67:


J005t67:
Abandonned. Moved Brand reosurce back  in stead.

PR #707, opened 2025-03-16
Load flags using a text or CSV file

Rob2029:
(Updated) Added a button in the Load flags by lat lon section to be able to load multiple points from a text or CSV file
This helps users to load multiple point without having to type each coordinate 
The format is:
latitude,longitude,flagColor,flagName

Also added a button to save a txt file with the same format

for example:
Latitude,Longitude,Color,Notes
53.4370564,-111.160047,0,pointA
53.4380564,-111.160047,0,
53.4390564,-111.160047,2,pointC

flagColor is 0 for red, 1 for green and 2 for yellow
flagName is optional, but the row must have a comma, if not that line wont be loaded

flagsscreenCapture

### DIFF
--- a/SourceCode/GPS/Forms/Field/FormEnterFlag.Designer.cs
+++ b/SourceCode/GPS/Forms/Field/FormEnterFlag.Designer.cs
@@ -39,6 +39,7 @@
             this.btnYellow = new System.Windows.Forms.Button();
             this.btnRed = new System.Windows.Forms.Button();
             this.btnGreen = new System.Windows.Forms.Button();
+            this.btnLoadFlags = new System.Windows.Forms.Button();
             ((System.ComponentModel.ISupportInitialize)(this.nudLatitude)).BeginInit();
             ((System.ComponentModel.ISupportInitialize)(this.nudLongitude)).BeginInit();
             this.SuspendLayout();
@@ -220,6 +221,17 @@
             this.btnGreen.UseVisualStyleBackColor = false;
             this.btnGreen.Click += new System.EventHandler(this.btnRed_Click);
             // 
+            // btnLoadFlags
+            // 
+            this.btnLoadFlags.Font = new System.Drawing.Font("Tahoma", 20.25F, System.Drawing.FontStyle.Bold, System.Drawing.GraphicsUnit.Point, ((byte)(0)));
+            this.btnLoadFlags.Location = new System.Drawing.Point(44, 168);
+            this.btnLoadFlags.Name = "btnLoadFlags";
+            this.btnLoadFlags.Size = new System.Drawing.Size(196, 71);
+            this.btnLoadFlags.TabIndex = 434;
+            this.btnLoadFlags.Text = "Load Flags";
+            this.btnLoadFlags.UseVisualStyleBackColor = true;
+            this.btnLoadFlags.Click += new System.EventHandler(this.btnLoadFlags_Click);
+            // 
             // FormEnterFlag
             // 
             this.AutoScaleDimensions = new System.Drawing.SizeF(6F, 13F);
@@ -227,6 +239,7 @@
             this.BackColor = System.Drawing.Color.WhiteSmoke;
             this.ClientSize = new System.Drawing.Size(771, 266);
             this.ControlBox = false;
+            this.Controls.Add(this.btnLoadFlags);
             this.Controls.Add(this.btnGreen);
             this.Controls.Add(this.btnRed);
             this.Controls.Add(this.btnYellow);
@@ -263,5 +276,6 @@
         private System.Windows.Forms.Button btnYellow;
         private System.Windows.Forms.Button btnRed;
         private System.Windows.Forms.Button btnGreen;
+        private System.Windows.Forms.Button btnLoadFlags;
     }
 }
--- a/SourceCode/GPS/Forms/Field/FormEnterFlag.Designer.cs
+++ b/SourceCode/GPS/Forms/Field/FormEnterFlag.Designer.cs
@@ -32,14 +32,15 @@
             this.label8 = new System.Windows.Forms.Label();
             this.label1 = new System.Windows.Forms.Label();
             this.label18 = new System.Windows.Forms.Label();
-            this.nudLatitude = new AgOpenGPS.NudlessNumericUpDown();
-            this.nudLongitude = new AgOpenGPS.NudlessNumericUpDown();
             this.labelPoint = new System.Windows.Forms.Label();
             this.btnCancel = new System.Windows.Forms.Button();
             this.btnYellow = new System.Windows.Forms.Button();
             this.btnRed = new System.Windows.Forms.Button();
             this.btnGreen = new System.Windows.Forms.Button();
             this.btnLoadFlags = new System.Windows.Forms.Button();
+            this.nudLatitude = new AgOpenGPS.NudlessNumericUpDown();
+            this.nudLongitude = new AgOpenGPS.NudlessNumericUpDown();
+            this.btnSaveFlags = new System.Windows.Forms.Button();
             ((System.ComponentModel.ISupportInitialize)(this.nudLatitude)).BeginInit();
             ((System.ComponentModel.ISupportInitialize)(this.nudLongitude)).BeginInit();
             this.SuspendLayout();
@@ -49,9 +50,10 @@
             this.label9.AutoSize = true;
             this.label9.Font = new System.Drawing.Font("Tahoma", 15.75F);
             this.label9.ImeMode = System.Windows.Forms.ImeMode.NoControl;
-            this.label9.Location = new System.Drawing.Point(510, 40);
+            this.label9.Location = new System.Drawing.Point(680, 49);
+            this.label9.Margin = new System.Windows.Forms.Padding(4, 0, 4, 0);
             this.label9.Name = "label9";
-            this.label9.Size = new System.Drawing.Size(163, 25);
+            this.label9.Size = new System.Drawing.Size(211, 33);
             this.label9.TabIndex = 202;
             this.label9.Text = "( +180 to -180 )";
             // 
@@ -60,9 +62,10 @@
             this.label8.AutoSize = true;
             this.label8.Font = new System.Drawing.Font("Tahoma", 15.75F);
             this.label8.ImeMode = System.Windows.Forms.ImeMode.NoControl;
-            this.label8.Location = new System.Drawing.Point(215, 40);
+            this.label8.Location = new System.Drawing.Point(287, 49);
+            this.label8.Margin = new System.Windows.Forms.Padding(4, 0, 4, 0);
             this.label8.Name = "label8";
-            this.label8.Size = new System.Drawing.Size(134, 25);
+            this.label8.Size = new System.Drawing.Size(173, 33);
             this.label8.TabIndex = 201;
             this.label8.Text = "( +90 to -90)";
             // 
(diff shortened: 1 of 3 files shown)


SourceCode/GPS/Forms/Field/FormEnterFlag.cs
```diff
@@ -103,5 +103,54 @@ private void btnRed_Click(object sender, EventArgs e)
...
+            fileDialog.Filter = "Text Document | *.txt| CSV Document | *.csv";
+            fileDialog.Title = "Please select points file";
+            fileDialog.Multiselect = false;
+
+            if (fileDialog.ShowDialog() == DialogResult.OK)
+            {
+                string filePath = fileDialog.FileName;
+                try
+                {
+                    string[] lines = System.IO.File.ReadAllLines(filePath);
```
FantasyTeddy, line 120: It would be great, if we could move the entire reading of the file (everything inside the `try` block) into a separate class (e.g. `FlagFileReader`)
This would allow us to write tests to ensure that everything works as expected.

SourceCode/GPS/Forms/Field/FormEnterFlag.cs
```diff
@@ -103,5 +103,54 @@ private void btnRed_Click(object sender, EventArgs e)
...
+
+                    for (int i = 1; i < lines.Length; i++)
```
FantasyTeddy, line 122:
```suggestion
                    foreach (string line in lines)
```

FantasyTeddy (commented):
Three general questions:
1. AgOpenGPS already writes/reads a "Flags.txt" file. Why is this solution needed if this functionality already kind of exists?
2. Where does the text/CSV file usually come from? Is that generated by another application?
3. Do you also plan to add the option for exporting/saving the current flag list?

SourceCode/GPS/Forms/Field/FormEnterFlag.cs
Rob2029: Regarding your questions:
1. This solution is needed when you want to add multiple flags using only latitude and longitude, since the northing and easting is calculated by AOG. 
2. The csv or text file comes from a GIS software
3. The exporting of flags is already done by AOG, I think it should be posible to add the importing of flags from another field

AOG does not add a header row to the flags.txt, if we add a header we can extract lat, lon, color, name and easily import flags from another field.
This feature comes from a discussion in january about adding multiple flags generated outside AOG, it is not posible to just modify the flags.txt file since it needs easting and northing

SourceCode/GPS/Forms/Field/FormEnterFlag.cs
Rob2029: I was using foreach but needed to ignore the header row because it show the error message when trying to parse the columns of the header so I change it to the for starting at 1, I will think of a way to avoid this

FantasyTeddy:
> 1. This solution is needed when you want to add multiple flags using only latitude and longitude, since the northing and easting is calculated by AOG.

Okay, fair point.

> 2. The csv or text file comes from a GIS software

👍 

> 3. The exporting of flags is already done by AOG, I think it should be posible to add the importing of flags from another field

Do you plan to add this functionality too?

SourceCode/GPS/Forms/Field/FormEnterFlag.cs
FantasyTeddy: This new class (which I would call `FlagFileReader`) would have to be created by you. So you can make it do exactly what you want.

SourceCode/GPS/Forms/Field/FormEnterFlag.cs
FantasyTeddy: Ah, I did not notice that you start from 1.
There is a handy LINQ method [`Skip`](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.skip) that you can probably use here.

SourceCode/GPS/Forms/Field/FormEnterFlag.cs
Rob2029: Ok I'll give it a try

SourceCode/GPS/Forms/Field/FormEnterFlag.cs
Rob2029: Thanks I'll use it

Rob2029:
> > 1. This solution is needed when you want to add multiple flags using only latitude and longitude, since the northing and easting is calculated by AOG.
> 
> Okay, fair point.
> 
> > 2. The csv or text file comes from a GIS software
> 
> 👍
> 
> > 3. The exporting of flags is already done by AOG, I think it should be posible to add the importing of flags from another field
> 
> Do you plan to add this functionality too?

Yes I plan to add this functionality, I'll added here first and then I'll try to code it in the class you suggested

SourceCode/GPS/Forms/Field/FormEnterFlag.cs
Rob2029: I have two questions, where can I code that class?
Also can I add a header to the flags.txt file? in order to be able to read it using that header to pick the right columns when loading the flags.txt in order to add the functionality to load flags from other fields

SourceCode/GPS/Forms/SaveOpen.Designer.cs
```diff
@@ -1882,6 +1882,7 @@ public void FileSaveFlags()
 
                     int count2 = flagPts.Count;
                     writer.WriteLine(count2);
+                    writer.WriteLine("Latitude, Longitude, Elevation, Easting, Northing, Heading, Color, ID, Notes");
```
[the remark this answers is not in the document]
Rob2029: I remove the header, did not think in backwards compatibility
I think the header is very useful when loading files in GIS software, but not a requierment

SourceCode/GPS/Forms/Field/FormEnterFlag.cs
Rob2029: ok I will give it a try to code the class
About loading flags.txt file I tought you asked for that. I already write the functionality to be ablee to read a flags.txt file, extract latitude, longitude, color and notes from it and then save it on the current flags.txt but i can remove it so it only can load points from a text file
I did that so the user can have a new field and be able to load the flags from another field. Not sure if the "From existing" feature does this

FantasyTeddy: > About loading flags.txt file I tought you asked for that. I already write the functionality to be ablee to read a flags.txt file, extract latitude, longitude, color and notes from it and then save it on the current flags.txt but i can remove it so it only can load points from a text file
> I did that so the user can have a new field and be able to load the flags from another field. Not sure if the "From existing" feature does this

Oh I'm sorry, I think this was a misunderstanding.
My question regarding the additional functionality was just to add the inverse of what you already implemented: if you can import text/CSV files that contain flag information, it would be nice if you could also export the same information (probably in the same format).

Rob2029: Oh ok, I think is not necessary since AOG already saves a flags file, but I can add a button to save flags information in the same format. Should I keep the functionality to load flags.txt from other fields?

FantasyTeddy: > ..., but I can add a button to save flags information in the same format.

I think that would be a nice addition!

> Should I keep the functionality to load flags.txt from other fields?

I don't think that is necessary. If really needed, one could just copy the "Flags.txt" over from another field.

Rob2029: Ok I will add a Save button
I you just copy the Flag.txt file to another field folder you lose the flags in the current field, but loading the flags.txt file from this button adds each flag to the current list so I think it is nice to have that option as well. Just thinking in the posible situations

FantasyTeddy: >If you just copy the Flag.txt file to another field folder you lose the flags in the current field, but loading the flags.txt file from this button adds each flag to the current list so I think it is nice to have that option as well. Just thinking in the posible situations

Of course, you are right.
But theoretically, if you also add an export option, one could work around this limitation by first exporting the flags from an existing field and then importing them into the target field.

Rob2029: Yes, so I will remove the loading of Flags.txt and add an export/save button

FantasyTeddy: This looks very good so far!

Now the only thing left is to extract the core functionality into the separate class (which should maybe not be called `FlagFileReader` anymore, but rather `FlagFileStreamer` like the already existing ones in `AgOpenGPS.Core/Streamers`).

Rob2029: Thanks! I am trying to do that but my coding skills are not on par.
I will take a close look on how the other similar classes work

FantasyTeddy: Let me know if you need any help with that.

Rob2029: Thanks I do need help. I am trying to use similar code to the `FlagListStreamer` class but since the Load function has to read a txt file and update the flag list is complicated, at least for me. I am not a programmer

SourceCode/GPS/Forms/Field/FormEnterFlag.cs
```diff
@@ -103,5 +109,105 @@ private void btnRed_Click(object sender, EventArgs e)
...
+            if (fileDialog.ShowDialog() == DialogResult.OK)
+            {
+                //use Streamwriter to create file
+                using (StreamWriter writer = new StreamWriter(fileDialog.FileName))
+                {
+                    try
+                    {
+                        writer.WriteLine("Latitude,Longitude,Color,Notes");
+
+                        int count2 = Convert.ToInt32(mf.flagPts.Count);
```
FantasyTeddy, line 188: Why do this conversion? `mf.flagPts.Count` is already of type `int`.

SourceCode/GPS/Forms/Field/FormEnterFlag.cs
```diff
@@ -103,5 +109,105 @@ private void btnRed_Click(object sender, EventArgs e)
...
+                        writer.WriteLine(count2);
```
FantasyTeddy, line 189: Is it necessary to write the number of flags into the output file? (Doesn't that lead to the file not being able to load afterwards?)

SourceCode/GPS/Forms/Field/FormEnterFlag.Designer.cs
FantasyTeddy, line 1: The form is somehow cut at the bottom:
image
Do you have this problem too?

FantasyTeddy (changes requested):
I originally wanted to take a closer look at the code after the extraction into a separate class. Since we are now not doing that, I added a few more comments.

SourceCode/GPS/Forms/Field/FormEnterFlag.cs
Rob2029: not necessary, that was a slip

SourceCode/GPS/Forms/Field/FormEnterFlag.Designer.cs
Rob2029: Yes, not sure why since I did not change any properties of the form. It does not happen in v6.6.2

FantasyTeddy: Very weird... I fixed it, but I'm unsure what caused this to happen...

SourceCode/GPS/Forms/Field/FormEnterFlag.cs
```diff
@@ -103,5 +107,97 @@ private void btnRed_Click(object sender, EventArgs e)
...
+            {
+                using (StreamWriter writer = new StreamWriter(fileDialog.FileName))
+                {
+                    try
+                    {
+                        writer.WriteLine("Latitude,Longitude,Color,Notes");
+
+                        int count = mf.flagPts.Count;
+                        
+                        for (int i = 0; i < count; i++)
```
FantasyTeddy, line 183:
```suggestion
                        foreach (CFlag flag in mf.flagPts)
```

FantasyTeddy (changes requested):

SourceCode/GPS/Forms/Field/FormEnterFlag.cs
FantasyTeddy: I'm sorry, this suggestion requires some additional work. You have to use the `flag` variable inside the loop in place of `mf.flagPts[i]` (otherwise it does not compile).

FantasyTeddy (approved):
Thank you very much for your patience and the contribution!

Rob2029:
> Thank you very much for your patience and the contribution!

Thank you for helping me this much!

PR #709, opened 2025-03-17
Remove unused tools (OGL.exe & UDP_SimTool.exe)

FantasyTeddy:
These tools were removed:
- **OGL.exe**: This is only an OpenGL test application without any real benefit for the user.
- **UDP_SimTool.exe**: This has been replaced by the ModSim.exe tool.


J005t67 (approved):

Davidwedel:
I know this was already closed, but ModSim and Simtool did differ slightly. At least on the last versions I used, Simtool could output every sentence AOG could parse, which was nice for testing specific sentences, or how AOG would react to many more sentences(all at once). ModSim just output's one type (I think panda but don't remember)

FantasyTeddy:
Oh, that is good to know.

I can see a few options here:
1. Accept the missing functionality (probably not the best option)
2. Revert this commit partially, to restore UDP_SimTool.exe as it was before
3. Include the missing functionality in ModSim (I don't know how much work that would be...)
4. Add the UDP_SimTool with source code (like we did for ModSim & GPS_Out)

My favorite solution would probably be number 4, because this would allow us to merge ModSim & UDP_SimTool together more easily in the future.
@Davidwedel: What do you think?

richardklasens:
I’m not sure but I believe you can select all sentences as well in ModSim

FantasyTeddy:
image
image
Yes, looks pretty much the same to me...

richardklasens:
And ModSim acts as a steering module means it follows the steering commands from AOG

FantasyTeddy:
Okay, so let's keep it as it is now (with UDP_SimTool removed).
Do you agree @Davidwedel?

Davidwedel:
Yeah looks great. There must have been an update to ModSim since I last used it.

PR #714, opened 2025-03-20
Refactor ISOXML export

FantasyTeddy:
- Use the [Dev4Agriculture.ISO11783.ISOXML](https://www.nuget.org/packages/Dev4Agriculture.ISO11783.ISOXML) NuGet package to generate the ISOXML file
- Combine the V3 and V4 export methods to reduce the amount of duplicated code
- Use the correct version number of AgOpenGPS as the `ManagementSoftwareVersion`
- Only generate each track once (was duplicated before for AB & Curve)


richardklasens (commented):
I’d like to test this in an AgLeader environment before it gets merged.

J005t67 (commented):
This is a big step forward regarding maintainability. I checked each commit individually, but I have no comments. Your changes will be of great help when I need to do a few more to these files when I will switch to the refctored implementations of boundary, track etc.
You would have my approval, but I don't like to approve before Richard tested it.

FantasyTeddy:
Thank you very much for the review.

I'm looking forward to the feedback from @richardklasens testing with AgLeader.

richardklasens:
> I'm looking forward to the feedback from @richardklasens testing with AgLeader.

Done Testing, I've imported v4 to AgLeader's SMS software, it seems to load fine. The headline lines arent shown up as separate headland but that can be related to the SMS software because thats a hork to work with.
WIth v3 import the guidancelines don't import but again this can be related to the AgLeader Software.
For now it's approved

richardklasens (approved):

FantasyTeddy:
Thank you for testing!

PR #718, opened 2025-03-23
Fixes 6.7 to release/6.7

richardklasens:
Fixes/6.7


SourceCode/GPS/Forms/Settings/FormConfig.Designer.cs
FantasyTeddy, line 1: This form has a huge amount of changes. Are you sure, that this is intentional?

richardklasens: probably because i reverted a particular commit and its auto-generated code... this can maybe break other things cause i haven't checked the whole form.

FantasyTeddy: No, don't redo the revert. That was not the commit that caused the big changes.
The next commit about the "New from Default" caused those.

PR #719, opened 2025-03-25
Refactor/isday ismetric

J005t67:
Add infrastructure for IsDay and IsMetric and apply in existing viewmodels

### DIFF
--- a/SourceCode/AgOpenGPS.Core/ViewModels/Base/DayNightAndUnitsViewModel.cs
+++ b/SourceCode/AgOpenGPS.Core/ViewModels/Base/DayNightAndUnitsViewModel.cs
@@ -31,12 +31,12 @@ namespace AgOpenGPS.Core.ViewModels
 
         public bool IsDay
         {
-            get { return _isMetric; }
+            get { return _isDay; }
             set
             {
-                if (value != _isMetric)
+                if (value != _isDay)
                 {
-                    _isMetric = value;
+                    _isDay = value;
                     NotifyAllPropertiesChanged();
                 }
             }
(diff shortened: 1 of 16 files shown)


SourceCode/AgOpenGPS.Core/ViewModels/Base/DayNightAndUnitsViewModel.cs
```diff
@@ -0,0 +1,53 @@
+﻿using AgLibrary.ViewModels;
+using System.Collections.ObjectModel;
+
+namespace AgOpenGPS.Core.ViewModels
+{
+    // The class DayNightAndUnitsViewModel implements the properties IsDay and IsMetric.
+    // Moreover, an instance of this class propagates changes to these properites to its childs.
+    public class DayNightAndUnitsViewModel : ViewModel
```
FantasyTeddy, line 8: Are you sure that using such a base ViewModel with those properties is the way to go? What if we need access to those features inside something other than a ViewModel?

J005t67: As far as I can see (when I search for isDay and isMetric), these properties are only needed in the views. So exposing them as ViewModel properties is the way to get them accros. To keep the databindings as easy as possible, I decided to let each viewmodel expose these properties.

FantasyTeddy: Okay.
How do you plan to do the update when one of these properties changes? How will that propagate to all the necessary ViewModels?

J005t67: Sorry, forgot to implement the propagation. The changes will simply be propagated to all ViewModels.

PR #725, opened 2025-04-02
Fixes/6.7

richardklasens:
Added fixes for release
- fix glitch in autodual swap 
- put back New option which is now called New From Default
- fix a glitch in settings when there were older settings files with capitalized featuresettings

### DIFF
--- a/SourceCode/AgLibrary/Settings/XmlSettingsHandler.cs
+++ b/SourceCode/AgLibrary/Settings/XmlSettingsHandler.cs
@@ -152,11 +152,32 @@ namespace AgLibrary.Settings
             }
             else if (fieldType.IsClass)
             {
-                // Deserialize XML into the custom object
-                var serializer = new XmlSerializer(fieldType);
-                var nestedObj = serializer.Deserialize(reader);
-                pinfo.SetValue(obj, nestedObj);
+                try
+                {
+                    
+                    if (reader.NodeType == XmlNodeType.Element && !reader.IsEmptyElement)
+                    {
+                        string innerXml = reader.ReadOuterXml().Trim();
+                        innerXml = innerXml.Replace(">True<", ">true<").Replace(">False<", ">false<");
+
+                        if (!string.IsNullOrEmpty(innerXml))
+                        {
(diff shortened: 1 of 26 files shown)


SourceCode/GPS/Forms/Position.designer.cs
```diff
@@ -985,9 +985,11 @@ public void UpdateFixPosition()
 
                 //setAngVel = glm.toDegrees(setAngVel);
 
-                if (isChangingDirection && ahrs.imuHeading == 99999)
-                    p_254.pgn[p_254.status] = 0;
-
+                if (!Properties.Settings.Default.setAutoSwitchDualFixOn)
+                {
+                    if (isChangingDirection && ahrs.imuHeading == 99999)
```
J005t67, line 990: Why not convert this nested if to a single if statement wit two times && ?

J005t67:
I see that this review request is only about a merge, not about new changes. If you prefer to ignore some of my review comments now, for efficiency (I should have made them earlier), I understand.

SourceCode/AgLibrary/Settings/XmlSettingsHandler.cs
```diff
@@ -152,11 +152,32 @@ private static bool SetFieldValue(FieldInfo pinfo, XmlTextReader reader, object
...
+                                object nestedObj = serializer.Deserialize(stringReader);
+                                pinfo.SetValue(obj, nestedObj);
+                               
+                            }
+                        }
+                    }
+                }
+                catch (Exception ex)
+                {
+                    Debug.WriteLine($"❌ Fout bij deserialisatie van veld {pinfo.Name}: {ex.Message}");
```
FantasyTeddy, line 177: I don't think that an exception here should be handled by just doing `Debug.WriteLine`. I would consider throwing a new exception, or not catching the exception at all.
Maybe we need to re-visit the error handling here at some point, because it is basically impossible to figure out what went wrong with the current implementation.

richardklasens: It's a left over from my testing. I will remove it

FantasyTeddy (approved):

PR #726, opened 2025-04-02
Refactor: filter fields based on distance

J005t67:
Note that this is a small change (improvement) to the spec too.
The table will not filter based on a fixed maximimum distance of 10 km, (with the risk that many field won't fit the table) Instead, the table simply displays the 6 fields that are most nearby.

### DIFF
--- a/SourceCode/AgOpenGPS.Core/ViewModels/StartNewFieldViewModels/NearFieldTableViewModel.cs
+++ b/SourceCode/AgOpenGPS.Core/ViewModels/StartNewFieldViewModels/NearFieldTableViewModel.cs
@@ -10,7 +10,7 @@ namespace AgOpenGPS.Core.ViewModels
             SortMode = FieldSortMode.ByDistance;
         }
 
-        public new void UpdateFields()
+        public override void UpdateFields()
         {
             Collection<FieldDescriptionViewModel> viewModels = new Collection<FieldDescriptionViewModel>();
             var descriptions = _appModel.Fields.GetFieldDescriptions();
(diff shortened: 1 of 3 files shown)


SourceCode/AgOpenGPS.Core/ViewModels/StartNewFieldViewModels/NearFieldTableViewModel.cs
```diff
@@ -0,0 +1,38 @@
...
+namespace AgOpenGPS.Core.ViewModels
+{
+    public class NearFieldTableViewModel : FieldTableViewModel
+    {
+        public NearFieldTableViewModel(ApplicationModel appModel) : base(appModel)
+        {
+            SortMode = FieldSortMode.ByDistance;
+        }
+
+        public new void UpdateFields()
```
FantasyTeddy, line 14: This is an immediate red flag for me.
Are you sure that this is the way to go here? Could we not specify the number of fields to take in a property and do this in the `FieldTableViewModel`?

J005t67: OK, removed the hardcoded '6'. Simply let the view display the number of rows that fit.

FantasyTeddy: Oh I'm sorry for the confusion.
I'm not against limiting the number of rows, but using the `new` keyword to "overwrite" the method.

J005t67: Oops,  indeed 'new' does not what I intended.

FantasyTeddy (approved):

PR #733, opened 2025-04-08
Refactor/worldgrid

J005t67:
I am still working towards the moment that I can start using the refactored Field aspect in the Core project.

I will start with the Flags. However, the code that will draw the flags has dependencies to the class CCamera and CFonts. so I have to move these to the Core project first. Because Camera has an unwanted dependency to WorldGrid, I decided to solve that first.

Note: this PR does more that removing the dependency between CCamera and WorldGrid.


FantasyTeddy (approved):
Very well structured commits and changes. Nice work!

PR #737, opened 2025-04-10
Refactor/camera

J005t67:
I want to move the Camera class to the core project, but decided to focus on variable names first.
Next PR will remove unwanted dependencies and move the class to the core project.

### DIFF
--- a/SourceCode/GPS/Forms/FormPan.cs
+++ b/SourceCode/GPS/Forms/FormPan.cs
@@ -9,59 +9,62 @@ namespace AgOpenGPS
     public partial class FormPan : Form
     {
         private readonly FormGPS mf = null;
+        private const double adjustFactor = 0.04;
 
         public FormPan(Form callingForm)
         {
             mf = callingForm as FormGPS;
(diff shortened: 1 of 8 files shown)


SourceCode/GPS/Forms/FormPan.cs
```diff
@@ -9,59 +9,62 @@ namespace AgOpenGPS
     public partial class FormPan : Form
     {
         private readonly FormGPS mf = null;
+        private const double adjustFactor = 0.04;
```
FantasyTeddy, line 12: I know we don't have any naming guidelines, but I prefer to use PascalCase for constant values (as does [Microsoft](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names#naming-conventions)).

J005t67: Just like me ;-)  => Changed

FantasyTeddy (approved):

FantasyTeddy (approved):

PR #740, opened 2025-04-12
Choose the first visible track when track button is pressed

### LINKED ISSUE #739: Hidden track is used when first opening field
When the "tracks" button is pressed after opening a field, the first track in the list is selected, regardless of whether it is marked as "hidden".

In the example screenshot below, you can see five tracks on this field. The first one, **AB 268.75343°** is marked as hidden (red box). After opening the field and pressing the track button, that hidden track is selected. 

It would be better if the first non-hidden track is selected (in this example **AB 3.88507°**).

A work-around is to move any hidden tracks down from position 1 in the list.

Image

mattmccormack:
Tested this out in the simulator and it works as expected. Resolves #739.

When `btnTrack` is pressed:

- [X] All tracks marked "hidden" -> track index 0 selected (as before)
- [X] First and second track marked "hidden", third and fourth marked "visible" -> track index 2 selected
- [X] All tracks marked "visible" -> track index 0 selected (as before)

### DIFF
--- a/SourceCode/GPS/Forms/Controls.Designer.cs
+++ b/SourceCode/GPS/Forms/Controls.Designer.cs
@@ -77,7 +77,12 @@ namespace AgOpenGPS
             {
                 if (trk.idx == -1)
                 {
-                    trk.idx = 0;
+                    //find index of first visible track
+                    trk.idx = trk.gArr.FindIndex(track => track.isVisible);
+
+                    //otherwise default to index 0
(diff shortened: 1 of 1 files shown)


richardklasens (approved):

mattmccormack:
@richardklasens could this be a candidate for `hotfix/6.7`?

richardklasens:
Yes good idea!

PR #742, opened 2025-04-12
Properly close Quick AB Curve creator form if insufficient points are recorded

### LINKED ISSUE #738: Clicking green tick after adding an A point for a curve hangs AOG
New user found this the other day, tested in 6.7.0 releases and its still there. Limited PC access atm to investigate further.

Steps to reproduce:
- open field
- Tracks, add, AB curve
- enter A point, Green tick becomes active
- immediately click the tick
- White box appears where dialog was, hangs apps, no way for user to close ago without terminating process
-

mattmccormack:
This fixes #738 by making sure the form closes

### DIFF
--- a/SourceCode/GPS/Forms/Guidance/FormQuickAB.cs
+++ b/SourceCode/GPS/Forms/Guidance/FormQuickAB.cs
@@ -227,10 +227,12 @@ namespace AgOpenGPS
             }
             else
             {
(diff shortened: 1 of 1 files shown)


FantasyTeddy (approved):
I think the perfect solution would be to not enable the OK button until enough points are recorded.

But this is a very valid solution too :)

PR #746, opened 2025-04-14
Remove CPolygon

J005t67:


J005t67:
> Why remove? It’s unused?

It is unused. As you can see, the build is still fine after I removed it. 
I removed it, because it is unused and I don't plan to use any of it. I don't like the methods that require separate x and y arguments. We have more explicit types available now: GeoCoord, GeoDir etc that better express what the intention  is, (This helps to spot bugs and some mistakes are even caught by the compiler that way.)

richardklasens (approved):

PR #747, opened 2025-04-15
Refactor/move fonts to core

J005t67:


SourceCode/AgOpenGPS.Core/DrawLib/Texture2D.cs
```diff
@@ -7,15 +7,18 @@ namespace AgOpenGPS.Core.DrawLib
...
-            _textureId = GL.GenTexture();
-            if (bitmap != null) SetBitmap(bitmap);
+            _bitmap = bitmap;
+            // To avoid crashes during start-upup (when no OpenGL context has been created yet),
+            // delay creation of texture to the first call to Bind() 
         }
 
         public void Bind()
         {
+            if (0 == _textureId) CreateTexture();
```
FantasyTeddy, line 21: I will never get used to the constant being on the left of the equality operator :D

SourceCode/AgOpenGPS.Core/DrawLib/Texture2D.cs
```diff
@@ -53,6 +56,7 @@ public void DrawCentered(
 
         public void SetBitmap(Bitmap bitmap)
         {
+            if (0 == _textureId) CreateTexture();
             GL.BindTexture(TextureTarget.Texture2D, _textureId);
```
FantasyTeddy, line 60: Shouldn't we maybe call `Bind` here instead?

J005t67: That was already a good idea before my last change. => changed

FantasyTeddy (approved):

PR #748, opened 2025-04-17
AgShare implementation

richardklasens:


SourceCode/GPS/Classes/AgShareApi.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+        Settings.Default.AgShareApiKey = key;
+        Settings.Default.Save();
+    }
+
+    public static async Task<bool> TestApiKeyAsync()
+    {
+        if (string.IsNullOrWhiteSpace(ApiKey)) return false;
+
+        var request = new HttpRequestMessage(HttpMethod.Get, "http://localhost:5000/api/isoxmlfields");
```
FantasyTeddy, line 28: The hostname and port should probably be configurable.
Additionally, I think we should only allow HTTPS requests, because we are using authentication.

SourceCode/GPS/Properties/Settings.cs
```diff
@@ -247,6 +247,8 @@ public static Settings Default
         public Size setWindow_tramLineSize = new Size(921, 676);
         public bool setAutoSwitchDualFixOn = false;
         public double setAutoSwitchDualFixSpeed = 2.0;
+        public string AgShareApiKey = "apikey";
```
FantasyTeddy, line 250: Why not use an empty string `""` as default?

SourceCode/GPS/Properties/Settings.cs
```diff
@@ -247,6 +247,8 @@ public static Settings Default
...
+        public bool PublicField = false;
```
FantasyTeddy, line 251: Which field is public? All of them? The current one?

SourceCode/GPS/Forms/SaveOpen.Designer.cs
```diff
@@ -71,11 +72,25 @@ public void ExportFieldAs_ISOXMLv4()
                     AppModel.LocalPlane,
                     trk,
                     ISO11783_TaskFile.Version.V4);
+
+                if (Settings.Default.AgShareApiKey != "apikey")
```
FantasyTeddy, line 76: I think this code does not belong into this method. I would create a new method that handles the upload to AgShare.

SourceCode/GPS/AgOpenGPS.csproj
```diff
@@ -18,7 +18,9 @@
     <RunCodeAnalysis>false</RunCodeAnalysis>
     <CodeAnalysisRuleSet>BriansRules.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
-
+  <PropertyGroup>
+  <LangVersion>latest</LangVersion>
```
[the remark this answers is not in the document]
J005t67: This is not something I am very familiar with, but my opinion anyway:
- this is not a small change, that should be part of a large PR. If we want this, we should commit and test it separately.
- I would prefer a explicit choice over a default that might change when we don't notice it or don't expect it.
- I think a read something last week that Avalonia needs 8 or higher, (or maybe it was a more general advise, to switch to a higher language version before porting to Avalonia.)
- I'm currently working on more viewmodels (about 16, one for each configurationpage) that I like to create lazy, (only create them when the user switches to the corresponding page). In this case, the conditional assignment operator '??=' would help a lot to reduce code size.
- If I rememeber correctly, language version 9 also supports autogenerated properties that will 'help' to make the Properties code more compact. To me this is very scary, and no thelpfull at all. When debugging databindings, I often start with setting a breakpoint in the getter of a property. This is not possible when the property code is autogenerated. (And I do not want a language version that allows other people to write cod ethat is hard to maintain)
Conclusion: my preference for now is to switch to language version 8 explicitly in a separate PR.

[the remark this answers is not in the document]
J005t67: Just ran into this:


[https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version]

SourceCode/GPS/Forms/AgShareSettingsForm.cs
```diff
@@ -0,0 +1,144 @@
...
+    private Label label2;
+    private TextBox txtApiKey;
+    private FormGPS formGPS;
+
+    public AgShareSettingsForm()
+    {
+        InitializeComponent();
+        txtApiKey.Text = Settings.Default.AgShareApiKey ?? "";
+    }
+        public AgShareSettingsForm(FormGPS formGPS) : this()
```
J005t67: Add empty line and correct indent.

SourceCode/GPS/Forms/AgShareSettingsForm.cs
```diff
@@ -0,0 +1,144 @@
...
+    {
+        this.formGPS = formGPS;
+    }
+    private async void btnTest_Click(object sender, EventArgs e)
```
J005t67: Add empty line

SourceCode/GPS/Forms/AgShareSettingsForm.cs
```diff
@@ -0,0 +1,144 @@
...
+    {
+        lblStatus.Text = "Verbinden...";
```
J005t67: Use translations

SourceCode/GPS/Forms/AgShareSettingsForm.cs
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    private void btnClose_Click(object sender, EventArgs e)
+    {
+        Close();
+    }
+
+    private void InitializeComponent()
+    {
+            this.txtApiKey = new System.Windows.Forms.TextBox();
```
J005t67: Correct indentation

SourceCode/GPS/Forms/AgShareSettingsForm.cs
```diff
@@ -0,0 +1,144 @@
...
+            this.MinimizeBox = false;
+            this.Name = "AgShareSettingsForm";
+            this.ShowIcon = false;
+            this.ShowInTaskbar = false;
+            this.StartPosition = System.Windows.Forms.FormStartPosition.CenterScreen;
+            this.Text = "AgShare";
+            this.TopMost = true;
+            this.ResumeLayout(false);
+            this.PerformLayout();
+
```
J005t67: Remove empty line

SourceCode/GPS/Forms/Controls.Designer.cs
```diff
@@ -1342,7 +1343,15 @@ private void enterSimCoordsToolStripMenuItem_Click(object sender, EventArgs e)
             {
                 form.ShowDialog(this);
             }
-        }                
+        }
+        private void AgShareApiMenuItem_Click(object sender, EventArgs e)
```
J005t67, line 1348: Add empty line between functions and methods

SourceCode/GPS/Forms/Controls.Designer.cs
```diff
@@ -1342,7 +1343,15 @@ private void enterSimCoordsToolStripMenuItem_Click(object sender, EventArgs e)
...
+        {
+            using (var form = new AgShareSettingsForm(this))
+            {
+                form.ShowDialog(this);
+            }
+
+        }
         private void hotKeysToolStripMenuItem_Click(object sender, EventArgs e)
```
J005t67, line 1356: Add empty line

SourceCode/GPS/Classes/AgShareUploader.cs
```diff
@@ -0,0 +1,73 @@
...
+                return existingId;
+        }
+
+        var newId = Guid.NewGuid();
+        File.WriteAllText(path, newId.ToString());
+        return newId;
+    }
+
+    public static object BuildFieldUploadJsonWithConversion(
+    string fieldName,
```
J005t67, line 32: Add indentation to function arguments

SourceCode/GPS/Classes/AgShareUploader.cs
```diff
@@ -0,0 +1,73 @@
...
+    List<vec2> localBoundary,
+    List<CTrk> tracks,
+    CNMEA converter)
+
+    {
+        var boundary = new List<List<double>>();
+        foreach (var p in localBoundary)
+        {
+            converter.ConvertLocalToWGS84(p.northing, p.easting, out double lat, out double lon);
```
J005t67: This function is deprecated I believe. Please use the one that takes a vec2 as an argument and returns a Wgs84

SourceCode/GPS/Classes/AgShareUploader.cs
```diff
@@ -0,0 +1,73 @@
...
+            boundary.Add(new List<double> { lat, lon });
+        }
+
+        var abLines = tracks.Select(t => new
+        {
+            name = t.name,
+            type = t.curvePts.Count > 0 ? "Curve" : "AB",
+            points = t.curvePts.Count > 0
+                ? t.curvePts.Select(c =>
```
J005t67: I have never seen before code that splits the ternary  operator (?  and : ) over so many lines. Better rewrite it using a if and else.

J005t67: Although LinQ expressions can be very powerfull and helpfull, not so many people can read them easily. Please be a bit empathic to your co-developers and try to rewrite this with easy readably code.

Another problem with chained Linq expressions is that, the compiler might decide to optimize and do things in a different order than you expect. This can make it hard to debug. 

Also if one is trying to figure out what is going on, and using the debugger to do so, this can result in a disapointment because of the previous remark.
So in my opinion, this code is pretty hard to maintain, (close to unmaintainable) and shoud be simplified by using easier constructs like a for loop.

SourceCode/GPS/Classes/AgShareApi.cs
```diff
@@ -24,33 +25,53 @@ public static async Task<bool> TestApiKeyAsync()
...
 
     public static async Task<bool> UploadIsoXmlFieldAsync(string fieldId, object jsonPayload)
     {
         if (string.IsNullOrWhiteSpace(ApiKey)) return false;
 
-        using var client = new HttpClient();
         var content = new StringContent(JsonSerializer.Serialize(jsonPayload), Encoding.UTF8, "application/json");
         content.Headers.ContentType = new MediaTypeHeaderValue("application/json");
 
+        // ✅ gebruik gedeelde client
```
J005t67: Remove Dutch comment

SourceCode/GPS/Classes/AgShareUploader.cs
```diff
@@ -60,9 +60,10 @@ public static object BuildFieldUploadJsonWithConversion(
             isPublic = false,
             originLat = boundary.Average(p => p[0]),
             originLon = boundary.Average(p => p[1]),
-            boundary,
+            boundary = boundary,
```
J005t67, line 76: I don't understand the intention of this assignment. Please add comments or correct, whatever is apropriate

SourceCode/GPS/Classes/AgShareUploader.cs
```diff
@@ -37,33 +41,45 @@ public static object BuildFieldUploadJsonWithConversion(
...
+        // Return the final field object including the boundary and AB lines
         return new
         {
             name = fieldName,
             isPublic = false,
             originLat = boundary.Average(p => p[0]),
             originLon = boundary.Average(p => p[1]),
             boundary = boundary,
-            abLines
+            abLines = abLines // bevat nu name, type, coords[]
```
J005t67: translate comment

SourceCode/AgOpenGPS.Core/AgShare/AgShareClient.cs
```diff
@@ -0,0 +1,64 @@
...
+                Debug.WriteLine("API key test exception: " + ex.Message);
+                return false;
+            }
+        }
+
+        public async Task<bool> UploadIsoXmlFieldAsync(string fieldId, object jsonPayload)
+        {
+            if (string.IsNullOrWhiteSpace(ApiKey)) return false;
+
+            // ✅ gebruik gedeelde client
```
J005t67, line 43: I see a funny checkmark character here. I woud prefer to remove it, because for me it only adds complexity. But please leave it in if you think I'm exaggerating.

J005t67: Do not forget to translate the comment line

SourceCode/GPS/Forms/AgShareSettingsForm.cs
```diff
@@ -2,42 +2,46 @@
...
+        bool result = await _agShareClient.TestApiKeyAsync();
         lblStatus.Text = result ? "✔ API key valid" : "❌ API key Not Valid";
         lblStatus.ForeColor = result ? System.Drawing.Color.Green : System.Drawing.Color.Red;
     }
 
     private void btnSave_Click(object sender, EventArgs e)
     {
-        AgShareApi.SaveApiKey(txtApiKey.Text);
+        SaveApiKey(txtApiKey.Text);
         MessageBox.Show("API key opgeslagen.", "Opgeslagen", MessageBoxButtons.OK, MessageBoxIcon.Information);
```
J005t67: Translate msg

SourceCode/GPS/Forms/SaveOpen.Designer.cs
```diff
@@ -54,7 +55,7 @@ public void ExportFieldAs_ISOXMLv3()
             }
         }
 
-        public async void ExportFieldAs_ISOXMLv4()
+        public void ExportFieldAs_ISOXMLv4()
```
J005t67, line 58: Good to see that you could avoid the 'async' and 'await'. This really improves maintanability.

SourceCode/GPS/Forms/SaveOpen.Designer.cs
```diff
@@ -72,25 +73,31 @@ public async void ExportFieldAs_ISOXMLv4()
...
             catch (Exception e)
             {
                 Log.EventWriter("Export Field as ISOXML: " + e.Message);
             }
+        }
+
+        public void UploadFieldToAgShare()
+        {
+            if (Settings.Default.AgShareApiKey == "apikey")
+                return;
```
J005t67, line 86: Many coding standards forbid one-line if statements, but allow one of these two options:
1) add {} around return;
2) put return on same line as condition

SourceCode/GPS/AgOpenGPS.csproj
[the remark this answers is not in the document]
FantasyTeddy: > * this is not a small change, that should be part of a large PR. If we want this, we should commit and test it separately.

Unfortunately, it **_is_** a small change. There is nothing that **needs** to be changed if the `LangVersion` is increased. And because C# is fully backwards compatible, there is virtually zero risk of breaking existing code.
**However**, the risk is with any future change that the new `LangVersion` enables. And that's why we need to be careful if we really want to make this change.

> * I would prefer a explicit choice over a default that might change when we don't notice it or don't expect it.

And that is exactly what the default (not specifying `LangVersion`, **not** specifying `default`) does: it uses a well defined [default language version](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-versioning#defaults) (C# 7.3 for .NET Framework).

> * I think a read something last week that Avalonia needs 8 or higher, (or maybe it was a more general advise, to switch to a higher language version before porting to Avalonia.)

Since the Avalonia project will be its own assembly anyway, there is no need to make the current `AgOpenGPS` project follow any recommendation for Avalonia (because it will stay a WinForms project).

> * I'm currently working on more viewmodels (about 16, one for each configurationpage) that I like to create lazy, (only create them when the user switches to the corresponding page). In this case, the conditional assignment operator '??=' would help a lot to reduce code size. ??= is only available in language version 8 and higher.

There are **a lot** of cool new C# features in the newer versions, but I think it would be best to just _wait_ until we are fully ready to enjoy these new features in the new projects.

> * If I remember correctly, language version 9 also supports autogenerated properties that will 'help' to make the Properties code more compact. To me this is very scary, and no thelpfull at all. When debugging databindings, I often start with setting a breakpoint in the getter of a property. This is not possible when the property code is autogenerated. (And I do not want a language version that allows other people to write cod ethat is hard to maintain)

New language features always come with a learning curve. Sure, they are not all "better" than what we already have, but calling it "hard to maintain" is quite a bold statement...

>   Conclusion: my preference for now is to switch to language version 8 explicitly in a separate PR.

I strongly advise against changing the `LangVersion` at all. But if you are aware of all the risks and accept them, I won't object any further.

---

> Just ran into this warning:
> [...]

That is exactly the reason, why I'm currently against changing the `LangVersion`...

SourceCode/GPS/Forms/AgShareSettingsForm.cs
FantasyTeddy: Already done.

SourceCode/GPS/Forms/AgShareSettingsForm.cs
FantasyTeddy: Already done.

SourceCode/GPS/Forms/AgShareSettingsForm.cs
FantasyTeddy: Auto-generated.

SourceCode/GPS/Classes/AgShareUploader.cs
FantasyTeddy: > Although LinQ expressions can be very powerfull and helpfull, not so many people can read them easily. Please be a bit empathic to your co-developers and try to rewrite this with easy readably code.

I don't agree with that statement. Yes, LINQ can be harder to understand if you are coming from a purely imperative (not functional) programming background. However, I think that understanding the concept of functional programming (and therefore LINQ) is very powerful and enables for some very concise and less error-prone code.

> Another problem with chained Linq expressions is that, the compiler might decide to optimize and do things in a different order than you expect. This can make it hard to debug.

Really? Never heard of that before. Can you give some source for that?

> Also if one is trying to figure out what is going on, and using the debugger to do so, this can result in a disapointment because of the previous remark. So in my opinion, this code is pretty hard to maintain, (close to unmaintainable) and shoud be simplified by using easier constructs like a for loop.

Visual Studio is very good in providing a good debugging experience with LINQ statements. It is true that chained LINQ statements are a little bit harder to follow, but one-line LINQ statements certainly are not.

SourceCode/GPS/AgOpenGPS.csproj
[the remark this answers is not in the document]
richardklasens: Use 8.0 instead of latest?

SourceCode/GPS/Forms/SaveOpen.Designer.cs
FantasyTeddy: How is the second option better? That is also a one-line if-statement (even one line less), when it is on the same line as the condition.

I think we need to setup some kind of coding guideline for this project, if we really want to enforce all those (rather opinionated) formatting rules. Otherwise it is very unfair for new people to just "know" what they have to conform to.
Or even better: setup code analysis that does all these checks automatically and produces build warnings if they are not followed.

SourceCode/GPS/AgOpenGPS.csproj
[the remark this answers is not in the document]
lansalot: On the subject of upgrades, installations and unintended side effects.....

https://github.com/AvaloniaUI/Avalonia/discussions/14277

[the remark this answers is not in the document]
FantasyTeddy: Again: I would not change `LangVerison` at all. Yes, it is nice to be able to use all these new C# features, but for me it isn't worth the risk this introduces. (And then we have something to look forward when advancing the project 😉)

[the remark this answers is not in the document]
richardklasens: Then I don’t understand the “backwards compatibility” you mentioned

SourceCode/GPS/Classes/AgShareApi.cs
richardklasens: Yes, but since we don’t have a working Backend + frontend yet on the other end of the API, It has to be this for now.

SourceCode/GPS/Properties/Settings.cs
richardklasens: Will fix

SourceCode/GPS/Properties/Settings.cs
richardklasens: Default is false.. means every field is not published globally in the backend of the API.

SourceCode/GPS/AgOpenGPS.csproj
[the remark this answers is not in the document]
FantasyTeddy: Backwards compatibility in this instance means that all old code will run virtually identically in a newer language version.
- old code + newer `LangVersion` ✅ 

However, any new code that will be developed using the newer language version falls under the risks mentioned above.
- new code + old runtime ❌

SourceCode/GPS/Classes/AgShareApi.cs
FantasyTeddy: Which part of my comment are you referring to? The configurable hostname and port, or the HTTPS part?

richardklasens: I don’t have to be configurable?

SourceCode/GPS/Properties/Settings.cs
richardklasens: Yes you can on the other side on the front end or later in AgOpenGPS

SourceCode/GPS/Classes/AgShareUploader.cs
richardklasens: So we leave it?

SourceCode/GPS/Forms/SaveOpen.Designer.cs
richardklasens: Same here, leave it?

SourceCode/GPS/Classes/AgShareApi.cs
FantasyTeddy: Okay, I agree. I was just thinking that it wouldn't hurt to implement this already, but it can also be done later.

SourceCode/GPS/Forms/SaveOpen.Designer.cs
J005t67: With code like this:

if (someCondition)
    doSomeThing;

The typical mistake will be that somebody changes it to:

if (someCondition)
   doSomeThing;
   doSomeAdditionalThing;

J005t67: With code like this (sorry, do not know how to layout this properly, so read the ..... as spaces)

if (someCondition)
....doSomething;

There is some chance that people will change it to something like:

if (someCondition)
....doSomething;
....doSomethingMore;

or

if (someCondition)
....doSomethingMore;
....doSomething

And assume that both doSomething; and doSomethingMore; are both executed conditionally.

J005t67: Just Googled it. Another mistake could happen when you temporarily want to comment out the line, like this

if (someCondition)
//....doSomething;

SourceCode/GPS/Classes/AgShareUploader.cs
J005t67: My earlier comments, were are already outdated, when I typed them The code has already been improved. Both regarding the ternary operator and the too complex LINQ expression

SourceCode/GPS/Forms/SaveOpen.Designer.cs
FantasyTeddy: I am aware of these pitfalls and I won't deny them.
However, I'm arguing that your second option suffers from very similar issues:

```csharp
if (someCondition) doSomething;
```
Someone could easily try to add another statement like this:

```csharp
if (someCondition) doSomething; doSomethingMore;
```
Or try to temporarily remove the executing code like this:

```csharp
if (someCondition) //doSomething;
```
The only "safe" way of doing it would be your option one; to add curly braces `{}` around the statement.

And/or we introduce automated format linting, which would flag all (or at least most) of the dangerous code variants mentioned.
(By the way, feedback on #759 would be appreciated.)

PR #749, opened 2025-04-18
fix youturn with offset

P00qwerty:


FantasyTeddy:
@richardklasens: Do we want to merge this into `master`?

richardklasens:
It’s ok to do that. If this is the only real glitch it’s good to update release

FantasyTeddy:
Okay. Maybe we should add a `hotfix/6.7` branch and use that to fix bugs like these (and maybe #735 & #742 ?). What do you think?

lansalot:
Definitely, a branch to merge all fixes into before going to main

richardklasens:
Changed base branch and merged

Issue #722, opened 2025-03-30
Make boundary section detection lazier


gunicsba:
Issue: Our current settings is that as soon as the outer section touches the boundary it turns section off.
Which is what we expect but in reality this means driving on the boundary curve  with a real machine it turns the section on and off.

Not sure what would be the right solution but in reality I don't think anyone has under 25 cm sections so maybe that 10-20 centimeter of overshoot on the boundary would be acceptable?

For now we just force the sections on when driving the boundary curve.

I only moved the boundary curve 2 centimeter and the outer section is off:
![Image](https://github.com/user-attachments/assets/02c410ea-5381-47a3-af93-f7a78b24a092)

Coverage was set to 50%

richardklasens:
If I’m right, Turn on the SectionFullyOutsideBnd button to turn off when the section is fully outside. It’s in the section width config next to the coverage setting.

gunicsba:
Yes Richard you're right but we do want it to be off when it's really outside. but when it's just barely outside it should stay on.

exentler:
I had similar issue when I was spreding fertiliser with my Bogballe M35W.
I used manual on the first outer round

P-R9:
Hello, 

why not make it as the most commertial Section Control Terminales? I also saw this issue by using Section Control and the border. So maybe the option I show here could be a solution.
In my opinion, the key word would be overlap tolerances in the settings of Sectioncontrol, similar to the example on the Amatron 4 below. 
 I also faced this problem because if driving with a Sprayer at the Headland Tramline you get the Problem that the last section deactivates if you have Boundary overlap at off and drive with 100% overlap in the case of herbicides. So in my opinion driver shouldn't press any workswitch at headland turning for that I have section control but as it is you don't drive perfect in the headland tramline and if you drive a bit far out or in curves which is normal in the field like gunicsbas exemple shows it would be nice if the system would think for the driver in this case. 

Best option would be a tolerance setting for the last sections in % or cm.
And another good change would be if you also could set the border overlap in % so it could be more individual, if you want it to be on set it to 100% and if off set it to 0% pretty easy to understand for everybody and it would be the same as for the normal overlap.

Exemples Amazone Amatron 4 overlap settings:

https://www.manualslib.com/manual/1618417/Amazone-Amatron-4.html?page=79#manual
https://www.manualslib.com/manual/1618417/Amazone-Amatron-4.html?page=80#manual

gunicsba:
[...]

PR #750, opened 2025-04-18
Added a helper function for boundary detection

gunicsba:
Lazier boundary detection

### DIFF
--- a/SourceCode/GPS/Forms/OpenGL.Designer.cs
+++ b/SourceCode/GPS/Forms/OpenGL.Designer.cs
@@ -859,8 +859,9 @@ namespace AgOpenGPS
                 for (int j = 0; j < tool.numOfSections; j++)
                 {
                     //only one first left point, the rest are all rights moved over to left
-                    isLeftIn = j == 0 ? bnd.IsPointInsideFenceArea(section[j].leftPoint) : isRightIn;
-                    isRightIn = bnd.IsPointInsideFenceArea(section[j].rightPoint);
+                    double boundaryCoverage = (100.0 - tool.minCoverage) / 100;
+                    isLeftIn = bnd.IsPointInsideFenceArea(section[j].leftPoint.ShrinkToward(section[j].rightPoint, boundaryCoverage));
+                    isRightIn = bnd.IsPointInsideFenceArea(section[j].rightPoint.ShrinkToward(section[j].leftPoint, boundaryCoverage));
 
                     if (!tool.isSectionOffWhenOut)
                     {
--- a/SourceCode/GPS/Classes/vec3.cs
+++ b/SourceCode/GPS/Classes/vec3.cs
@@ -128,6 +128,11 @@ namespace AgOpenGPS
         {
             return new vec2(lhs.easting + rhs.easting, lhs.northing + rhs.northing);
         }
+
+        public vec2 ShrinkToward(vec2 other, double factor)
+        {
+            return this + ((other - this) * factor);
+        }
     }
 
 }


gunicsba:
As described in https://github.com/AgOpenGPS-Official/AgOpenGPS/issues/722

SourceCode/GPS/Forms/OpenGL.Designer.cs
```diff
@@ -859,8 +859,9 @@ private void oglBack_Paint(object sender, PaintEventArgs e)
                 for (int j = 0; j < tool.numOfSections; j++)
                 {
                     //only one first left point, the rest are all rights moved over to left
-                    isLeftIn = j == 0 ? bnd.IsPointInsideFenceArea(section[j].leftPoint) : isRightIn;
-                    isRightIn = bnd.IsPointInsideFenceArea(section[j].rightPoint);
+                    double boundaryCoverage = (100.0 - tool.minCoverage) / 100;
+                    isLeftIn = bnd.IsPointInsideFenceArea(section[j].leftPoint.ShrinkToward(section[j].rightPoint, boundaryCoverage));
+                    isRightIn = bnd.IsPointInsideFenceArea(section[j].rightPoint.ShrinkToward(section[j].leftPoint, boundaryCoverage));
```
mattmccormack, line 864: If `tool.minCoverage` is less than 50%, then the two ends of each section are "shrunk" past each other, so left becomes right, and right becomes left. 10% gives the same outcome as 90%.

In the case of this screenshot, it says left edge of first section is inside the boundary and right edge of first section is outside:
image

mattmccormack: Should ShrinkToward() be applied only to the left point of the first section, and the right point of the last section? Not left and right of each individual section.

SourceCode/GPS/Forms/OpenGL.Designer.cs
```diff
@@ -859,8 +859,9 @@ private void oglBack_Paint(object sender, PaintEventArgs e)
                 for (int j = 0; j < tool.numOfSections; j++)
                 {
                     //only one first left point, the rest are all rights moved over to left
-                    isLeftIn = j == 0 ? bnd.IsPointInsideFenceArea(section[j].leftPoint) : isRightIn;
```
mattmccormack, line 862: The ternary operator ensures that the left edge is checked only for the first section (`j == 0`). For subsequent sections, the result from the previous right edge check is reused (because `section[n].leftPoint == section[n-1].rightPoint`), eliminating redundant checks of each point.

gunicsba: Well yeah but...
currently it works like this:
0aaaaaa1bbbbbb2cccccc3
So when we first run the cycle we grab 0 then the 2nd cycle will grab 1 and so on. All is good.
But now we introduce something like:
0aAaaAa1bBbbBb2cCccCc3
so in our calculation we'll use the first A instead of 0 and the 2nd A instead of 1. But the next section should use B instead of the 2nd A. Hence the reason of me saying no to the cache.

SourceCode/GPS/Forms/OpenGL.Designer.cs
gunicsba: In this case yes we should shrink towards right. But if the boundary is on the right side we should be shrinking towards the left, no?
And we don't really know where the boundary is :( 

As for tool.minCoverage that's definitely a bug I didn't spot. Should we be:
Using 1/2 of it? (since we substract it twice)
Or should we add another variable dedicated to the Allowed Boundary overlap?

SourceCode/GPS/Forms/OpenGL.Designer.cs
mattmccormack: You're right. I was only thinking about the outer-most sections.

SourceCode/GPS/Forms/OpenGL.Designer.cs
```diff
@@ -859,8 +859,9 @@ private void oglBack_Paint(object sender, PaintEventArgs e)
...
-                    isRightIn = bnd.IsPointInsideFenceArea(section[j].rightPoint);
+                    double boundaryCoverage = (100.0 - tool.minCoverage) / 100;
+                    isLeftIn = bnd.IsPointInsideFenceArea(section[j].leftPoint.ShrinkToward(section[j].rightPoint, boundaryCoverage));
+                    isRightIn = bnd.IsPointInsideFenceArea(section[j].rightPoint.ShrinkToward(section[j].leftPoint, boundaryCoverage));
 
                     if (!tool.isSectionOffWhenOut)
```
mattmccormack, line 866: Should the behaviour change depending on `tool.isSectionOffWhenOut`? I think if that setting is `false` then you don't want the *shrink* to turn off a section too early/too aggressively if you want 100% application within the boundary and don't care about boundary overlap. What do you think?

```C#
if (tool.isSectionOffWhenOut)
{
    isLeftIn = bnd.IsPointInsideFenceArea(section[j].leftPoint.ShrinkToward(section[j].rightPoint, boundaryCoverage));
    isRightIn = bnd.IsPointInsideFenceArea(section[j].rightPoint.ShrinkToward(section[j].leftPoint, boundaryCoverage));

    //both points must be in boundary
    section[j].isInBoundary = isLeftIn && isRightIn ;
}
```
(checks points B and C shrunk inwards from A and D)
Slide1

```C#
else
{
    //re-use right point as left for second and subsequent sections
    isLeftIn = (j==0) ? bnd.IsPointInsideFenceArea(section[j].leftPoint) : isRightIn;
    isRightIn = bnd.IsPointInsideFenceArea(section[j].rightPoint);

    //at least one point must be in boundary
    section[j].isInBoundary = isLeftIn || isRightIn ;
}
```
(checks points A and D as before)
Slide2

SourceCode/GPS/Forms/OpenGL.Designer.cs
mattmccormack: I'm not sure reusing `tool.minCoverage` for boundary overlap is the best approach since it's meant for field coverage. A separate setting like `tool.boundaryTolerance` might be clearer. For example, if the setting is at 100% to cover all ground, does it make sense to turn off a section if 99% is within the boundary? That feels counterintuitive. 

If we stick with `tool.minCoverage`, it should probably be capped between 100% - 50% to avoid the above issue. If you halve it, then 50% meaning 25% wouldn't be intuitive. In fact, is there a real need for more than 50% overlap? If the operator wants full overlap, wouldn't it be better to turn off `tool.isSectionOffWhenOut`?

SourceCode/GPS/Forms/OpenGL.Designer.cs
mattmccormack: I've been thinking some more. What about only applying the "shrink" when one side is outside the boundary?

For example, if the right edge is outside, and the left edge is inside, check if the right edge shrunk by % is inside, and if so, keep the section on.

```C#
for (int j = 0; j < tool.numOfSections; j++)
{
    //re-use right point as left for second and subsequent sections
    isLeftIn = (j==0) ? bnd.IsPointInsideFenceArea(section[j].leftPoint) : isRightIn;
    isRightIn = bnd.IsPointInsideFenceArea(section[j].rightPoint);

    if (tool.isSectionOffWhenOut)
    {  
        //off if both edges are outside
        if (!isLeftIn && !isRightIn)
        {
            section[j].isInBoundary = false;
        }
        else
        {
            //only check the shrunk point if the outer edge isn't within the boundary
            isLeftInShrunk = isLeftIn || bnd.IsPointInsideFenceArea(section[j].leftPoint.ShrinkToward(section[j].rightPoint, boundaryCoverage));
            isRightInShrunk = isRightIn || bnd.IsPointInsideFenceArea(section[j].rightPoint.ShrinkToward(section[j].leftPoint, boundaryCoverage));

            section[j].isInBoundary = isLeftInShrunk && isRightInShrunk;
        }
    }
    else
    {
        //at least one point must be in boundary
        section[j].isInBoundary = isLeftIn || isRightIn ;
    }
}
```

SourceCode/GPS/Forms/OpenGL.Designer.cs
gunicsba: Realistically for planters its about 30cm of overlap. For normal sprayers maybe 30-50cm would be enough.

Someone suggested the use of cm, maybe we should go with that?

SourceCode/GPS/Forms/OpenGL.Designer.cs
gunicsba: Yes this should be perfect I believe.

SourceCode/GPS/Forms/OpenGL.Designer.cs
mattmccormack: cm would be ok, as long as setting > section width is handled nicely

richardklasens:
Can we fix this an approve / merge / close the PR?

SourceCode/GPS/Forms/OpenGL.Designer.cs
gunicsba: Sorry I didn't notice your question in time.

I've done a few interviews and we believe the tolerance in cm would work the best.

Is your code available somewhere?

As for the value:
Since most sprayers are 50cm between nozzles, some machines are 25cm but that's very rare. 
So I think we could easily allow 20cm of Boundary Overlap but having a value would be the best.

And we'd shrink the sections virtually by that amount.

mattmccormack: Hey @gunicsba, cm sounds good to me. Will you change your shrink towards method to take a distance value now? My test code is just what you see above - not worth pushing.

richardklasens:
Since it isnt actively ontopic here, i close the PR for now

PR #754, opened 2025-04-19
Fixed Tool Width not updating in the config form when loading a new v…

### LINKED ISSUE #753: Configuration menu does not update tool width when loading a new vehicle
When opening the configuration menu the currently loaded vehicle's tool width is correctly displayed, however if you open any other vehicle from the list the width in both the sidebar and the summary at the bottom do not update until you click the green check and re-open the configuration menu.

This is due to the code running these two lines when the configuration form loads

```
lblVehicleToolWidth.Text = Convert.ToString((int)(mf.tool.width * 100 * mf.cm2CmOrIn));
SectionFeetInchesTotalWidthLabelUpdate();
```

However the OnClick action for the "Open" button doesn't set the lblVehicleToolWidth.Text property from the just-loaded vehicle before calling the update function, causing it to not change from the vehicle that was loaded when the form originally opens.

A PR to solve this is coming momentarily.

spencerryan:
Hey All,

First PR against this project. This fixes the tool width not updating correctly when loading a new vehicle. This looks like it has already been fixed in develop by the new 'Profile' system, but I didn't know if you accepted PRs against master for any releases that may happen between 6.7.0 and develop.

Feel free to reject this if that's the right thing to do. 

Fixes #753

### DIFF
--- a/SourceCode/GPS/Forms/Settings/ConfigVehicle.Designer.cs
+++ b/SourceCode/GPS/Forms/Settings/ConfigVehicle.Designer.cs
@@ -54,6 +54,7 @@ namespace AgOpenGPS
                         //reset AOG
                         mf.LoadSettings();
 
+                        lblVehicleToolWidth.Text = Convert.ToString((int)(mf.tool.width * 100 * mf.cm2CmOrIn));
                         SectionFeetInchesTotalWidthLabelUpdate();
 
                         SendSettings();


mattmccormack:
Hi @spencerryan, thanks for the PR. Can you please tell me which version you experience this problem with? It seems ok on the latest release `6.7.0`.

`SectionFeetInchesTotalWidthLabelUpdate();` updates the tool width text on the vehicle open/save tab after the open button is pressed. 

`lblVehicleToolWidth` is on the sections page (circled below), and is updated when you navigate to that tab in the settings.

image

spencerryan:
It is broken in 6.7.0.

There are more details in the bug, but its a domino effect. The code doesn't update lblVehicleToolWidth.Text which causes lblSummaryWidth.Text to never change, but only when a new vehicle is loaded after the configuration form is open.


To Reproduce:

- Build at least two vehicles with different tool widths and number of sections
- Ensure one of them as active/loaded and close the configuration form
- Open the configuration form and note that the Width in the list (Units, width, sections, etc) is correct and the width at the bottom is correct next to the okay checkmark/close button
- Click the other vehicle in the list and click Open and say yes to confirm
- Note that all of the data about the vehicle/tool has changed except width, which remains the previous vehicle/tool
- Close/Reopen the Configuration Dialog and the width is now correct


Quite confusing if you're switching vehicles and you're not sure why the width isn't correct.

mattmccormack:
This is the behaviour I get with the latest 6.7.0 release.
TractorA is 700cm 7 sections
TractorB is 450cm 3 sections

[video.webm](https://github.com/user-attachments/assets/5b641914-6cc9-4f47-aaca-00310d8a2de6)

spencerryan:
That's not the behavior I've seen in any version of 6, including 6.7.0 (either prebuilt, building it myself, or building master)

You can see in my video that when the vehicle is switched from 7320-Corn to 7320-Test the number of sections correctly goes from 1 to 3, but the width doesn't update (it should go to 45')

I didn't dig extensively through the code but there shouldn't be a difference here between metric and imperial, the fact is that btnVehicleLoad_Click() doesn't update lblVehicleToolWidth.Text before calling SectionFeetInchesTotalWidthLabelUpdate();


https://github.com/user-attachments/assets/8a8b9e70-b6e5-4ca2-93f3-a0dd419477ea

mattmccormack:
I changed both of my vehicles to imperial and can now exactly replicate the bug you describe. 

When set to imperial, `SectionFeetInchesTotalWidthLabelUpdate()` is dependent on the value of `lblVehicleToolWidth.Text`. This is not the case when set to metric. 

I think this should be merged into `hotfix/6.7` rather than `master`.

spencerryan:
Ah yes you're right. I vaguely recall that as I was stepping through it last night but focused on the Imperial stuff since it's what we use.

mattmccormack:
Ideally the code would be refactored so that it's not dependent on text displayed on another tab. But I think this PR is sufficient to pull into the hotfix branch for 6.7 given the bug is masked by new Profile implementation.

spencerryan:
That's a similar thought path I went down last night, I even started looking at building it to do the math/text the way the metric part of the function worked, but once I realized that develop has this fixed in a different way I figured the one liner was more straight forward since that's already exactly how the OnLoad function does it.

PR #755, opened 2025-04-19
Restores the original cross track background texture PNG

### LINKED ISSUE #743: Cross track error background texture changed
The cross track display used to look like this: 
Image

now has a white bar behind it and it's darker:
Image

This was introduced in this commit: https://github.com/AgOpenGPS-Official/AgOpenGPS/commit/425d8e765f40a2d9c94920a31704bdff18ea8ffd

Cross Track Background used to use `Resources.z_crossTrackBkgnd`
Image

Now it uses `Resources.ZoomOut48` (the white bar is the minus sign stretched out).
Image

@J005t67 can we reintroduce the original texture?

mattmccormack:
I've re-added the resource that used to be called `z_crossTrackBkgnd.png`. Using that in place of `ZoomOut48.png` behind the cross track error display.

Fixes #743 

Would it be better to just draw a rectangle in OGL rather than using a texture for this?


SourceCode/GPS/Classes/ScreenTextures.cs
```diff
@@ -44,7 +44,7 @@ public Texture2D CrossTrackBackground
         {
             get
             {
-                if (_crossTrackBackGround == null) _crossTrackBackGround = new Texture2D(Resources.ZoomOut48);
+                if (_crossTrackBackGround == null) _crossTrackBackGround = new Texture2D(Resources.CrossTrackBackground);
```
J005t67, line 47: I don't know exactly what happened, but it seems that I introduced a copy/paste error a few weeks ago. Thank you for fixing it! I really like it when experienced users, use recent development versions and help to keep it stable and reliable. Thanks again!

J005t67 (approved):
I like the idea of replacing the texture with some drawing code, but let's do that after this PR has been completed.

richardklasens:
Do we want this in fixes and then to develop?

mattmccormack:
> Do we want this in fixes and then to develop?

Hey Richard, this bug didn't make it into 6.7.0 release, so I don't think you need to put this into a fixes branch.

PR #770, opened 2025-04-28
Don't double-zip the build output

mattmccormack:
When I download a build artifact (e.g. one built as part of a pull request Action) I get `AgOpenGPS.zip` within `AgOpenGPS.zip`. The `release` Action doesn't suffer from this problem.

`actions/upload-artifact@v4` apparently creates a .zip of whatever files you give it.


FantasyTeddy (approved):
Nice catch, thank you!

PR #771, opened 2025-05-03
Remove dependencies from Field and Fields and FieldDescription to streamers

J005t67:
Same content as other recent pull request. The other one has so many merge related changes that I decided to make this new PR.

In order to keep the Models small and reusable, I removed the dependencies from the models to the streamers.
The streamers know how to stream the models, but the models do not know anything about streaming anymore.


richardklasens (approved):

richardklasens:
@J005t67 we have some conflicts now

PR #780, opened 2025-05-26
Fix for Issue 779

gunicsba:
Multiline coordinates are loaded properly

### DIFF
--- a/SourceCode/GPS/Forms/Field/FormFieldKML.cs
+++ b/SourceCode/GPS/Forms/Field/FormFieldKML.cs
@@ -149,12 +149,12 @@ namespace AgOpenGPS
                                 if (endIndex == -1)
                                 {
                                     //just add the line
-                                    if (startIndex == -1) coordinates += line.Substring(0);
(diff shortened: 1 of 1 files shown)


richardklasens (commented):
Does this also work on other types other then Valtra xml?

richardklasens:
Since i got no response, i close the PR for now.. (can be reopend if the last answer is solved

gunicsba:
@richardklasens  if you check the code there was a block to handle multiline coordinates. But for some reason we only support coordinates separated by a space character.
could we reconsider this?
Today the almighty chatgpt at first also generated KML with line breaks...

richardklasens:
I asked you for the file from Valtra and you’ve send me an ISOXML which imported fine.. but I get the point now it’s about a KML..

richardklasens (approved):

gunicsba:
> I asked you for the file from Valtra and you’ve send me an ISOXML which imported fine.. but I get the point now it’s about a KML..

I think you've specifically asked for ISOXML files in the original ticket we had the KML-s

PR #784, opened 2025-06-27
A Full new approach for AgShare Implementation. 

richardklasens:
I will squash merge


SourceCode/GPS/Classes/AgShare/AgShareClient.cs
```diff
@@ -0,0 +1,130 @@
...
+            baseUrl = url.TrimEnd('/');
+            BuildClient();
+        }
+
+        // Checks if the API key and connection are valid
+        public async Task<(bool ok, string message)> CheckApiAsync()
+        {
+            try
+            {
+                using (var tempClient = new HttpClient())
```
FantasyTeddy, line 58: Why create a new `HttpClient` if we already have one in the `client` field?

SourceCode/GPS/Classes/AgShare/AgShareClient.cs
```diff
@@ -0,0 +1,130 @@
...
+        {
+            baseUrl = serverUrl.TrimEnd('/');
+            apiKey = key;
+            BuildClient();
+        }
+
+        // Rebuilds the HttpClient with updated headers
+        private void BuildClient()
+        {
+            client = new HttpClient();
```
FantasyTeddy, line 33: We could also set the `baseUrl` to the `BaseAddress` property of the `HttpClient`, then we would not need to add `baseUrl` to every single request.

SourceCode/GPS/Classes/AgShare/AgShareClient.cs
```diff
@@ -0,0 +1,130 @@
...
+                }
+            }
+            catch (Exception ex)
+            {
+                return (false, $"Error: {ex.Message}");
+            }
+        }
+
+        // Uploads a field by ID with JSON payload
+        public async Task<(bool ok, string message)> UploadFieldAsync(Guid fieldId, object fieldPayload)
```
FantasyTeddy, line 82: Could we add some kind of `FieldUploadDto` to get some type-safety for the `fieldPayload`?

SourceCode/GPS/Classes/AgShare/AgShareClient.cs
```diff
@@ -0,0 +1,130 @@
...
+            string json = await response.Content.ReadAsStringAsync();
+            return JsonConvert.DeserializeObject<List<AgShareGetOwnFieldDto>>(json);
+        }
+
+        // Downloads a specific field as raw JSON string
+        public async Task<string> DownloadFieldAsync(Guid fieldId)
+        {
+            var response = await client.GetAsync($"{baseUrl}/api/fields/{fieldId}");
+            response.EnsureSuccessStatusCode();
+            return await response.Content.ReadAsStringAsync();
```
FantasyTeddy, line 117: Could we do the deserialization here as well? (Just like in `GetOwnFieldsAsync`)

SourceCode/GPS/Classes/vec3.cs
```diff
@@ -3,8 +3,6 @@
 using AgOpenGPS.Core.Models;
 using System;
 
-namespace AgOpenGPS
```
FantasyTeddy, line 6: Was this intentional, to remove this namespace?

SourceCode/GPS/Classes/AgShare/AgShareDTO.cs
FantasyTeddy, line 1: I think we should clearly distinguish between objects that represent the actual objects from AgShare (like `AgShareGetOwnFieldDto` or `AgShareFieldDto`) and objects that are an internal representation used to pass data around (like `FieldSnapshot`).
Maybe we could at least put them into different files?

SourceCode/GPS/Classes/AgShare/AgShareDTO.cs
```diff
@@ -0,0 +1,64 @@
...
+        public string Type { get; set; }
+        public List<CoordinateDto> Coords { get; set; }
+    }
+
+    public class AgShareGetOwnFieldDto
+    {
+        public Guid Id { get; set; }
+        public string Name { get; set; }
+        public List<CoordinateDto> OuterBoundary { get; set; }
+        public double AreaHa => GeoUtils.CalculateAreaInHa(OuterBoundary);
```
FantasyTeddy, line 57: Is this calculated property used somewhere?

SourceCode/GPS/Classes/AgShare/AgShareDownloader.cs
```diff
@@ -0,0 +1,244 @@
...
+using AgOpenGPS.Properties;
+using AgOpenGPS.Core.Models;
+using AgOpenGPS.Classes.AgShare.Helpers;
+
+namespace AgOpenGPS
+{
+    /// <summary>
+    /// Central helper class for downloading, parsing and saving AgShare fields locally.
+    /// </summary>
+    public class CAgShareDownloader
```
FantasyTeddy, line 17: Please do not use the `C` prefix for classes. [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation) is considered an anti-pattern in C#.

SourceCode/GPS/Classes/AgShare/AgShareDownloader.cs
```diff
@@ -0,0 +1,244 @@
...
+            try
+            {
+                string json = await client.DownloadFieldAsync(fieldId);
+                var dto = JsonConvert.DeserializeObject<AgShareFieldDto>(json);
+                var model = AgShareFieldParser.Parse(dto);
+                string fieldDir = Path.Combine(RegistrySettings.fieldsDirectory, model.Name);
+                FieldFileWriter.WriteAllFiles(model, fieldDir);
+                return true;
+            }
+            catch (Exception ex)
```
FantasyTeddy, line 39: Catching generic `Exception` is considered bad practice, because it can hide bugs that should be prevented earlier.

SourceCode/GPS/Classes/AgShare/AgShareDownloader.cs
```diff
@@ -0,0 +1,244 @@
...
+            return (downloaded, skipped);
+        }
+
+
+    }
+
+    /// <summary>
+    /// Utility class that writes a LocalFieldModel to standard AgOpenGPS-compatible files.
+    /// </summary>
+    public static class FieldFileWriter
```
FantasyTeddy, line 107: Can we put this class into its own file?

SourceCode/GPS/Classes/AgShare/AgShareUploader.cs
```diff
@@ -0,0 +1,202 @@
...
+using AgOpenGPS.Core.Models;
+
+namespace AgOpenGPS
+{
+    public class CAgShareUploader
+    {
+        private readonly FormGPS gps;
+
+        // Create a snapshot from the current GPS session to upload
+        public static FieldSnapshot CreateSnapshot(FormGPS gps)
```
FantasyTeddy, line 22: `FormGPS` should not be passed to this lower level class. Can we do this in a way that does not require that?

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
```diff
@@ -0,0 +1,82 @@
+﻿using AgOpenGPS.Core.Models;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace AgOpenGPS.Classes.AgShare.Helpers
+{
+    public static class AgShareFieldParser
```
FantasyTeddy, line 8: If a file contains a single class, it should be named after the class.

FantasyTeddy (approved):

PR #785, opened 2025-06-29
Fix issues with Resources due to the different versions of System.Memory Package.

richardklasens:


FantasyTeddy:
Can you quickly explain what the problem is that this fixes?

FantasyTeddy (approved):

PR #788, opened 2025-06-30
New IsoXML Methode, refactored and cleaned up

richardklasens:


J005t67:
I really like it that you split up the original file. Good software is all about 'Separation of concerns'.
But,..... in general, you can benifit from the split even more when you reduce or eliminate the dependencies to the main form or the application model.

SourceCode/GPS/Protocols/ISOBUS/IsoXmlParserHelpers.cs
```diff
@@ -0,0 +1,133 @@
...
+            }
+
+            mf.trk.gArr.Add(track);
+        }
+
+        private static vec2 ParseVec2(XmlNode pnt, ApplicationModel appModel)
+        {
+            double.TryParse(pnt.Attributes["C"]?.Value, NumberStyles.Float, CultureInfo.InvariantCulture, out double lat);
+            double.TryParse(pnt.Attributes["D"]?.Value, NumberStyles.Float, CultureInfo.InvariantCulture, out double lon);
+            GeoCoord geo = appModel.LocalPlane.ConvertWgs84ToGeoCoord(new Wgs84(lat, lon));
```
J005t67, line 280: I don't think we need this function.  You can leave it to the caller to convert the GeoCoord to a vec2

SourceCode/GPS/Protocols/ISOBUS/IsoXmlParserHelpers.cs
```diff
@@ -0,0 +1,133 @@
...
+            return new vec2(geo);
+        }
+
+        private static GeoCoord ParseGeoCoord(XmlNode pnt, ApplicationModel appModel)
+        {
+            double.TryParse(pnt.Attributes["C"]?.Value, NumberStyles.Float, CultureInfo.InvariantCulture, out double lat);
+            double.TryParse(pnt.Attributes["D"]?.Value, NumberStyles.Float, CultureInfo.InvariantCulture, out double lon);
```
J005t67, line 287: It seems that the function that we need most in this class is missing: ParseWgs84 or TryParseWgs84. 
The following code fragment is duplicated about 5 times in the code:

 double.TryParse(pnt.Attributes["C"]?.Value, NumberStyles.Float, CultureInfo.InvariantCulture, out double lat);
 double.TryParse(pnt.Attributes["D"]?.Value, NumberStyles.Float, CultureInfo.InvariantCulture, out double lon);
          ..........(new Wgs84(lat, lon));

I would prefer to replace all occurences by one new function (ParseWgs84 or TryParseWgs84)

The nice thing is, it would not need the ApplicatonModel or LocalPlane

SourceCode/GPS/Forms/Field/FormFieldISOXML.cs
```diff
@@ -1,11 +1,13 @@
-﻿using AgLibrary.Logging;
+// Updated version of FormFieldISOXML with logic delegated to IsoXmlFieldBuilder
```
FantasyTeddy: I don't think this comment adds any value. This is only relevant now, but in the future it won't be.

SourceCode/GPS/Forms/Field/FormFieldISOXML.cs
```diff
@@ -16,140 +18,103 @@ namespace AgOpenGPS
...
                 int index = 0;
+
                 try
                 {
-                    //scan thru all the fields
                     foreach (XmlNode nodePFD in pfd)
                     {
-                        double.TryParse(nodePFD.Attributes["D"].Value, NumberStyles.Float, CultureInfo.InvariantCulture, out double area);
+                        double area;
+                        double.TryParse(nodePFD.Attributes["D"].Value, NumberStyles.Float, CultureInfo.InvariantCulture, out area);
```
FantasyTeddy: Why not use the shorter version from before?

SourceCode/GPS/Forms/Field/FormFieldISOXML.cs
```diff
@@ -16,140 +18,103 @@ namespace AgOpenGPS
...
                         area *= 0.0001;
 
-                        // PFD - A=ID, C=FieldName, D = Area in sq m
                         tree.Nodes.Add(nodePFD.Attributes["C"].Value + " Area: " + area + " Ha  " + nodePFD.Attributes["A"].Value);
                         tree.Nodes[tree.Nodes.Count - 1].Tag = index++;
 
-                        //nodes in current Partial Field like PLN, GGP, LSG etc
                         XmlNodeList fieldParts = nodePFD.ChildNodes;
+                        TreeNode fieldNode = tree.Nodes[tree.Nodes.Count - 1]; // laatste veld-node
```
FantasyTeddy: Dutch comment :)

SourceCode/GPS/Forms/Field/FormFieldISOXML.cs
```diff
@@ -16,140 +18,103 @@ namespace AgOpenGPS
...
 
-                        //do the boundary first - find all the polygons (PLN)
+                        // GGP-based guidance
                         foreach (XmlNode nodePart in fieldParts)
                         {
-                            // PLN/LSG/PNT Count the points
-                            //grab the polygons
-                            if (nodePart.Name == "PLN")
```
FantasyTeddy, line 98: Why do you not need to parse the "PLN" nodes anymore?

SourceCode/GPS/Forms/Field/FormFieldISOXML.cs
```diff
@@ -16,140 +18,103 @@ namespace AgOpenGPS
...
                         foreach (XmlNode nodePart in fieldParts)
                         {
-                            if (nodePart.Name == "GGP" && nodePart.ChildNodes[0].Attributes.GetNamedItem("B") != null && nodePart.ChildNodes[0].Attributes.GetNamedItem("C") != null)
+                            if (nodePart.Name == "LSG" && nodePart.Attributes["A"] != null && nodePart.Attributes["A"].Value == "5")
                             {
-                                //in GPN "B" is the name and "C" is the type
-                                if (nodePart.ChildNodes[0].Attributes["C"].Value == "1")
-                                    tree.Nodes[tree.Nodes.Count - 1].Nodes.Add("ABLine: " + nodePart.ChildNodes[0].Attributes["B"].Value);
-                                else if (nodePart.ChildNodes[0].Attributes["C"].Value == "2")
-                                    tree.Nodes[tree.Nodes.Count - 1].Nodes.Add("A+: " + nodePart.ChildNodes[0].Attributes["B"].Value);
```
FantasyTeddy, line 131: Did you remove all other types except "AB-Line" and "Curve" because they are not relevant?

SourceCode/GPS/Forms/Field/FormFieldISOXML.cs
```diff
@@ -16,140 +18,103 @@ namespace AgOpenGPS
...
-                                }
-                                else if (nodePart.Attributes["A"].Value == "10")
-                                {
-                                    bndType = "Headland:";
+                                    string name = gpn.Attributes["B"].Value;
+                                    string type = gpn.Attributes["C"].Value;
+
+                                    if (type == "1") // AB-line
+                                    {
+                                        TreeNode abNode = new TreeNode("AB-" + name);
```
FantasyTeddy: Is it a good idea to display the nodes as "AB-..." instead of "AB: ..." ?

SourceCode/GPS/Forms/Field/FormFieldISOXML.cs
```diff
@@ -177,642 +146,73 @@ private void FormFieldISOXML_Load(object sender, EventArgs e)
...
+            else
+            {
+                btnBuildFields.Enabled = false;
+                btnAddDate.Enabled = false;
+                btnAddTime.Enabled = false;
+                tboxFieldName.Enabled = false;
+            }
         }
 
         private async void btnBuildFields_Click(object sender, EventArgs e)
```
FantasyTeddy: Because you removed the asynchronous operation (`await mf.FileSaveEverythingBeforeClosingField()`) you can remove the `async` here.

SourceCode/GPS/Protocols/ISOBUS/IsoXmlFieldBuilder.cs
```diff
@@ -0,0 +1,213 @@
...
+using AgOpenGPS.Core;
+
+namespace AgOpenGPS.Protocols.ISOBUS
+{
+    // Handles all logic for building a field from ISOXML data
+    public class IsoXmlFieldBuilder
+    {
+        private readonly XmlNodeList _fieldParts;
+        private readonly string _fieldName;
+        private readonly FormGPS _mf;
```
FantasyTeddy: It is rather unfortunate, that we need to have `FormGPS` in this class... but it will be quite difficult to remove, so this is just a remark.

SourceCode/GPS/Protocols/ISOBUS/IsoXmlFieldBuilder.cs
```diff
@@ -0,0 +1,213 @@
...
+            if (_coordCount == 0) return false;
+
+            double avgLat = _latSum / _coordCount;
+            double avgLon = _lonSum / _coordCount;
+            origin = new Wgs84(avgLat, avgLon);
+            return true;
+        }
+
+        // Build the outer boundary and inner boundaries
+        public void TryBuildBoundaries()
```
FantasyTeddy: I would remove the `Try` prefix for this method. This should be used for methods like `TryExtractOrigin` above, where you return `bool` and assign an `out` parameter.

SourceCode/GPS/Protocols/ISOBUS/IsoXmlParserHelpers.cs
```diff
@@ -0,0 +1,145 @@
...
+using System.Collections.Generic;
+using System.Globalization;
+using System.Xml;
+using AgOpenGPS.Core;
+using AgOpenGPS.Core.Models;
+
+namespace AgOpenGPS.Protocols.ISOBUS
+{
+    // Static helpers to parse GGP and LSG guidance lines
+    public static class IsoXmlParserHelpers
```
FantasyTeddy, line 14: Do you have some kind of strategy, which parts of the parsing goes into `IsoXmlFieldBuilder` and which parts into `IsoXmlParserHelpers`?

SourceCode/GPS/Protocols/ISOBUS/IsoXmlFieldBuilder.cs
```diff
@@ -0,0 +1,213 @@
...
+                }
+                else if (node.Name == "LSG" && node.Attributes["A"]?.Value == "5")
+                {
+                    IsoXmlParserHelpers.ParseLSGNode(node, _appModel, _mf);
+                }
+            }
+        }
+
+        // Save field files after job is started
+        public void SaveFieldFiles(string directoryPath)
```
FantasyTeddy: I think the two methods `SaveFieldFiles` and `FinalizeField` are kind of out of place... maybe they should stay in `FormFieldISOXML`?

SourceCode/GPS/Forms/Field/FormFieldISOXML.cs
```diff
@@ -16,140 +18,103 @@ namespace AgOpenGPS
...
-        private double lonK, latK;
-
         private XmlDocument iso;
-
         private string xmlFilename;
         private XmlNodeList pfd;
-
         private int idxFieldSelected;
 
         public FormFieldISOXML(Form _callingForm)
```
J005t67: This file has a nasty dependency to the main form of type FormGPS. The code in this file will only work if a FormGPS is passed to the constructor here, so there is no use trying to hide the dependnecy by specifying an argument of type Form.

SourceCode/GPS/Forms/Field/FormFieldISOXML.cs
```diff
@@ -16,140 +18,103 @@ namespace AgOpenGPS
 {
     public partial class FormFieldISOXML : Form
```
J005t67: If we apply Pascal-casing strictly, the correct class and file name would be FormFieldIsoxml or FormFieldIsoXml.
See for example the class names XmlNode and XmlDocument

SourceCode/GPS/Forms/Field/FormFieldISOXML.cs
```diff
@@ -177,642 +146,73 @@ private void FormFieldISOXML_Load(object sender, EventArgs e)
...
-                tboxFieldName.Enabled = false;
-            }
-            else
+            if (idxFieldSelected >= 0)
             {
+                labelField.Text = idxFieldSelected + " " + pfd[idxFieldSelected].Attributes["C"].Value;
+                tboxFieldName.Text = pfd[idxFieldSelected].Attributes["C"].Value;
                 btnBuildFields.Enabled = true;
                 btnAddDate.Enabled = true;
                 btnAddTime.Enabled = true;
```
J005t67: I would eliminate the else brach (and part of the if branch), like this:
bool enabled = (idxFieldSelected >= 0);
if (enabled)
{
    // set labelField and tboxFieldName
}
btnBuildField.Enabled = enabled;
// three more buttons

SourceCode/GPS/Forms/Field/FormFieldISOXML.cs
J005t67: The use of async here scares me a bit. Is this really something we would like to do in the background? FormFieldIsoXml has access to FormGPS (and it does poke around in the instance variables of FormGPS).  Seems scary to me if this code sets for examle a new field origin while other code is still using another field with another origin.

SourceCode/GPS/Forms/Field/FormFieldISOXML.cs
```diff
@@ -821,7 +221,7 @@ private async void btnBuildFields_Click(object sender, EventArgs e)
 
         private void tboxFieldName_TextChanged(object sender, EventArgs e)
         {
-            System.Windows.Forms.TextBox textboxSender = (System.Windows.Forms.TextBox)sender;
+            TextBox textboxSender = (TextBox)sender;
```
J005t67: Small typo: textBoxSender needs capital B.

SourceCode/GPS/Protocols/ISOBUS/IsoXmlFieldBuilder.cs
```diff
@@ -0,0 +1,213 @@
...
+            foreach (XmlNode pnt in lsg.SelectNodes("PNT"))
+            {
+                if (double.TryParse(pnt.Attributes["C"]?.Value, NumberStyles.Float, CultureInfo.InvariantCulture, out double lat) &&
+                    double.TryParse(pnt.Attributes["D"]?.Value, NumberStyles.Float, CultureInfo.InvariantCulture, out double lon))
+                {
+                    GeoCoord geo = _appModel.LocalPlane.ConvertWgs84ToGeoCoord(new Wgs84(lat, lon));
+                    list.fenceLine.Add(new vec3(geo));
+                }
+            }
+            list.CalculateFenceArea(_mf.bnd.bndList.Count);
```
J005t67: Are you sure about this? Do we need something (possible unrelated?) from _mf to calculate something for a local var list?

Related to my question:
The whole code would be much more simple if the import code could simply return a complete field, in stead of using a dependency to the FormGPS to set a field-aspects like boundary there. Unfortunately, I did not finish the implementation of my new Field class yet, so the field aspects are still part of FormGPS. Maybe we should not spend to much time on improving this import code now. It will be a lot more easy and effective if the Field implementation is ready. (Same for the MVVM pattern.)

SourceCode/GPS/Protocols/ISOBUS/IsoXmlFieldBuilder.cs
richardklasens: Good point — agreed that having FormGPS in here couples logic with UI too tightly. Refactoring this would require deeper separation of track/curve processing logic, so for now it's left in place, but noted for future cleanup.

SourceCode/GPS/Protocols/ISOBUS/IsoXmlFieldBuilder.cs
richardklasens: removed for all functions

SourceCode/GPS/Protocols/ISOBUS/IsoXmlParserHelpers.cs
richardklasens: Good question! The current division is pragmatic but loosely based on responsibilities:

IsoXmlParserHelpers contains static, format-specific utilities for parsing individual XML nodes (GGP, LSG, PLN, etc.) into domain objects (CTrk, vec3, etc.). These are stateless and reusable.

IsoXmlFieldBuilder orchestrates the field-level building process, using the helpers to extract relevant content and inject it into the main appmodel (FormGPS, bndList, etc.). It maintains context (e.g. _mf) and handles higher-level logic like deciding what constitutes an outer boundary, headland, or guidance section.

That said, I agree it would be beneficial to further formalize this separation — possibly by extracting logic from FormGPS into service-style components to decouple data parsing from UI.

SourceCode/GPS/Protocols/ISOBUS/IsoXmlFieldBuilder.cs
```diff
@@ -0,0 +1,213 @@
...
+    {
+        private readonly XmlNodeList _fieldParts;
+        private readonly string _fieldName;
+        private readonly FormGPS _mf;
+        private readonly ApplicationModel _appModel;
+
+        private double _latSum = 0, _lonSum = 0;
+        private int _coordCount = 0;
+
+        public IsoXmlFieldBuilder(XmlNodeList fieldParts, string fieldName, ApplicationModel appModel, FormGPS mf)
```
J005t67: I would prefer the name IsoXmlFieldImporter

SourceCode/GPS/Protocols/ISOBUS/IsoXmlFieldBuilder.cs
richardklasens: Yes, you're right... I moved them to the Form Class

SourceCode/GPS/Forms/Field/FormFieldISOXML.cs
richardklasens: Done

SourceCode/GPS/Forms/Field/FormFieldISOXML.cs
richardklasens: Good point, you're absolutely right regarding PascalCase. I'll rename FormFieldISOXML to FormFieldIsoXml to stay consistent with BCL-style class names like XmlDocument and XmlNode.

SourceCode/GPS/Forms/Field/FormFieldISOXML.cs
richardklasens: Solved, it isnt needed anymore

SourceCode/GPS/Protocols/ISOBUS/IsoXmlFieldBuilder.cs
richardklasens: You're absolutely right, relying on _mf.bnd.bndList.Count inside ParseBoundaryFromLSG() created an unnecessary dependency on the main form.

I've refactored the method to make it pure and self-contained:
The caller now passes the boundary index (int index) explicitly, which is used for CalculateFenceArea() and FixFenceLine().

This removes the coupling with _mf, improves testability, and makes the method easier to reuse elsewhere. Thanks for pointing it out!

SourceCode/GPS/Protocols/ISOBUS/IsoXmlFieldBuilder.cs
richardklasens: Done

SourceCode/GPS/Forms/Field/FormFieldISOXML.cs
richardklasens: Solved

SourceCode/GPS/Classes/CBoundaryList.cs
```diff
@@ -17,5 +17,11 @@ public CBoundaryList()
             area = 0;
             isDriveThru = false;
         }
+
+        public void CalculateFenceArea()
+        {
+
+        }
```
FantasyTeddy: I think this got added accidentally.

SourceCode/GPS/Forms/Field/FormFieldISOXML.cs
```diff
@@ -14,817 +15,257 @@
...
-                                double.TryParse(pnt.Attributes["D"].Value, NumberStyles.Float, CultureInfo.InvariantCulture, out lonK);
-
-                                lat += latK;
-                                lon += lonK;
-                                counter += 1;
-                            }
-                        }
+                        // Fallback to estimate if D is missing or 0
+                        area = EstimateAreaFromPln(nodePFD, mf.AppModel) * 0.0001;
                     }
```
FantasyTeddy: I think something went wrong with the brackets here: the `if` statement has the following body:
```csharp
area *= 0.0001;
```
... and not this part inside the `{` `}`.

SourceCode/GPS/Forms/Field/FormFieldISOXML.cs
```diff
@@ -14,817 +15,257 @@
...
-                                if (nodePart.ChildNodes[0].Attributes["C"].Value == "1")
-                                    tree.Nodes[tree.Nodes.Count - 1].Nodes.Add("ABLine: " + nodePart.ChildNodes[0].Attributes["B"].Value);
-                                else if (nodePart.ChildNodes[0].Attributes["C"].Value == "2")
-                                    tree.Nodes[tree.Nodes.Count - 1].Nodes.Add("A+: " + nodePart.ChildNodes[0].Attributes["B"].Value);
-                                else if (nodePart.ChildNodes[0].Attributes["C"].Value == "3")
-                                    tree.Nodes[tree.Nodes.Count - 1].Nodes.Add("Curve: " + nodePart.ChildNodes[0].Attributes["B"].Value);
-                                else if (nodePart.ChildNodes[0].Attributes["C"].Value == "4")
-                                    tree.Nodes[tree.Nodes.Count - 1].Nodes.Add("Pivot: " + nodePart.ChildNodes[0].Attributes["B"].Value);
-                                else if (nodePart.ChildNodes[0].Attributes["C"].Value == "5")
-                                    tree.Nodes[tree.Nodes.Count - 1].Nodes.Add("Spiral: " + nodePart.ChildNodes[0].Attributes["B"].Value);
```
FantasyTeddy, line 137: Maybe re-add these types as well?

SourceCode/GPS/Protocols/ISOBUS/IsoXmlFieldImporter.cs
```diff
@@ -0,0 +1,186 @@
...
+using System.IO;
+using System.Xml;
+
+namespace AgOpenGPS.Protocols.ISOBUS
+{
+    // Handles all logic for building a field from ISOXML data
+    public class IsoXmlFieldImporter
+    {
+        private readonly XmlNodeList _fieldParts;
+        private readonly string _fieldName;
```
FantasyTeddy: The field `_fieldName` is never used.

SourceCode/GPS/Protocols/ISOBUS/IsoXmlParserHelpers.cs
FantasyTeddy: > IsoXmlParserHelpers contains static, format-specific utilities for parsing individual XML nodes (GGP, LSG, PLN, etc.) into domain objects (CTrk, vec3, etc.). These are stateless and reusable.

Depending on how you define "stateless", I don't think this is correct. The methods that take `FormGPS` as an argument are modifying it, which is not stateless I would argue.

SourceCode/GPS/Protocols/ISOBUS/IsoXmlFieldImporter.cs
```diff
@@ -0,0 +1,186 @@
...
+using AgOpenGPS.Core.Models;
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+using System.Xml;
+
+namespace AgOpenGPS.Protocols.ISOBUS
+{
+    // Handles all logic for building a field from ISOXML data
```
FantasyTeddy: I think these comments are quite useful. Maybe consider making them [XML doc comments](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/), since they can be useful for the one using it. (Also applies to the methods with comments)

SourceCode/GPS/Forms/Field/FormFieldISOXML.cs
```diff
@@ -14,817 +16,291 @@
...
-                                double.TryParse(pnt.Attributes["D"].Value, NumberStyles.Float, CultureInfo.InvariantCulture, out lonK);
-
-                                lat += latK;
-                                lon += lonK;
-                                counter += 1;
-                            }
-                        }
+                        // Fallback to estimate if D is missing or invalid
+                        area = EstimateAreaFromPln(nodePFD, mf.AppModel) * 0.0001;
                     }
```
FantasyTeddy: If the area needs to be converted into hectares in both cases, maybe do that outside the `if`-`else` statement.

SourceCode/GPS/Protocols/ISOBUS/CurveCABTools.cs
```diff
@@ -0,0 +1,121 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace AgOpenGPS
+{
+    // Stateless curve processing helpers matching CABCurve behavior
+    public static class CurveCABTools
+    {
+        // Perform minimum spacing, interpolation and heading calculation
+        public static void Preprocess(ref List<vec3> points, double minSpacing, double interpolationSpacing)
```
FantasyTeddy: Why use `ref` here? (And everywhere else in this class?)

SourceCode/GPS/Protocols/ISOBUS/CurveCABTools.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            MakePointMinimumSpacing(ref points, minSpacing);
+            InterpolatePoints(ref points, interpolationSpacing);
+            CalculateHeadings(ref points);
+        }
+
+        // Remove points that are too close to each other
+        public static void MakePointMinimumSpacing(ref List<vec3> points, double minSpacing)
```
FantasyTeddy: Is this the same method that already exists in `CABCurve`? If so, I would try to replace that method there.

SourceCode/GPS/Protocols/ISOBUS/CurveCABTools.cs
```diff
@@ -0,0 +1,121 @@
...
+            if (glm.DistanceSquared(final, compare) > 0.00001)
+            {
+                spaced.Add(final);
+            }
+
+            points = spaced;
+        }
+
+        // Add interpolated points between existing ones
+        public static void InterpolatePoints(ref List<vec3> points, double spacingMeters)
```
FantasyTeddy:
```suggestion
        private static void InterpolatePoints(ref List<vec3> points, double spacingMeters)
```

SourceCode/GPS/Protocols/ISOBUS/CurveCABTools.cs
FantasyTeddy, line 1: I'm not sure this file is at the correct location. Since it seems to be related to `CABCurve` rather than ISO-XML, I would put it closer to that class.

SourceCode/GPS/Protocols/ISOBUS/CurveCABTools.cs
```diff
@@ -0,0 +1,121 @@
...
+                    result.Add(new vec3(x, y, 0));
+                }
+            }
+
+            result.Add(points[points.Count - 1]);
+            points = result;
+        }
+
+        // Compute segment-based headings (AgOpenGPS: heading = atan2(dx, dy))
+        public static void CalculateHeadings(ref List<vec3> points)
```
FantasyTeddy: Same question here: is that the same method as in `CABCurve`?

SourceCode/GPS/Protocols/ISOBUS/IsoXmlFieldImporter.cs
FantasyTeddy: Why did you remove the comments?

SourceCode/GPS/Protocols/ISOBUS/IsoXmlParserHelpers.cs
```diff
@@ -0,0 +1,261 @@
...
+                origin = new Wgs84();
+                return false;
+            }
+
+            origin = new Wgs84(latSum / count, lonSum / count);
+            return true;
+        }
+
+        // Parse PLN boundaries into CBoundaryList objects
+        public static List<CBoundaryList> ParseBoundaries(XmlNodeList fieldParts, ApplicationModel appModel)
```
FantasyTeddy, line 53: This method is not used. Did you intend to use it in `IsoXmlFieldImporter.GetBoundaries()`?

SourceCode/GPS/Protocols/ISOBUS/IsoXmlParserHelpers.cs
```diff
@@ -0,0 +1,261 @@
...
+                    GeoCoord geo = appModel.LocalPlane.ConvertWgs84ToGeoCoord(new Wgs84(lat, lon));
+                    list.fenceLine.Add(new vec3(geo));
+                }
+            }
+
+            return list;
+        }
+
+        // Parse GGP → GPN → LSG line
+        public static CTrk ParseGGPNode(XmlNode node, ApplicationModel appModel)
```
FantasyTeddy:
```suggestion
        private static CTrk ParseGGPNode(XmlNode node, ApplicationModel appModel)
```

SourceCode/GPS/Protocols/ISOBUS/IsoXmlParserHelpers.cs
```diff
@@ -0,0 +1,261 @@
...
+            string name = gpn.Attributes["B"]?.Value ?? node.Attributes["B"]?.Value ?? "Unnamed";
+
+            if (lineType == "1") return ParseABLine(lsg, name, appModel);
+            else if (lineType == "3") return ParseCurveLine(lsg, name, appModel);
+
+            return null;
+        }
+
+        // Parse LSG line directly
+        public static CTrk ParseLSGNode(XmlNode lsg, ApplicationModel appModel)
```
FantasyTeddy:
```suggestion
        private static CTrk ParseLSGNode(XmlNode lsg, ApplicationModel appModel)
```

SourceCode/GPS/Protocols/ISOBUS/CurveCABTools.cs
richardklasens: I will leave this class as is, when CABCurve an CABLine are decoupled from mf, this class can be deleted later. 
I add a comment in the class about this

SourceCode/GPS/Protocols/ISOBUS/CurveCABTools.cs
richardklasens: See below

SourceCode/GPS/Protocols/ISOBUS/CurveCABTools.cs
richardklasens: Removed ref

SourceCode/GPS/Protocols/ISOBUS/IsoXmlParserHelpers.cs
richardklasens: i forgot to change it in the importer class to use the helper. Changed and cleaned up

SourceCode/GPS/Protocols/ISOBUS/CurveCABTools.cs
richardklasens: Since i got some crazy erros which i cant solve immediatly, i put the ref back and leave it for now. It does not harm any

FantasyTeddy (approved):

PR #792, opened 2025-07-09
Improves application shutdown and data handling

richardklasens:
Enhances the application shutdown process by:

- Providing visual feedback during shutdown.
- Performing field data saving in a background task to improve performance and prevent UI freezes.
- Enhancing AgShare upload handling during shutdown to avoid potential issues.

-  Solves a problem that fields weren't stored locally when AgShare was active

### DIFF
--- a/SourceCode/GPS/Forms/Field/FormFieldDir.cs
+++ b/SourceCode/GPS/Forms/Field/FormFieldDir.cs
@@ -77,7 +77,7 @@ namespace AgOpenGPS
                 return;
             }
 
-            if (mf.isJobStarted) mf.FileSaveEverythingBeforeClosingField();
+            if (mf.isJobStarted) _ = mf.FileSaveEverythingBeforeClosingField();
 
             //append date time to name
 
--- a/SourceCode/GPS/Forms/FormGPS.cs
+++ b/SourceCode/GPS/Forms/FormGPS.cs
@@ -1,4 +1,4 @@
-//Please, if you use this, share the improvements
+﻿//Please, if you use this, share the improvements
 
 using AgLibrary.Logging;
 using AgOpenGPS;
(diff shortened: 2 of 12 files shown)


SourceCode/GPS/Forms/Field/FormFieldDir.cs
```diff
@@ -77,7 +77,7 @@ private void btnSave_Click(object sender, EventArgs e)
                 return;
             }
 
-            if (mf.isJobStarted) mf.FileSaveEverythingBeforeClosingField();
+            if (mf.isJobStarted) _ = mf.FileSaveEverythingBeforeClosingField();
```
FantasyTeddy, line 80: Is this really what you want in those cases?
You are starting the task for saving everything, but you are not waiting for it to finish. Is it really okay to do that?

SourceCode/GPS/Forms/FormGPS.cs
```diff
@@ -584,39 +583,87 @@ private async void FormGPS_FormClosing(object sender, FormClosingEventArgs e)
...
-                    btnSectionMasterManual.PerformClick();
+        private async Task ShowSavingFormAndShutdown(int choice)
+        {
+            using (FormSaving savingForm = new FormSaving())
+            {
+                savingForm.InitializeSteps(isJobStarted);
+                savingForm.Show();
 
-                if (Settings.Default.AgShareEnabled)
+                await Task.Delay(300); // Let UI settle
```
FantasyTeddy, line 599: There are a lot of delays in this code. Is this really what you want? You are slowing down the entire shutdown process that way...

SourceCode/GPS/Forms/FormGPS.cs
```diff
@@ -584,39 +583,87 @@ private async void FormGPS_FormClosing(object sender, FormClosingEventArgs e)
...
+
+                if (isJobStarted)
                 {
-                    TimedMessageBox(5000, "AgShare", "Uploading field to AgShare...\nPlease wait and get a beer.");
-                    isAgShareUploadStarted = true;
-                    agShareUploadTask = CAgShareUploader.UploadAsync(snapshot, agShareClient, this);
+                    // STEP 0: Parameters
+                    await Task.Delay(300);
+                    savingForm.UpdateStep(0, ShutdownSteps.ParamsDone);
```
FantasyTeddy, line 605: It would be nice if the responsibility for choosing the correct steps would be clearer.
At the moment `FormSaving` has a partial responsibility for adding the initial steps, but this method (`ShowSavingFormAndShutdown`) does all the updating.

Additionally, it is unfortunate that you need to know two different things to update a step: you need to know the index of the step in the list **and** the appropriate text for that step.
Having a method where you can just say "Set step `Params` to `Done`" would be way nicer.

SourceCode/GPS/Forms/FormGPS.cs
```diff
@@ -584,39 +583,87 @@ private async void FormGPS_FormClosing(object sender, FormClosingEventArgs e)
...
 
-                    e.Cancel = true;
-                    await DelayedShutdownAfterUpload(choice);
-                    return;
+                    // STEP 2: Save Field
+                    savingForm.UpdateStep(fieldSaveIndex, ShutdownSteps.SaveField);
+                    await FileSaveEverythingBeforeClosingField();
+                    savingForm.UpdateStep(fieldSaveIndex, ShutdownSteps.FieldSaved);
+
+                    // STEP 3: Settings
```
FantasyTeddy, line 637: Is this step doing something?

SourceCode/GPS/Forms/FormSaving.cs
```diff
@@ -0,0 +1,73 @@
...
+            else
+            {
+                lstSteps.Items.Add(ShutdownSteps.SaveSettings);
+                lstSteps.Items.Add(ShutdownSteps.Finalizing);
+            }
+        }
+
+        public void UpdateStep(int index, string text)
+        {
+            if (index >= 0 && index < lstSteps.Items.Count)
```
FantasyTeddy, line 40: I think this range check is not what you want in this case.
The `index` is always predetermined and if it is not in the valid range, it should crash the application because it is a programming error.

SourceCode/GPS/Forms/Field/FormFieldDir.cs
richardklasens: In this case there wasn’t any step needed to await the task, instead of rewrite it here, I picked the Fire&Forget method here.

SourceCode/GPS/Forms/FormGPS.cs
richardklasens: It does not really matter… the saving goes a lot faster then the GUI can show.. since the program is closing/computer shutting down, it’s ok ( I think)

SourceCode/GPS/Forms/FormGPS.cs
richardklasens: Not sure what I should change then?

SourceCode/GPS/Forms/FormGPS.cs
richardklasens: Not yet, the settings.save is happening later

SourceCode/GPS/Forms/FormSaving.cs
richardklasens: Because we change the index due to AgShare enabled or not

SourceCode/GPS/Forms/Field/FormFieldDir.cs
FantasyTeddy: Okay and that is true for all other calls to `FileSaveEverythingBeforeClosingField`?

SourceCode/GPS/Forms/FormGPS.cs
FantasyTeddy: Okay, so you are basically just doing it to make it "look nice"?

SourceCode/GPS/Forms/FormGPS.cs
FantasyTeddy: I will try to make a proposal how that could look like.

SourceCode/GPS/Forms/FormSaving.cs
FantasyTeddy: I know, but even with that it would be better to crash the program and tell the developer: "hey you messed up your logic" than just silently do nothing (which may result in hard to reproduce bug reports).

SourceCode/GPS/Forms/Field/FormFieldDir.cs
richardklasens: Yes

SourceCode/GPS/Forms/FormGPS.cs
richardklasens: Exactly

SourceCode/GPS/Forms/FormSaving.cs
richardklasens: I will change in the next commit.. good one

SourceCode/GPS/Forms/Field/FormFieldDir.cs
FantasyTeddy: Then I would seriously consider splitting the `FileSaveEverythingBeforeClosingField` method into a "StartSaving" and "WaitForSaved" method.
Fire and forget should only be used for exceptional cases and this is no exception then.

PR #793, opened 2025-07-10
user-selectable precision, optional simulate data

SK21:
Users can select less than 7 digits of precision for location for devices that require it. Simulate data is optional. If no AGIO data is available and 'Simulate' is checked, the simulated data will be sent. If 'Simulate' is not checked and AGIO data is unavailable,  GPS_Out won't send any data.


richardklasens (commented):
Hi David. Thanks for this.
Is is Field tested @SK21 ?

SK21:
No it hasn't been field tested.

PR #800, opened 2025-07-15
Replaces message boxes with custom dialogs

richardklasens:
This change replaces the standard `MessageBox` with a custom `FormDialog` throughout the application.

This provides a consistent look and feel and allows for more control over dialog appearance and behavior.

- Improves user experience by using custom dialogs.
- Simplifies code by centralizing dialog creation and management.

### DIFF
--- a/SourceCode/GPS/Forms/FormDialog.Designer.cs
+++ b/SourceCode/GPS/Forms/FormDialog.Designer.cs
@@ -1,0 +1,144 @@
+﻿namespace AgOpenGPS.Forms
+{
+    partial class FormDialog
+    {
+        /// <summary>
+        /// Required designer variable.
+        /// </summary>
+        private System.ComponentModel.IContainer components = null;
+
+        /// <summary>
+        /// Clean up any resources being used.
+        /// </summary>
+        /// <param name="disposing">true if managed resources should be disposed; otherwise, false.</param>
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing && (components != null))
+            {
+                components.Dispose();
+            }
+            base.Dispose(disposing);
+        }
+
+        #region Windows Form Designer generated code
+
+        /// <summary>
+        /// Required method for Designer support - do not modify
+        /// the contents of this method with the code editor.
+        /// </summary>
+        private void InitializeComponent()
+        {
+            this.lblMessage = new System.Windows.Forms.Label();
+            this.btnOK = new System.Windows.Forms.Button();
+            this.btnCancel = new System.Windows.Forms.Button();
+            this.lblTitle = new System.Windows.Forms.Label();
+            this.panel1 = new System.Windows.Forms.Panel();
+            this.panel1.SuspendLayout();
+            this.SuspendLayout();
+            // 
+            // lblMessage
+            // 
+            this.lblMessage.Anchor = ((System.Windows.Forms.AnchorStyles)((((System.Windows.Forms.AnchorStyles.Top | System.Windows.Forms.AnchorStyles.Bottom) 
+            | System.Windows.Forms.AnchorStyles.Left) 
+            | System.Windows.Forms.AnchorStyles.Right)));
+            this.lblMessage.BackColor = System.Drawing.Color.FromArgb(((int)(((byte)(255)))), ((int)(((byte)(192)))), ((int)(((byte)(128)))));
+            this.lblMessage.Font = new System.Drawing.Font("Tahoma", 21.75F, System.Drawing.FontStyle.Bold, System.Drawing.GraphicsUnit.Point, ((byte)(0)));
+            this.lblMessage.Location = new System.Drawing.Point(0, 53);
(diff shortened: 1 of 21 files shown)


SourceCode/GPS/Forms/FormDialog.Designer.cs
FantasyTeddy, line 1: Here are a few general remarks regarding the design of the dialog:

image

---
- Use different formatting for the title and the message. Maybe make the message smaller and not bold?
- I'm not sure if having the two buttons so far apart makes sense. What if we put both buttons in the lower right corner?

SourceCode/GPS/Forms/FormDialog.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows.Forms;
+using System.Windows.Forms.DataVisualization.Charting;
+
+namespace AgOpenGPS.Forms
+{
+    public partial class FormDialog : Form
+    {
+        public FormDialog(string message, string title = "Title", bool showCancel = true)
```
FantasyTeddy: Because `FormDialog` is always created in the `Show()` method, you can make this constructor `private`. That way it will always be used like this as intended.

FantasyTeddy: If you give `title` a default value (`"Title"`), you are implying that the title does not need to be set. However, I think both `message` and `title` are mandatory, aren't they?

SourceCode/GPS/Forms/FormDialog.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            InitializeComponent();
+            lblMessage.Text = message;
+            lblTitle.Text = title;
+            btnCancel.Visible = showCancel;
+        }
+
+
+        public static DialogResult Show(string title, string message, MessageBoxButtons buttons = MessageBoxButtons.OKCancel)
```
FantasyTeddy, line 24: Are you sure that you want to reuse the WinForms `MessageBoxButtons` and `DialogResult` here? I think it would be nice to define custom enums for these values, because you don't support all the different variants anyway.

SourceCode/GPS/Forms/FormDialog.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            this.DialogResult = DialogResult.OK;
+            this.Close();
+        }
+
+        private void btnCancel_Click(object sender, EventArgs e)
+        {
+            this.DialogResult = DialogResult.Cancel;
+            this.Close();
+        }
```
FantasyTeddy: You have set the `DialogResult` property on both buttons in the designer so I don't think it is necessary to do this manually (neither set `DialogResult` nor call `Close()`).

SourceCode/GPS/Forms/Pickers/FormFilePicker.cs
FantasyTeddy, line 1: This is quite a big change in itself. Could we maybe do that in a separate PR?

SourceCode/GPS/Forms/Pickers/FormFilePicker.cs
```diff
@@ -36,479 +34,282 @@ private void FormFilePicker_Load(object sender, EventArgs e)
         {
             order = 0;
             timer1.Enabled = true;
-            ListViewItem itm;
 
-            string[] dirs = Directory.GetDirectories(RegistrySettings.fieldsDirectory);
+            LoadFieldList();
```
FantasyTeddy, line 38: The `LoadFieldList()` method can return early if there is an error. Maybe `UpdateListView()` should be skipped in that case?

SourceCode/GPS/Forms/Pickers/FormFilePicker.cs
FantasyTeddy: I reviewed this refactoring now and I think this is awesome :) Great work!

SourceCode/GPS/Forms/FormDialog.cs
richardklasens: That's not in this scope for now. I will leave it with the Forms DialogResult  for now. We only use Ok and Cancel.

SourceCode/GPS/Forms/FormDialog.cs
```diff
@@ -0,0 +1,42 @@
...
+        public static DialogResult Show(string title, string message, MessageBoxButtons buttons = MessageBoxButtons.OKCancel)
+        {
+            bool showCancel = (buttons == MessageBoxButtons.OKCancel || buttons == MessageBoxButtons.YesNo || buttons == MessageBoxButtons.YesNoCancel);
+            return new FormDialog(message, title, showCancel).ShowDialog();
+        }
+
+
+        private void btnOK_Click(object sender, EventArgs e)
+        {
+            // Not used, but here for future compatibility
```
FantasyTeddy, line 33: What do you have in mind regarding "future compatibility"?

FantasyTeddy (approved):

PR #801, opened 2025-07-19
Update NMEA parsing heading and roll for SkyTraq receivers

THiamu:
The "PSTI, 032 and 035" messages are completely outdated! I'm using this code with "PSTI, 036" messages for 3 years now inside old version 5.5., working good. Would like to see getting it in the latest program to make cheap SkyTraq PX1172RH receivers usable for the community.

I was kind of responsable for the old code about 5 years ago. See also these posts:
https://discourse.agopengps.com/t/skytraq-dual-gnss-board/19614
https://discourse.agopengps.com/t/dual-gps-using-skytraq-receivers/1577

### DIFF
--- a/SourceCode/AgIO/Source/Forms/NMEA.Designer.cs
+++ b/SourceCode/AgIO/Source/Forms/NMEA.Designer.cs
@@ -209,9 +209,9 @@ namespace AgIO
                     ParseTRA();
                 }
 
-                else if (words[0] == "$PSTI" && (words[1] == "032" || words[1] == "035") )
+                else if (words[0] == "$PSTI" && words[1] == "036") //Heading, Pitch and Roll Messages from SkyTraq PX1172RH modules... only available if RTK
                 {
-                    ParseSTI032(); //there is also an $PSTI,030,... wich contains different data!
+                    ParseSTI036(); //there are also different $PSTI,0XX,... sentences which contain compete different data!
                 }
             }// while still data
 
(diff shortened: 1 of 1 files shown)


FantasyTeddy:
Even though the older messages (`PSTI,032` and `PSTI,035`) are outdated, wouldn't it be nice if we could still support them?

Out of curiosity: is there some kind of specification for these messages, or have they been reverse-engineered?

THiamu:
> Even though the older messages (`PSTI,032` and `PSTI,035`) are outdated, wouldn't it be nice if we could still support them?

Probably yes, but then we have to make sure not to use them when we get the `PSTI,036` message. `032` and `035` are giving like raw data -> distances and course between two antennas. To get roll or heading you need to do some calculation - which was in the old code. With newer `036` messages the receivers does the calculation and gives the completed values especially for roll.
I think I will do some further editing on the code so it will work in each use case of SkyTraq receivers. The old wouldnt have worked if the wrong of two receivers were connected as I read the code. It also has been edited the last few years after my first version...
To have it always working we need it to do like -> IF `PSTI,036` then dont use `035` or `032`; and also IF `PSTI,035` then dont use `032` otherwise we would get doubled or even conflicting data.

 
> Out of curiosity: is there some kind of specification for these messages, or have they been reverse-engineered?

You can see the datasheet of the receivers on their website with messages explained. See here
https://navspark.mybigcommerce.com/content/PX1172RH_DS.pdf

FantasyTeddy (approved):

FantasyTeddy:
Awesome! Thank you very much for your explanation and the quick change to support both old and new.

PR #802, opened 2025-07-25
Adds boundary creation from track lines

richardklasens:
Implements a new feature to build field boundaries from existing track lines. 

- Adds a new tool within the boundary editor to create boundaries by analyzing and connecting logged tracks.
- Includes an OpenGL preview for visualizing the boundary generation process.
- Provides options for adjusting and refining the generated boundary.


SourceCode/GPS/Classes/CBoundaryBuilder.cs
```diff
@@ -0,0 +1,568 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Globalization;
+using System.IO;
+using AgLibrary.Logging;
+
+namespace AgOpenGPS.Classes
+{
+    public class CBoundaryFromTracks
```
FantasyTeddy: If there is a single class in a file, that class should have the same name as the file.

Additionally, I really think we should not use the naming convention of prefixing each class with `C...`

SourceCode/GPS/Classes/CBoundaryBuilder.cs
```diff
@@ -0,0 +1,568 @@
...
+    {
+        #region Constants
```
FantasyTeddy: I am not a huge fan of regions.
- They are useless when viewing the code outside of Visual Studio
- If used for separating functionality, they usually indicate that the code should be split up (e.g. different class)

SourceCode/GPS/Classes/vec3.cs
```diff
@@ -56,8 +57,12 @@ public struct vec2
     {
         public double easting;
         public double northing;
+        public override string ToString()
```
FantasyTeddy: Is this used anywhere?

SourceCode/GPS/Forms/Field/FormBoundary.cs
```diff
@@ -435,6 +439,28 @@ private void btnBingMaps_Click(object sender, EventArgs e)
...
+        {
+            if (mf.bnd.bndList.Count > 0)
+            {
+                var result = FormDialog.Show("Boundary Exists", "A boundary already exists. Do you want to remove it?", MessageBoxButtons.YesNo);
+                if (result != DialogResult.OK) return;
+
+                mf.bnd.bndList.Clear();
+            }
+
+            var tracks = CTrackLineReader.LoadTrackLines(mf.currentFieldDirectory, out string error);
```
FantasyTeddy: Is it necessary that you do the error checking here? It is not very efficient if you load the track lines twice, just to be able to display this error here. Couldn't you do that when really loading the track lines in `FormBuildBoundaryFromTracks`?

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.Designer.cs
FantasyTeddy, line 1: When trying out the tool, I found the track list quite confusing. It took me a while to figure out that I should use the arrows to select the different tracks.
I think it would be nice if I could also select a track in the list without enabling/disabling it at the same time.

SourceCode/GPS/btnImages/BoundaryFromTracks.png
FantasyTeddy, line 1: I'm not sure if this image conveys the meaning of the tool. Why the "plus" icon? Why the arrow? Where are the tracks?

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
```diff
@@ -0,0 +1,753 @@
...
+        private void btnClose_Click(object sender, EventArgs e)
+        {
+            BeginInvoke(new Action(() =>
+            {
+                if (_parentForm is FormBoundary boundaryForm)
+                {
+                    boundaryForm.GetType().GetField("isClosing", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance)
+                        ?.SetValue(boundaryForm, true);
+
+                    boundaryForm.Close();
```
FantasyTeddy: Why don't we just close the `FormBoundary` when opening `FormBuildBoundaryFromTracks`?

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
```diff
@@ -0,0 +1,753 @@
...
+
+            InitializeOpenGL();
+            LoadTracks();
+        }
+
+        private void InitializeOpenGL()
+        {
+            glControlPreview.Load += glControlPreview_Load;
+            glControlPreview.Paint += glControlPreview_Paint;
+            glControlPreview.Resize += glControlPreview_Resize;
```
FantasyTeddy: Can we attach these events in the designer?

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
```diff
@@ -0,0 +1,753 @@
...
+            btnSave.Enabled = false;
+        }
+        protected override void OnPaint(PaintEventArgs e)
+        {
+            base.OnPaint(e);
+
+            using (var pen = new Pen(Color.DarkGray,15))
+            {
+                var rect = new Rectangle(0, 0, this.Width - 1, this.Height - 1);
+                e.Graphics.DrawRectangle(pen, rect);
```
FantasyTeddy: Do we really need that border? If so, could we add it in the designer somehow?

SourceCode/GPS/Classes/CTrackLineReader.cs
```diff
@@ -0,0 +1,90 @@
...
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+
+namespace AgOpenGPS.Classes
+{
+    internal static class CTrackLineReader
+    {
+        public static List<CTrk> LoadTrackLines(string fieldDirectory, out string error)
```
FantasyTeddy: I don't think that this is a good use case for an `out` parameter.
Consider using a tuple as a return value instead.

We should also seriously consider if it is a good idea for each functionality to implement its own load method. Couldn't we reuse the functionality in `SaveOpen.Designer.cs` (`FileLoadTracks()`) somehow?

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.Designer.cs
```diff
@@ -0,0 +1,290 @@
...
+            this.btnRefreshPreview.BackColor = System.Drawing.Color.Transparent;
+            this.btnRefreshPreview.FlatAppearance.BorderSize = 0;
+            this.btnRefreshPreview.FlatStyle = System.Windows.Forms.FlatStyle.Flat;
+            this.btnRefreshPreview.Font = new System.Drawing.Font("Microsoft Sans Serif", 12F, System.Drawing.FontStyle.Bold, System.Drawing.GraphicsUnit.Point, ((byte)(0)));
+            this.btnRefreshPreview.Image = global::AgOpenGPS.Properties.Resources.AutoTrack;
+            this.btnRefreshPreview.Location = new System.Drawing.Point(578, 623);
+            this.btnRefreshPreview.Name = "btnRefreshPreview";
+            this.btnRefreshPreview.Size = new System.Drawing.Size(95, 95);
+            this.btnRefreshPreview.TabIndex = 2;
+            this.btnRefreshPreview.Text = "Refresh";
```
FantasyTeddy: I would call this "Reset" to indicate that everything will be discarded.

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.Designer.cs
```diff
@@ -0,0 +1,290 @@
...
+            this.btnBuildBoundary.UseVisualStyleBackColor = false;
+            this.btnBuildBoundary.Click += new System.EventHandler(this.btnBuildBoundary_Click);
+            // 
+            // btnRefreshPreview
+            // 
+            this.btnRefreshPreview.BackColor = System.Drawing.Color.Transparent;
+            this.btnRefreshPreview.FlatAppearance.BorderSize = 0;
+            this.btnRefreshPreview.FlatStyle = System.Windows.Forms.FlatStyle.Flat;
+            this.btnRefreshPreview.Font = new System.Drawing.Font("Microsoft Sans Serif", 12F, System.Drawing.FontStyle.Bold, System.Drawing.GraphicsUnit.Point, ((byte)(0)));
+            this.btnRefreshPreview.Image = global::AgOpenGPS.Properties.Resources.AutoTrack;
```
FantasyTeddy: Maybe consider using a slightly different icon for the reset button.
(For example the `Reset_Default` icon without the steering wheel inside would be perfect.)

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
```diff
@@ -0,0 +1,753 @@
...
+        }
+        #endregion
+
+        #region Track Manipulation
+        private void ShiftSelectedTrackPoint(bool isPointA, double deltaMeters)
+        {
+            var trk = _activeTrack;
+            if (trk == null) return;
+
+            btnBuildBoundary.Enabled = true;
```
FantasyTeddy: Why is the "Build" button only enabled after shifting a track once?

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
```diff
@@ -0,0 +1,753 @@
...
+
+        #region View Management
+        private void UpdateViewBounds()
+        {
+            ResetViewBounds();
+            CalculateViewBounds();
+            ApplyViewMargins();
+        }
+
+        private void ResetViewBounds()
```
FantasyTeddy: I would merge this method with the `CalculateViewBounds` method. Resetting the values alone would not produce valid bounds, it needs to be called together with the calculation.

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
```diff
@@ -0,0 +1,753 @@
...
+                    UpdateMinMax(trk.ptB);
+                }
+                else if (trk.mode == TrackMode.Curve)
+                {
+                    foreach (var pt in trk.curvePts)
+                        UpdateMinMax(new vec2(pt.easting, pt.northing));
+                }
+            }
+
+            if (double.IsInfinity(_viewLeft))
```
FantasyTeddy: Why only check the left value for infinity?

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
```diff
@@ -0,0 +1,753 @@
...
+            double margin = Math.Max(10, (_viewRight - _viewLeft) * VIEW_MARGIN_FACTOR);
+            _viewLeft -= margin;
+            _viewRight += margin;
+            _viewTop += margin;
+            _viewBottom -= margin;
+        }
+
+        private void RequestRedraw()
+        {
+            if (!_redrawPending)
```
FantasyTeddy, line 386: Is this check really necessary? Shouldn't the `Invalidate` method on the OpenGL control do something similar already?

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
```diff
@@ -0,0 +1,753 @@
...
+
+        private void glControlPreview_Resize(object sender, EventArgs e)
+        {
+            SetupViewport();
+            RequestRedraw();
+        }
+
+        private void SetupViewport()
+        {
+            glControlPreview.MakeCurrent();
```
FantasyTeddy: Shouldn't this be called at the beginning of `InitializeGL` already?

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
```diff
@@ -0,0 +1,753 @@
...
+        private void glControlPreview_Paint(object sender, PaintEventArgs e)
+        {
+            RenderScene();
+        }
+
+        private void RenderScene()
+        {
+            glControlPreview.MakeCurrent();
+            GL.Clear(ClearBufferMask.ColorBufferBit | ClearBufferMask.DepthBufferBit);
+            SetupViewport();
```
FantasyTeddy, line 441: Why do you need to setup the viewport again here?

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
```diff
@@ -0,0 +1,753 @@
...
+
+            GL.End();
+        }
+
+        private void DrawCircle(vec2 center, Color color, float radius)
+        {
+            GL.Color3(color);
+            GL.Begin(PrimitiveType.LineLoop);
+
+            for (int i = 0; i < 16; i++)
```
FantasyTeddy: Here you could use another constant for the number of segments (16).

SourceCode/GPS/Forms/FormGPS.Designer.cs
```diff
@@ -3190,7 +3190,7 @@ private void InitializeComponent()
         public System.Windows.Forms.Button btnAutoSteer;
         private System.Windows.Forms.HScrollBar hsbarSteerAngle;
         private OpenTK.GLControl oglZoom;
-        private OpenTK.GLControl oglMain;
+        internal OpenTK.GLControl oglMain;
```
FantasyTeddy: I would not expose the entire control. Instead consider adding a method to `FormGPS` that calls the `Invalidate` method (like `InvalidateOpenGL` or something).

SourceCode/GPS/Classes/CBoundaryBuilder.cs
```diff
@@ -0,0 +1,568 @@
...
+                return FinalBoundary;
+            }
+            catch (Exception ex)
+            {
+                Log.EventWriter($"Error building boundary: {ex.Message}");
+                return new List<vec3>();
+            }
+        }
+
+        public bool SaveToBoundaryFile(string fieldDirectory)
```
FantasyTeddy: Maybe you could move all the code for saving the boundary into another class? (Similar to how you have all the code for loading the tracks in `CTrackLineReader`)

SourceCode/GPS/Classes/CBoundaryBuilder.cs
```diff
@@ -0,0 +1,568 @@
...
+
+        #region Public API
+        public void SetTracks(List<CTrk> tracks)
+        {
+            try
+            {
+                InputTracks = tracks?.ToList() ?? new List<CTrk>();
+                Log.EventWriter("Tracks set successfully");
+            }
+            catch (Exception ex)
```
FantasyTeddy: What kind of exception could occur here?

SourceCode/GPS/Classes/CBoundaryBuilder.cs
```diff
@@ -0,0 +1,568 @@
...
+                    double maxY1 = Math.Max(s1.Start.northing, s1.End.northing) + tolerance;
+
+                    double minX2 = Math.Min(s2.Start.easting, s2.End.easting);
+                    double maxX2 = Math.Max(s2.Start.easting, s2.End.easting);
+                    double minY2 = Math.Min(s2.Start.northing, s2.End.northing);
+                    double maxY2 = Math.Max(s2.Start.northing, s2.End.northing);
+
+                    return !(maxX1 < minX2 || maxX2 < minX1 || maxY1 < minY2 || maxY2 < minY1);
+                }
+        private bool LineSegmentsIntersect(vec2 p1, vec2 p2, vec2 p3, vec2 p4, out vec2 intersection)
```
FantasyTeddy: Again, I'm not a huge fan of `out` parameters like this. Maybe consider returning `vec2?` instead?

SourceCode/GPS/Classes/CBoundaryBuilder.cs
```diff
@@ -0,0 +1,568 @@
...
+            foreach (vec3 pt in bnd.fenceLine)
+            {
+                yield return FormattableString.Invariant(
+                    $"{pt.easting:F6},{pt.northing:F6},{pt.heading:F6}");
+            }
+        }
+
+        #endregion
+
+        #region Static Helpers
```
FantasyTeddy: Maybe consider moving these helper methods either to the `vec2` struct or the `glm` helper class.

FantasyTeddy (commented):
Great job!

Sorry for the many comments. Feel free to ignore any of them that you don't think are relevant.

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
richardklasens: We recalculate the view bounds dynamically whenever a track is shortened, extended, or modified. Since these changes affect the visible coordinate space, the viewport needs to be updated before every render to reflect the current view range. Therefore, calling SetupViewport() inside RenderScene() ensures that OpenGL always uses the correct projection.

FantasyTeddy: I see. I know it would be just a minor improvement, but could we call `SetupViewport()` inside of `UpdateViewBounds()` to achieve the same thing?

SourceCode/GPS/Forms/FormGPS.Designer.cs
richardklasens: Do you have an idea to do it differently without refactoring the OglMain calls in the program?

FantasyTeddy: Yes, you could just add a method like this to `FormGPS`:
```csharp
public void InvalidateOpenGL()
{
    oglMain.Invalidate();
}
```
and then call this method from `FormBuildBoundaryFromTracks` instead.

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.Designer.cs
richardklasens: I like the List as it is, because its used to turn off tracks you dont need to use for the bnd. 
I will add text to the arrows. Its touch-screen friendly...

SourceCode/GPS/btnImages/BoundaryFromTracks.png
richardklasens: It's a combination of add boundary and Guidance Lines (found in the lower right corner on the Mainform)

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
richardklasens: We can't immediately close FormBoundary because it's also opened as a ShowDialog(). Closing it while FormBuildBoundaryFromTracks (also a dialog) is shown would break the UI stack. That's why we use Hide() first and only close it after the new form finishes.

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
richardklasens: Yes, It works best as is

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
richardklasens: Done

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
richardklasens: Yes

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
richardklasens: If i move SetupViewPort() to UpdateViewBounds() it wont work, the MainOGL crashes

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
FantasyTeddy: And what if we close the `FormBoundary` _after_ `ShowDialog()`?

richardklasens: You can’t because FormBoundaryFromTracks is opened as ShowDialog, which is necessary, and then the UI blocks it to anything like closing another form. So. I leave it as “hide” at opening and “close” when closing the form

FantasyTeddy: I'm not sure if we both understand the same.
My suggestion is to add this to the end of the `btnBuildBoundaryFromTracks_Click` method (_after_ the call to `ShowDialog()`) in `FormBoundary`:
```csharp
isClosing = true;
Close();
```

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
richardklasens: Done

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
richardklasens: Moved to Designer

SourceCode/GPS/Classes/CBoundaryBuilder.cs
richardklasens: changed to BoundaryBuilder

SourceCode/GPS/Classes/CBoundaryBuilder.cs
richardklasens: i'll keep it as it is.

SourceCode/GPS/Classes/vec3.cs
richardklasens: No removed

SourceCode/GPS/Forms/Field/FormBoundary.cs
richardklasens: Moved

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
richardklasens: Changed it

SourceCode/GPS/Classes/CTrackLineReader.cs
richardklasens: removed the TrackLineReader and use FileLoadTracks.cs from SaveOpen.Designer.cs

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.Designer.cs
richardklasens: Done

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.Designer.cs
richardklasens: Done

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
richardklasens: Removed, my first thought it was handy, but it isnt

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
richardklasens: Changed

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.cs
richardklasens: add check for all

SourceCode/GPS/Classes/CBoundaryBuilder.cs
richardklasens: I leave it here for now

SourceCode/GPS/Classes/CBoundaryBuilder.cs
richardklasens: Yeah no exception, its already checked earlier.

SourceCode/GPS/Classes/CBoundaryBuilder.cs
richardklasens: Changed

SourceCode/GPS/Classes/CBoundaryBuilder.cs
richardklasens: moved to vec2 struct and made public structs

SourceCode/GPS/Classes/CBoundaryBuilder.cs
```diff
@@ -0,0 +1,532 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Globalization;
+using System.IO;
+using AgLibrary.Logging;
+
+namespace AgOpenGPS.Classes
+{
+    public class BoundaryBuilder
```
FantasyTeddy: Maybe consider also renaming the file to `BoundaryBuilder.cs`

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.Designer.cs
```diff
@@ -0,0 +1,315 @@
...
+        protected override void OnPaint(PaintEventArgs e)
+        {
+            base.OnPaint(e);
+
+            using (var pen = new Pen(Color.DarkGray, 15))
+            {
+                var rect = new Rectangle(0, 0, this.Width - 1, this.Height - 1);
+                e.Graphics.DrawRectangle(pen, rect);
+            }
+        }
```
FantasyTeddy, line 35: That wasn't really what I meant by moving this into the designer file. I meant that you could add a `Panel` or something that has the gray border, without having to do custom drawing.

SourceCode/GPS/Forms/OpenGL.Designer.cs
```diff
@@ -678,6 +678,11 @@ private void oglMain_Paint(object sender, PaintEventArgs e)
 
             }
         }
+        public void InvalidateOpenGL()
```
FantasyTeddy: Do you call this somewhere?

SourceCode/GPS/Classes/CBoundaryBuilder.cs
richardklasens: No all other classes start with a C, so in the case of consistency it's better to do not in my brains 😀

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.Designer.cs
richardklasens: But this is seventeen times easier

SourceCode/GPS/Forms/OpenGL.Designer.cs
richardklasens: No it wasn't needed to call it

SourceCode/GPS/Classes/CBoundaryBuilder.cs
FantasyTeddy: Consistency should be considered when two options are otherwise mostly equivalent.
In this case however, I think we should really look forward and try to not carry over too many bad practices...

SourceCode/GPS/Forms/Field/FormBuildBoundaryFromTracks.Designer.cs
FantasyTeddy: Okay... I hope that does not mess with the designer...

SourceCode/GPS/Forms/OpenGL.Designer.cs
FantasyTeddy: So... we can remove it? ;)

FantasyTeddy (approved):

PR #819, opened 2025-08-16
Replaces `WebEye.Controls.WinForms.WebCameraControl` with `OpenCvSharp4.VideoCapture`

FantasyTeddy:
Uses the `VideoCapture` from `OpenCvSharp4` to render the web cam.

Unfortunately, there is no easy way to get a list of all available cameras. This was needed to populate the combo box and allow users to choose which camera to display.
Is this functionality actually needed? If so, there is quite some more work that needs to be done.

### DIFF
--- a/SourceCode/AgDiag/AgDiag.csproj
+++ b/SourceCode/AgDiag/AgDiag.csproj
@@ -22,7 +22,7 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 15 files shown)


richardklasens:
Choosing a camera should sure be an option.. some have built in webcam and put in usb cam. So you need to pick the right one

FantasyTeddy:
Okay, then this isn't the way to go.

Any suggestions what I should try instead?

PR #822, opened 2025-08-17
Enhances lightbar readability with text background

richardklasens:
Improves the lightbar display by adding a text background for better readability of the cross-track error.

Key changes:
- Makes it the same behavior as SteerBarGuidanceBar (Center block)
- Maintains original visual style (dot sizes, colors, spacing).
- Implements dynamic text coloring based on error magnitude, similar to the steer bar.


SourceCode/GPS/Forms/OpenGL.Designer.cs
```diff
@@ -1934,100 +1934,116 @@ private void DrawFlags()
             catch { }
         }
 
+        // Helper: draw a single arrow as a triangle, with optional halo ---
+        private void DrawArrowTriangle(float cx, float cy, float size, bool isRight, bool isHalo)
```
FantasyTeddy: The parameter `isHalo` is not used.

FantasyTeddy: This is minor, but why not use `double` instead of `float`?

richardklasens: Could be for sure but I don't think we need that kind of accuracy for the drawing?

FantasyTeddy: It's much less for accuracy as it is for consistency. Almost everywhere else we are already using `double` (and you even have to cast to `float` explicitly to make this work).

SourceCode/GPS/btnImages/Config/HeadlandDistance.png
FantasyTeddy, line 1: I think this got added accidentally?

richardklasens: No it's for the Config Form

PR #825, opened 2025-08-17
Adds RTK fix recovery sound

richardklasens:
Adds a sound to notify the user when RTK fix is recovered after a loss.

- Introduces a debounce timer to avoid false positives due to intermittent signal.
- Implements logic to play the recovery sound only once after the RTK fix has been stable for a defined duration.
- Adds new sound resource for RTK fix recovery.

Additional there is more logic added to prevent multiple sounds being played when the Auto Or Manual Section Functions are controlled by the Autosteer Activation


SourceCode/GPS/Resources/rtk_lost.wav
FantasyTeddy, line 1: This does not really sound like an alarm anymore. Is that on purpose?

FantasyTeddy: Much better now :)

PR #826, opened 2025-08-17
Split `FormConfig` into separate controls

FantasyTeddy:
This is a first step towards modularizing the `FormConfig`.
Doing this for every tab should help immensely for the following:
- Reduces the size of the `FormConfig.Designer.cs` file.
- Hopefully prevent future merge conflicts.
- Maybe get rid of the various `*.Designer.cs` files under the `FormConfig`.


richardklasens:
Good approach to move each tab into its own UserControl It makes the main FormConfig much simpler, and avoids mixing logic into designer partials. Thanks for this!
@J005t67 What do you think?

FantasyTeddy:
If this is the right direction, let's merge this and continue gradually with the other tabs.

PR #829, opened 2025-08-19
Adds global hotkey support

richardklasens:
Implements global hotkey handling using a message filter.

This allows the application to respond to key presses even when it's not the active window.
A new helper class is introduced to manage the message filtering.
The hotkey filter checks if the user is typing in a text box and ignores the hotkey if they are.

### DIFF
--- a/SourceCode/GPS/Forms/UDPComm.Designer.cs
+++ b/SourceCode/GPS/Forms/UDPComm.Designer.cs
@@ -464,6 +464,41 @@ namespace AgOpenGPS
         }
 
         #region keystrokes
+
+        private HotkeyMessageFilter _hotkeyFilter;
+
+        // Will be used to handle app-wide hotkeys
+        public bool HandleAppWideKey(Keys key, Keys mods)
+        {
+            // build same data as in ProcessCmdKey
+            var keyData = key | mods;
+            // Create a dummy key
+            var msg = Message.Create(IntPtr.Zero, 0, IntPtr.Zero, IntPtr.Zero);
+            return ProcessCmdKey(ref msg, keyData);
+        }
+
+        protected override void OnHandleCreated(EventArgs e)
+        {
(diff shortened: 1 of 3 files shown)


richardklasens:
this fixes #829

SourceCode/GPS/Helpers/HotkeyMessageFilter.cs
```diff
@@ -0,0 +1,44 @@
...
+        private const int WM_KEYDOWN = 0x0100;
+        private const int WM_SYSKEYDOWN = 0x0104;
+
+        public HotkeyMessageFilter(FormGPS mf) => _mf = mf ?? throw new ArgumentNullException(nameof(mf));
+
+        public bool PreFilterMessage(ref Message m)
+        {
+            if (m.Msg != WM_KEYDOWN && m.Msg != WM_SYSKEYDOWN) return false;
+
+            // Als gebruiker in een tekstbox aan het typen is => niet onderscheppen
```
FantasyTeddy: We have some Dutch comments in here :)

FantasyTeddy (commented):
I guess you have seen the warning for [`Application.AddMessageFilter(IMessageFilter)`](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.application.addmessagefilter):

> Adding message filters to the message pump for an application can degrade performance.

SourceCode/GPS/Forms/UDPComm.Designer.cs
```diff
@@ -465,7 +465,7 @@ protected override CreateParams CreateParams
 
         #region keystrokes
 
-        private HotkeyMessageFilter _hotkeyFilter;
+        private Helpers.HotkeyMessageFilter _hotkeyFilter;
```
FantasyTeddy: Can we do a `using AgOpenGPS.Helpers` instead of prefixing the namespace?

PR #832, opened 2025-08-22
Refactors field loading with stateless file operations

richardklasens:
This refactor introduces a stateless file loading mechanism for field data, improving code maintainability and testability.

Key changes:
- Implements stateless readers and writers for field assets, ensuring no UI dependencies or global state access.
- Introduces small DTOs to manage parsed data for various file types.
- Updates the job form UX to use asynchronous operations, enhancing responsiveness.

This approach streamlines file operations, making the codebase more robust and easier to extend.


SourceCode/GPS/Forms/Field/FormJob.cs
```diff
@@ -83,142 +83,183 @@ private void FormJob_Load(object sender, EventArgs e)
...
-            Log.EventWriter("Job Form, Field Resume");
-
-            //back to FormGPS
-            DialogResult = DialogResult.OK;
-            Close();
+            btnJobNew.Enabled = false;
+            UseWaitCursor = true;
+            try
+            {
+                await SaveIfJobStartedAsync();
```
FantasyTeddy: In UI applications (WinForms included) it is important to marshal back to the UI thread. We should therefore think about requiring `ConfigureAwait(true)` explicitly to communicate that properly. There are also analyzers out there that can check for that.

SourceCode/GPS/Forms/Field/FormJob.cs
```diff
@@ -83,142 +83,183 @@ private void FormJob_Load(object sender, EventArgs e)
...
+        private async void btnJobResume_Click(object sender, EventArgs e)
         {
-            if (mf.isJobStarted)
+            btnJobResume.Enabled = false;
+            UseWaitCursor = true;
+            try
             {
-                _ = mf.FileSaveEverythingBeforeClosingField();
-            }
+                // Optional: als er echt een job draait, eerst netjes saven
```
FantasyTeddy: Found another Dutch comment :)

SourceCode/GPS/Forms/Field/FormJob.cs
```diff
@@ -83,142 +83,183 @@ private void FormJob_Load(object sender, EventArgs e)
...
+                btnJobResume.Enabled = true;
             }
         }
 
+        private async void btnJobOpen_Click(object sender, EventArgs e)
+        {
+            try
+            {
+                btnJobOpen.Enabled = false;
+                UseWaitCursor = true;
```
FantasyTeddy: To be consistent, this should be done outside the `try`.

SourceCode/GPS/Forms/Field/FormJob.cs
```diff
@@ -83,142 +83,183 @@ private void FormJob_Load(object sender, EventArgs e)
...
+
+                if (mf.isJobStarted)
+                {
+                    await mf.FileSaveEverythingBeforeClosingField();
+                }
```
FantasyTeddy:
```suggestion
                await SaveIfJobStartedAsync();
```

SourceCode/GPS/Forms/Field/FormJob.cs
```diff
@@ -83,142 +83,183 @@ private void FormJob_Load(object sender, EventArgs e)
...
                     {
                         try
                         {
                             // Skip 8 lines
-                            for (int i = 0; i < 8; i++)
-                            {
-                                reader.ReadLine();
-                            }
-                            //start positions
+                            for (int i = 0; i < 8; i++) reader.ReadLine();
```
FantasyTeddy: We should probably define, which of these formatting variants we want to use:
- **Option 1**:
  ```csharp
  if (condition)
  {
      Something();
  }
  ```
- **Option 2**:
  ```csharp
  if (condition)
      Something();
  ```
- **Option 3**:
  ```csharp
  if (condition) Something();
  ```
Personally, I don't like option 3, but it is mostly due to the fact that I'm used to option 2.

SourceCode/GPS/Forms/Field/FormJob.cs
richardklasens: Not a part anymore for this PR

SourceCode/GPS/Forms/Field/FormJob.cs
richardklasens: Not a part anymore

SourceCode/GPS/Forms/Field/FormJob.cs
richardklasens: Not a part anymore

SourceCode/GPS/Forms/Field/FormJob.cs
richardklasens: Option 2 for sure

SourceCode/GPS/Classes/FieldFilesLoader.cs
FantasyTeddy, line 1: The `FieldFilesLoader` and `FieldFilesSaver` are not used anymore and can be removed.

SourceCode/GPS/Classes/IO/BackPicFiles.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System.IO;
+using System.Globalization;
+
+namespace AgOpenGPS.Classes.IO
```
FantasyTeddy: What do you think if we move these files into a folder `Files` (outside of `Classes`) which would result in this namespace:
```suggestion
namespace AgOpenGPS.Files
```

SourceCode/GPS/Forms/SaveOpen.Designer.cs
```diff
@@ -2430,6 +1017,58 @@ private string GetGeoCoordToWgs84_KML(GeoCoord geoCoord)
                 latLon.Longitude.ToString("N7", CultureInfo.InvariantCulture) + ',' +
                 latLon.Latitude.ToString("N7", CultureInfo.InvariantCulture) + ",0 ";
         }
+        public void ExportFieldAs_ISOXMLv3()
```
FantasyTeddy, line 881: Why did you move the two methods `ExportFieldAs_ISOXMLv3` and `ExportFieldAs_ISOXMLv4` down?

SourceCode/GPS/Forms/SaveOpen.Designer.cs
```diff
@@ -27,1143 +23,124 @@ public partial class FormGPS
...
-            }
-            else
-            {
-                using (StreamReader reader = new StreamReader(filename))
-                {
-                    try
-                    {
-                        string line;
+        // NOTE: Add this using at the top of your Form file:
+        // using AgOpenGPS.Classes.IO;
```
FantasyTeddy: Why this comment?

SourceCode/GPS/Forms/SaveOpen.Designer.cs
```diff
@@ -27,1143 +23,124 @@ public partial class FormGPS
...
-                            else
-                                trk.gArr[i].name = words[0];
-
-                            trk.gArr[i].mode = TrackMode.AB;
-
-
-                            trk.gArr[i].heading = glm.toRadians(double.Parse(words[1], CultureInfo.InvariantCulture));
-                            trk.gArr[i].ptA.easting = double.Parse(words[2], CultureInfo.InvariantCulture);
-                            trk.gArr[i].ptA.northing = double.Parse(words[3], CultureInfo.InvariantCulture);
+        #region File IO (per filetype) � updated to AgOpenGPS.Classes.IO helpers
```
FantasyTeddy: Why have this region? Isn't the entire file "IO"?

SourceCode/GPS/Forms/SaveOpen.Designer.cs
```diff
@@ -27,1143 +23,124 @@ public partial class FormGPS
...
-                    }
-                    catch (Exception er)
-                    {
-                        TimedMessageBox(2000, "AB Line Corrupt", "Please delete it!!!");
-
-                        Log.EventWriter("FieldOpen, Loading ABLine, Corrupt ABLine File" + er);
-                    }
-                }
-            }
+        // --- Keep this helper as-is ---
```
FantasyTeddy: Why can't this helper be changed? What's so special about it?

SourceCode/GPS/Classes/IO/BoundaryFiles.cs
```diff
@@ -0,0 +1,109 @@
...
+                result.Add(b);
+                while (idx < lines.Length && string.IsNullOrWhiteSpace(lines[idx])) idx++;
+            }
+
+            return result;
+        }
+
+        public static void Save(string fieldDirectory, IReadOnlyList<CBoundaryList> boundaries)
+        {
+            FileIoUtils.EnsureDir(fieldDirectory);
```
FantasyTeddy: Shouldn't that maybe be part of the `GetFieldDir()` method in `SaveOpen.Designer.cs`? Then you only have to do it once.

SourceCode/GPS/Classes/IO/FileIOUtils.cs
```diff
@@ -0,0 +1,71 @@
...
+using AgOpenGPS.Core.Models;
+
+namespace AgOpenGPS.Classes.IO
+{
+    public static class FileIoUtils
+    {
+        // ---- Formatting helpers ----
+        public static string F3(double v) { return Math.Round(v, 3).ToString(CultureInfo.InvariantCulture); }
+        public static string F5(double v) { return Math.Round(v, 5).ToString(CultureInfo.InvariantCulture); }
+        public static string F1(double v) { return Math.Round(v, 1).ToString(CultureInfo.InvariantCulture); }
```
FantasyTeddy: Can we reduce this to a single method that takes the number of digits as an input? (And maybe also use a more descriptive name?)

SourceCode/GPS/Classes/IO/HeadlinesFiles.cs
```diff
@@ -0,0 +1,95 @@
...
+                        line = reader.ReadLine(); if (line == null) break;
+                        hp.mode = int.Parse(line, CultureInfo.InvariantCulture);
+
+                        line = reader.ReadLine(); if (line == null) break;
+                        hp.a_point = int.Parse(line, CultureInfo.InvariantCulture);
+
+                        line = reader.ReadLine(); if (line == null) break;
+                        int numPoints = int.Parse(line, CultureInfo.InvariantCulture);
+
+                        hp.trackPts = new List<vec3>();
```
FantasyTeddy: No need to initialize the list explicitly (done in the constructor of `CHeadPath`).

SourceCode/GPS/Classes/IO/HeadlinesFiles.cs
```diff
@@ -0,0 +1,95 @@
...
+                        for (int i = 0; i < numPoints && !reader.EndOfStream; i++)
+                        {
+                            var words = (reader.ReadLine() ?? string.Empty).Split(',');
+                            if (words.Length < 3) continue;
+
+                            double e, n, h;
+                            if (double.TryParse(words[0], NumberStyles.Float, CultureInfo.InvariantCulture, out e) &&
+                                double.TryParse(words[1], NumberStyles.Float, CultureInfo.InvariantCulture, out n) &&
+                                double.TryParse(words[2], NumberStyles.Float, CultureInfo.InvariantCulture, out h))
```
FantasyTeddy: Abbreviations tend to increase the cognitive effort to understand something...
```suggestion
                            if (double.TryParse(words[0], NumberStyles.Float, CultureInfo.InvariantCulture, out double easting) &&
                                double.TryParse(words[1], NumberStyles.Float, CultureInfo.InvariantCulture, out double northing) &&
                                double.TryParse(words[2], NumberStyles.Float, CultureInfo.InvariantCulture, out double heading))
```

SourceCode/GPS/Classes/IO/HeadlinesFiles.cs
```diff
@@ -0,0 +1,95 @@
...
+                            }
+                        }
+
+                        if (hp.trackPts.Count > 3) result.Add(hp);
+                    }
+                }
+                catch
+                {
+                    result.Clear();
+                }
```
FantasyTeddy: What kind of exception can occur here? Do we really want to ignore them?

SourceCode/GPS/Forms/SaveOpen.Designer.cs
```diff
@@ -27,1143 +23,124 @@ public partial class FormGPS
...
-                    Log.EventWriter("Saving Curve Line" + er.ToString());
-
-                    return;
-                }
-            }
-
-            if (trk.idx > (trk.gArr.Count - 1)) trk.idx = trk.gArr.Count - 1;
-
-            FileSaveABLines();
-            FileSaveCurveLines();
```
FantasyTeddy, line 333: Where did these two methods go to?

SourceCode/GPS/Classes/IO/TrackFiles.cs
```diff
@@ -0,0 +1,118 @@
...
+using AgOpenGPS.Core.Models;
+
+namespace AgOpenGPS.Classes.IO
+{
+    public static class TrackFiles
+    {
+        public static List<CTrk> Load(string fieldDirectory)
+        {
+            var result = new List<CTrk>();
+            var path = Path.Combine(fieldDirectory ?? "", "TrackLines.txt");
```
FantasyTeddy: Why `?? ""`?

SourceCode/GPS/Classes/IO/TrackFiles.cs
```diff
@@ -0,0 +1,118 @@
...
+            if (!File.Exists(path)) return result;
+
+            var lines = File.ReadAllLines(path);
+            int idx = 0;
```
FantasyTeddy: Why do it like this here, and not by using `ReadLine()` like in the other files?
(It seems way more complicated to have `idx` and keep incrementing it.)

SourceCode/GPS/Classes/IO/FieldPlaneFiles.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System.Globalization;
+using System.IO;
+using AgOpenGPS.Core.Models;
+
+namespace AgOpenGPS.Classes.IO
+{
+    public static class FieldPlaneFiles
+    {
+        public static LocalPlane LoadLocalPlaneOrDefault(string fieldDirectory, Wgs84 fallbackOrigin)
```
FantasyTeddy: I think this method should only return the origin, because that's the `LocalPlane` is handled outside.
```suggestion
        public static Wgs84 LoadLocalPlaneOrDefault(string fieldDirectory, Wgs84 fallbackOrigin)
```

SourceCode/GPS/Forms/SaveOpen.Designer.cs
```diff
@@ -27,1143 +23,124 @@ public partial class FormGPS
...
-                }
-            }
+        // --- Keep this helper as-is ---
+        private string GetFieldDir()
+        {
+            return Path.Combine(RegistrySettings.fieldsDirectory, currentFieldDirectory);
         }
 
-        //function to open a previously saved field, resume, open exisiting, open named field
-        public void FileOpenField(string _openType)
```
FantasyTeddy, line 703: Could we also extract the contents (or most of) this method into a `FieldFiles` class?

SourceCode/GPS/Forms/SaveOpen.Designer.cs
richardklasens: It makes more sense to have them below the Normal Save/Loading since we dont use them inside AgOpenGPS for normal operation

richardklasens: and they will probably will be removed in the next PR

SourceCode/GPS/Forms/SaveOpen.Designer.cs
richardklasens: Removed

SourceCode/GPS/Forms/SaveOpen.Designer.cs
richardklasens: Removed

SourceCode/GPS/Classes/IO/BackPicFiles.cs
richardklasens: Yes good one

SourceCode/GPS/Forms/SaveOpen.Designer.cs
richardklasens: Removed

SourceCode/GPS/Classes/IO/BoundaryFiles.cs
richardklasens: Extend method with EnsureBool and use the caller to verify the FieldDirectory

SourceCode/GPS/Classes/IO/FileIOUtils.cs
richardklasens: fixed

SourceCode/GPS/Forms/SaveOpen.Designer.cs
richardklasens: They aren't used anymore. The last version which used the AbLines.txt and Curvelines.txt was 5.7 i believe.

SourceCode/GPS/Classes/IO/TrackFiles.cs
richardklasens: It was an old leftover from my first idea, but i removed it from all Methods

SourceCode/GPS/Classes/IO/TrackFiles.cs
richardklasens: Good one. I figured it out and changed it due to consistancy to other methods

SourceCode/GPS/Classes/IO/FieldPlaneFiles.cs
richardklasens: Changed

SourceCode/GPS/Forms/SaveOpen.Designer.cs
richardklasens: I made something

SourceCode/GPS/Forms/Field/FormJob.cs
J005t67: Please, not option 2 . This is an accident waiting to happen and we discussed it before. At my work it was even forbidden by the coding standard.

richardklasens: So you prefer we keep
Option 1 @J005t67 ?

FantasyTeddy: @J005t67: Do you mean by "accident", that other statements are mistakenly added outside of the `if` statement, even though they should be part of it?
If so, I would argue that with enforced code formatting (which we now have) this risk is very much reduced, because you can immediately see that the indentation is wrong.
At the same time, option 3 suffers from the same problem (and is even worse in my opinion).

Option 1 would therefore be the only "safe" option, but is also the most verbose one.

SourceCode/GPS/Forms/SaveOpen.Designer.cs
FantasyTeddy: Ah perfect. Removing "dead" code is always nice :)

SourceCode/GPS/Classes/IO/BackPicFiles.cs
FantasyTeddy: I see, you didn't like `Files` :)

SourceCode/GPS/Forms/SaveOpen.Designer.cs
```diff
@@ -2430,6 +1001,93 @@ private string GetGeoCoordToWgs84_KML(GeoCoord geoCoord)
...
+            catch (Exception e)
+            {
+                Log.EventWriter("Export Field as ISOXML: " + e.Message);
+            }
+        }
+
+        /// <summary>
+        /// Run a loader that returns a value. On error: log, show TimedMessageBox, and return the provided fallback.
+        /// </summary>
+        private (bool ok, T value) TryLoad<T>(string context, Func<T> loader, T fallback)
```
FantasyTeddy: Usually, methods that start with `Try` work slightly differently (see [`TryParse`](https://learn.microsoft.com/de-de/dotnet/api/system.double.tryparse)):
```suggestion
        private bool TryLoad<T>(string context, Func<T> loader, out T result)
```
(That way you also don't need to pass the `fallback` value, because you can do that logic easily when calling the method.)

SourceCode/GPS/IO/TrackFiles.cs
```diff
@@ -0,0 +1,158 @@
...
+            {
+                // Skip optional header
+                var firstLine = reader.ReadLine();
+                if (firstLine == null) return result;
+                if (!firstLine.TrimStart().StartsWith("$", StringComparison.Ordinal))
+                {
+                    // rewind logic: treat this line as first "name"
+                    reader.BaseStream.Seek(0, SeekOrigin.Begin);
+                    reader.DiscardBufferedData();
+                }
```
FantasyTeddy: Why do we need this fallback logic?

SourceCode/GPS/IO/TrackFiles.cs
```diff
@@ -0,0 +1,158 @@
...
+
+                while (!reader.EndOfStream)
+                {
+                    // --- Name ---
+                    var name = (reader.ReadLine() ?? string.Empty).Trim();
+                    if (string.IsNullOrEmpty(name)) continue;
+
+                    // --- Average heading ---
+                    if (reader.EndOfStream) break;
```
FantasyTeddy: Do we really need to be that defensive in this loop?

SourceCode/GPS/IO/FileIOUtils.cs
```diff
@@ -0,0 +1,72 @@
...
+            Directory.CreateDirectory(fieldDirectory);
+        }
+
+        public static void SkipBlanks(string[] lines, ref int idx)
+        {
+            while (idx < lines.Length && string.IsNullOrWhiteSpace(lines[idx])) idx++;
+        }
+
+        // ---- Parsing helpers ----
+        public static bool TryParseEN(string line, out double e, out double n)
```
FantasyTeddy: Do not abbreviate:
```suggestion
        public static bool TryParseEastingNorthing(string line, out double easting, out double northing)
```
or even better:
```suggestion
        public static bool TryParseVec2(string line, out vec2 result)
```

SourceCode/GPS/IO/TrackFiles.cs
```diff
@@ -0,0 +1,158 @@
...
+
+                    // --- B point ---
+                    if (reader.EndOfStream) break;
+                    var bLine = reader.ReadLine();
+                    double bE, bN;
+                    if (!FileIoUtils.TryParseEN(bLine, out bE, out bN)) continue;
+
+                    // --- Nudge (skip) ---
+                    if (reader.EndOfStream) break;
+                    reader.ReadLine();
```
FantasyTeddy: Why skip the "nudge" value?

SourceCode/GPS/IO/TrackFiles.cs
```diff
@@ -0,0 +1,158 @@
...
+                {
+                    // --- Name ---
+                    var name = (reader.ReadLine() ?? string.Empty).Trim();
+                    if (string.IsNullOrEmpty(name)) continue;
+
+                    // --- Average heading ---
+                    if (reader.EndOfStream) break;
+                    var headingLine = reader.ReadLine() ?? string.Empty;
+                    double avgHeading;
+                    if (!double.TryParse(headingLine.Trim(), NumberStyles.Float, CultureInfo.InvariantCulture, out avgHeading))
```
FantasyTeddy: I'm actually not sure if it is a good idea to use `TryParse` everywhere instead of `Parse`. That way we silently ignore invalid values instead of complaining about it... is that really what we want?

SourceCode/GPS/IO/TrackFiles.cs
```diff
@@ -0,0 +1,158 @@
...
+                    // --- Mode ---
+                    int mode = 0;
+                    if (!reader.EndOfStream)
+                    {
+                        var modeLine = reader.ReadLine() ?? string.Empty;
+                        int.TryParse(modeLine.Trim(), NumberStyles.Integer, CultureInfo.InvariantCulture, out mode);
+                    }
+
+                    // --- Visibility (skip) ---
+                    if (!reader.EndOfStream) reader.ReadLine();
```
FantasyTeddy: Same question here: why skip the "visibility"?

SourceCode/GPS/IO/TrackFiles.cs
```diff
@@ -0,0 +1,158 @@
...
+                    {
+                        var tr = new CTrk
+                        {
+                            name = name,
+                            mode = TrackMode.Curve,
+                            curvePts = curvePts,
+                            heading = avgHeading
+                        };
+                        result.Add(tr);
+                    }
```
FantasyTeddy: This is not how it worked before; the mode was not considered when creating the `CTrk`. Are you sure that this is correct?

SourceCode/GPS/Forms/Field/FormJob.cs
J005t67: I prefer option 1. I like it better than option 3. In option 1 you can set the breakpoint inside the if.

FantasyTeddy: Visual Studio actually doesn't care. You can set a breakpoint for all three options on the statement inside the if.

SourceCode/GPS/Forms/Controls.Designer.cs
```diff
@@ -490,36 +490,20 @@ private void toolStripBtnFieldTools_Click(object sender, EventArgs e)
 
         private void btnJobMenu_Click(object sender, EventArgs e)
         {
+            // Remember current state before opening the Job dialog
+            var wasJobStarted = isJobStarted;
+            var prevFieldDir = currentFieldDirectory;
 
             Form f = Application.OpenForms["FormGPSData"];
+            if (f != null) { f.Focus(); f.Close(); }
```
FantasyTeddy: Do we really want to format like this?

SourceCode/GPS/Forms/Pickers/FormFilePicker.cs
```diff
@@ -39,10 +44,23 @@ private void FormFilePicker_Load(object sender, EventArgs e)
...
+
+            string[] dirs;
+            try
+            {
+                dirs = Directory.GetDirectories(RegistrySettings.fieldsDirectory);
+            }
+            catch
+            {
+                dirs = Array.Empty<string>();
+            }
```
FantasyTeddy, line 63: Catching every possible exception is rarely what you want: what if something like an `OutOfMemoryException` occurs? Do you really want to ignore that and hope that the application will recover?

SourceCode/GPS/Forms/SaveOpen.Designer.cs
```diff
@@ -2405,31 +1034,145 @@ private void FileUpdateAllFieldsKML()
...
+            {
+                Log.EventWriter($"[Load:{fileLabel}] failed: {ex}");
+                if (criticality == LoadCriticality.Required)
+                {
+                    TimedMessageBox(2500, gStr.gsFieldFileIsCorrupt, $"{fileLabel} is required and could not be processed.");
+                }
+                else
+                {
+                    TimedMessageBox(2000, "Optional file problem", $"{fileLabel} is missing or corrupt; it will be recreated on save.");
+                }
```
FantasyTeddy, line 982: How does a `Required` error differ from an `Optional` one? (Except from the error message)

SourceCode/GPS/IO/HeadlinesFiles.cs
```diff
@@ -0,0 +1,93 @@
...
+    {
+        public static List<CHeadPath> Load(string fieldDirectory)
+        {
+            var result = new List<CHeadPath>();
+            var path = Path.Combine(fieldDirectory, "Headlines.txt");
+            if (!File.Exists(path)) return result;
+
+            using (var reader = new StreamReader(path))
+            {
+                try
```
FantasyTeddy: There is still a `try`-`catch` in here that should not be...

SourceCode/GPS/IO/TrackFiles.cs
```diff
@@ -0,0 +1,150 @@
...
+                throw new ArgumentNullException(nameof(fieldDirectory));
+
+            var filename = Path.Combine(fieldDirectory, "TrackLines.txt");
+
+            using (var writer = new StreamWriter(filename, false))
+            {
+                writer.WriteLine("$TrackLines");
+                if (tracks == null || tracks.Count == 0) return;
+
+                for (int i = 0; i < tracks.Count; i++)
```
FantasyTeddy:
```suggestion
                foreach (CTrk track in tracks)
```

SourceCode/GPS/Forms/SaveOpen.Designer.cs
```diff
@@ -2405,31 +1034,145 @@ private void FileUpdateAllFieldsKML()
...
+        // Ensures a file exists; triggers TryLoad-based message when missing.
+        private bool EnsureFileExists(string dir, string fileName, LoadCriticality criticality, string label)
+        {
+            bool ok;
+            return TryLoad(label, criticality, () =>
+            {
+                var p = Path.Combine(dir, fileName);
+                if (!File.Exists(p))
+                {
+                    throw new FileNotFoundException($"{fileName} is missing", p);
```
FantasyTeddy: Don't you think it would be better to inform the user about what exactly went wrong? You are not using the message of the exception in the `TryLoad` method, so the user will only see that "... it could not be loaded" which is not wrong, but it also hides information that could be useful for finding the issue.

SourceCode/GPS/Forms/SaveOpen.Designer.cs
```diff
@@ -2405,31 +1034,145 @@ private void FileUpdateAllFieldsKML()
...
+                else
+                {
+                    TimedMessageBox(2000, "Optional file problem", $"{fileLabel} is missing or corrupt; it will be recreated on save.");
+                }
+                return false;
+            }
         }
 
+        // Ensures a file exists; triggers TryLoad-based message when missing.
+        private bool EnsureFileExists(string dir, string fileName, LoadCriticality criticality, string label)
```
FantasyTeddy: I'm not sure this method name is appropriate; "ensure" usually means that after calling the method, something is now existing for sure. So I would expect the method to create the file if it does not exist (which it doesn't and shouldn't).

SourceCode/GPS/IO/TrackFiles.cs
```diff
@@ -0,0 +1,150 @@
...
+            if (!File.Exists(path)) return result;
+
+            using (var reader = new StreamReader(path))
+            {
+                // Require header
+                var header = reader.ReadLine();
+                if (header == null || !header.TrimStart().StartsWith("$", StringComparison.Ordinal))
+                    throw new InvalidDataException("TrackLines.txt missing $ header.");
+
+                while (true)
```
FantasyTeddy: I think having an explicit exit condition would be better:
```suggestion
                while (!reader.EndOfStream)
```

SourceCode/GPS/IO/TrackFiles.cs
```diff
@@ -0,0 +1,150 @@
...
+                    if (headingLine == null) throw new InvalidDataException("Unexpected EOF after track name.");
+                    var heading = double.Parse(headingLine.Trim(), CultureInfo.InvariantCulture);
+
+
+                    // --- A point (easting,northing) ---
+                    var aLine = reader.ReadLine();
+                    if (aLine == null) throw new InvalidDataException("Unexpected EOF reading point A.");
+                    var aParts = aLine.Split(',');
+                    var aE = double.Parse(aParts[0], CultureInfo.InvariantCulture);
+                    var aN = double.Parse(aParts[1], CultureInfo.InvariantCulture);
```
FantasyTeddy: Please use more descriptive variable names than `aE` or `aN`.
Maybe you could already create the `vec2` here to avoid having to write "easting" and "heading" a lot.

SourceCode/GPS/IO/FieldFiles.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Collections.Generic;
+
+namespace AgOpenGPS.Classes.IO
+{
+    /// <summary>
+    /// Provides one-shot loading of all field-related files into a FieldData object.
+    /// </summary>
+    public static class FieldFiles
+    {
+        public static FieldData Load(string fieldDirectory)
```
FantasyTeddy: Shouldn't you be using this somewhere? The `FileOpenField` method in `SaveOpen.Designer.cs` is still very big and should probably be replaced by this?

SourceCode/GPS/Forms/SaveOpen.Designer.cs
richardklasens: I don’t care what the name is, it makes sense to me but if your approach is better, good to change it?

SourceCode/GPS/Forms/SaveOpen.Designer.cs
richardklasens: That should be changed yes.. good one

SourceCode/GPS/Forms/SaveOpen.Designer.cs
richardklasens: Required is Field.txt and Boundary.txt (which can be empty) 
The other ones are created when you actually use it. So good to log,  but no crash needed

SourceCode/GPS/Forms/Pickers/FormFilePicker.cs
richardklasens: Yeah but I’m not sure how to implement this?

SourceCode/GPS/Forms/Controls.Designer.cs
richardklasens: This is one of those option 1/2/3 things… 
Maybe change it to option 1?

SourceCode/GPS/IO/FieldFiles.cs
richardklasens: Yes maybe it’s good to do that with this class and clean the SaveOpen.Designer

SourceCode/GPS/Forms/Controls.Designer.cs
FantasyTeddy: Not quite... this is not even one of the options. The previous discussion was about single-line `if` statements.
This contains multiple statements and in my opinion, as soon as braces (`{}`) are involved, it needs to be multi-line. (And only a single statement per line.)

SourceCode/GPS/Forms/SaveOpen.Designer.cs
FantasyTeddy: That's what I gathered from the context, yes.
However, my question is: the current implementation only shows a different message box... no difference in behavior (like crashing, not crashing).

SourceCode/GPS/Forms/Pickers/FormFilePicker.cs
FantasyTeddy: Okay, let's leave it for the time being.

SourceCode/GPS/Forms/Controls.Designer.cs
richardklasens: Ah okay I will fix

SourceCode/GPS/Forms/SaveOpen.Designer.cs
richardklasens: The field won't load...

```suggestion
        private bool TryLoad<T>(string fileLabel, LoadCriticality criticality, Func<T> loader, out T result)
        {
            try
            {
                result = loader();
                return true;
            }
            catch (Exception)
            {
                Log.EventWriter($"[Load:{fileLabel}] failed");
                if (criticality == LoadCriticality.Required)
                {
                    TimedMessageBox(2500, gStr.gsFieldFileIsCorrupt, $"{fileLabel} is required and could not be loaded.");
                }
                else
                {
                    TimedMessageBox(2000, "Optional file problem", $"{fileLabel} is missing or corrupt but Field is Loaded");
                }
                result = default(T);
                return false;
            }
        }
```

Here the return will be false if the Boundary is missing.

SourceCode/GPS/Forms/SaveOpen.Designer.cs
richardklasens: I removed the function, the TryLoad will do the handling

SourceCode/GPS/IO/FieldFiles.cs
richardklasens: I cleaned it and started a DTO in FieldFiles.cs

SourceCode/GPS/Forms/SaveOpen.Designer.cs
richardklasens: Removed

SourceCode/GPS/Forms/SaveOpen.Designer.cs
FantasyTeddy: Yes, but it will be `false` regardless of whether the criticality is `Required` or `Optional`. There are no consequences except having a different message.

FantasyTeddy: Okay, I think I understand now. You are handling the "consequences" of the `TryLoad` return value when calling it... Let's keep this for the time being.

SourceCode/GPS/IO/ContourFiles.cs
```diff
@@ -0,0 +1,72 @@
...
+            using (var reader = new StreamReader(path))
+            {
+                string first = reader.ReadLine();
+                int verts0;
+                if (!string.IsNullOrEmpty(first) && !first.TrimStart().StartsWith("$") &&
+                    int.TryParse(first.Trim(), NumberStyles.Integer, CultureInfo.InvariantCulture, out verts0))
+                {
+                    var patch0 = FileIoUtils.ReadVec3Block(reader, verts0);
+                    if (patch0.Count > 0) result.Add(patch0);
+                }
```
FantasyTeddy: Why does the first line need to be handled differently?

SourceCode/GPS/IO/BoundaryFiles.cs
```diff
@@ -0,0 +1,140 @@
...
+    public static class BoundaryFiles
+    {
+        public static List<CBoundaryList> Load(string fieldDirectory)
+        {
+            var result = new List<CBoundaryList>();
+            var path = Path.Combine(fieldDirectory, "Boundary.txt");
+            if (!File.Exists(path)) return result;
+
+            var lines = File.ReadAllLines(path);
+            int idx = 0;
```
FantasyTeddy: This is again way more complicated than the previous implementation that just read each line by line...

SourceCode/GPS/IO/HeadlandFiles.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        public static void AttachLoad(string fieldDirectory, List<CBoundaryList> boundaries)
+        {
+            if (boundaries == null || boundaries.Count == 0) return;
+
+            var path = Path.Combine(fieldDirectory, "Headland.txt");
+            if (!File.Exists(path)) return;
+
+            var lines = File.ReadAllLines(path);
+            int idx = 0;
```
FantasyTeddy: Here too... can we do this line by line again?

SourceCode/GPS/IO/ContourFiles.cs
richardklasens: Changed

SourceCode/GPS/IO/BoundaryFiles.cs
richardklasens: Changed to read each line by line

SourceCode/GPS/IO/HeadlandFiles.cs
richardklasens: solved also

PR #836, opened 2025-08-24
First attempt at writing steering autotuning code

Krisz00000000000:
First time using github, hope this is OK There is a markdown file explaining the changes, but copying here for reference.

# AgOpenGPS Auto-Tuning Feature - Technical Description

## Overview

The auto-tuning feature is an intelligent system that automatically adjusts steering parameters in the AgOpenGPS precision agriculture software to optimize vehicle guidance performance. This system monitors real-time performance metrics and dynamically tunes critical parameters to maintain optimal steering accuracy and stability under varying field conditions.

### 1. Parameter Auto-Adjustment
The system automatically tunes the following critical steering parameters:
- **Proportional Gain**: Controls steering responsiveness to guidance errors
- **Max Limit**: Maximum steering output to prevent over-correction
- **Steer Response (goalPointLookAheadHold)**: Influences lookahead behavior for path following
- **Integral Gain**: Eliminates steady-state error in steering
- **Speed Factor**: Adapts steering behavior based on vehicle speed
- **Acquire Factor**: Controls how aggressively the system acquires the guidance line

### 2. Performance Monitoring
The auto-tuner continuously collects and analyzes performance data:
- **Cross-track Error**: Distance from the desired guidance line
- **Steering Angle Variation**: Oscillation detection to identify unstable behavior
- **Overshoot Detection**: Identifies when the system overcorrects guidance errors
- **Stability Metrics**: Determines when the system is in a steady state suitable for tuning

### Hierarchical Adjustment Strategy
The system employs a hierarchical approach to parameter adjustment based on performance metrics:

1. **High Error Conditions** (>0.15m cross-track error):
   - Aggressive tuning to quickly reduce guidance errors
   - Larger parameter adjustments to rapidly improve performance

2. **Moderate Error Conditions** (0.08-0.15m cross-track error):
   - Balanced tuning approach
   - Moderate parameter adjustments to maintain stability while improving accuracy

3. **Good Tracking Conditions** (<0.03m cross-track error):
   - Fine-tuning mode for precision optimization
   - Small, conservative adjustments to maintain optimal performance

### Bidirectional Learning for Steer Response
The Steer Response parameter uses an intelligent bidirectional learning algorithm:

1. **Base Value**: Starts from a default value of 3.0 (representing optimal baseline performance)
2. **Directional Testing**: Tests adjustments in both directions (increase and decrease) to determine optimal setting
3. **Performance Comparison**: Evaluates system performance with each adjustment
4. **Self-Correction**: Automatically resets to base value when performance degrades significantly
5. **Bounds Enforcement**: Maintains parameter values within valid ranges (1.0-7.0)

### Speed-Adaptive Behavior
The system adapts steering parameters based on vehicle speed:
- **Low Speed**: Increases responsiveness for better maneuverability
- **High Speed**: Reduces aggressive steering to maintain stability and prevent oscillation

### Safety and Stability Measures
To ensure stable operation and prevent erratic behavior:

1. **Extended Timing Constraints**: Minimum 15-second intervals between adjustments for sensitive parameters
2. **Gentle Adjustment Steps**: Small incremental changes to prevent jarring parameter shifts
3. **Stability Verification**: Only performs adjustments when system is determined to be stable
4. **Bounds Checking**: All parameters are constrained within predefined minimum and maximum values
5. **Real-time UI Synchronization**: Updates user interface controls immediately when parameters change

### Core Components

1. **CAutoTuner Class**: Main implementation of the auto-tuning algorithm
2. **AutoTuneConfig Class**: Configuration structure for storing parameter bounds and current values
3. **Performance Monitoring**: Real-time collection and analysis of guidance performance metrics
4. **UI Integration**: Synchronization with FormSteer and FormAutoTuning interfaces

### Data Flow

1. **Initialization**: 
   - Load default parameter bounds and values
   - Initialize performance monitoring queues
   - Set up learning algorithm variables

2. **Runtime Operation**:
   - Collect performance data every frame
   - Calculate performance metrics from historical data
   - Determine if parameter adjustments are needed
   - Apply adjustments within safety constraints
   - Update UI with current parameter values

3. **Persistence**:
   - Save optimal configurations to JSON file
   - Restore previous settings on application restart

### Error Prevention

The implementation includes robust error handling to prevent common issues:

1. **Range Validation**: All slider controls are protected against ArgumentOutOfRangeException
2. **Null Checks**: Prevents null reference exceptions in UI initialization
3. **Bounds Enforcement**: Ensures all parameter values remain within valid ranges
4. **Safe Conversion**: Proper scaling between internal values and UI slider positions

## Benefits

1. **Improved Guidance Accuracy**: Automatically optimizes steering parameters for minimal cross-track error
2. **Adaptive Performance**: Adjusts to changing field conditions, vehicle speeds, and implement characteristics
3. **Reduced Manual Tuning**: Eliminates the need for operators to manually adjust complex steering parameters
4. **Enhanced Stability**: Prevents oscillation and overshoot through intelligent parameter management
5. **Consistent Performance**: Maintains optimal guidance quality across different operating conditions

## Conclusion

The auto-tuning feature represents a significant advancement in AgOpenGPS guidance capabilities. By implementing intelligent, adaptive parameter adjustment with comprehensive safety measures, the system provides optimal steering performance with minimal operator intervention. The bidirectional learning algorithm for Steer Response and hierarchical adjustment strategy ensure that the system can adapt to various field conditions while maintaining stability and preventing common tuning errors.

### DIFF
--- a/SourceCode/GPS/AgOpenGPS.csproj
+++ b/SourceCode/GPS/AgOpenGPS.csproj
@@ -26,6 +26,7 @@
     </Reference>
     <Reference Include="System.Configuration" />
     <Reference Include="System.Management" />
+    <Reference Include="System.Web.Extensions" />
     <Reference Include="System.Windows.Forms.DataVisualization" />
   </ItemGroup>
   <ItemGroup>
--- a/SourceCode/GPS/Classes/CAutoTuner.cs
+++ b/SourceCode/GPS/Classes/CAutoTuner.cs
@@ -1,0 +1,620 @@
+﻿﻿﻿﻿﻿﻿﻿﻿﻿﻿﻿﻿﻿﻿//Auto-Tuning Algorithm for AgOpenGPS Steering System
+//Automatically adjusts Pure Pursuit and Steering parameters based on performance
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Web.Script.Serialization;
+using System.Windows.Forms;
+using AgOpenGPS.Properties;
+
+namespace AgOpenGPS
+{
+    public class CAutoTuner
+    {
+        private readonly FormGPS mf;
+
+        // Auto-tuning settings
+        public bool IsAutoTuningEnabled { get; set; } = false;
+        public bool IsLearning { get; private set; } = false;
+        
+        // Performance monitoring
+        private readonly Queue<double> crossTrackErrorHistory = new Queue<double>();
+        private readonly Queue<double> steerAngleHistory = new Queue<double>();
+        private readonly Queue<double> speedHistory = new Queue<double>();
+        
+        // Performance metrics
+        private double currentCrossTrackError = 0.0;
+        private double averageCrossTrackError = 0.0;
+        private double oscillationMagnitude = 0.0;
+        private double overshootDetected = 0.0;
+        private int frameCounter = 0;
+        
+        // Auto-tune parameters bounds and current values
+        public AutoTuneConfig Config { get; private set; }
+        
+        // Learning algorithm variables
+        private int stableFrameCount = 0;
+        private double bestPerformanceScore = double.MaxValue;
+        private AutoTuneConfig bestConfig;
+        private DateTime lastAdjustmentTime = DateTime.Now;
+        private DateTime lastSteerResponseAdjustmentTime = DateTime.Now;
+        
+        // SteerResponse bidirectional learning variables
+        private double steerResponseBaseValue = 3.0; // Base value to start from
+        private double lastSteerResponseValue = 3.0;
+        private double lastSteerResponsePerformance = double.MaxValue;
+        private bool steerResponseIncreaseDirection = true; // Track which direction to try first
+        private int steerResponseAdjustmentCount = 0;
+        
+        // Constants
+        private const int HISTORY_SIZE = 100;
+        private const int STABLE_FRAMES_REQUIRED = 50;
+        private const double MIN_ADJUSTMENT_INTERVAL_SECONDS = 5.0;
+        private const double MIN_STEER_RESPONSE_ADJUSTMENT_INTERVAL_SECONDS = 15.0; // Extended timing for sensitive parameter
+        private const double PERFORMANCE_IMPROVEMENT_THRESHOLD = 0.02;
+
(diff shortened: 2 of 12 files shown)


lansalot:
I know a lad that tried this, it was not bad slow-speed, but 8kmh and he was all over the place. Looking at the code, I think it just tries to do too much at once. A better approach might be to concentrate on changing one thing perhaps, like minpwm for inital response or pgain for speed at large XTE etc and then move onto the next?

gunicsba:
I agree with lansalot here. For me it's always the wheel angle sensor calibration that needs to be done properly.

Suppose we have a was that can do +/- 30 degrees. If we start with WAS 0 being 3 degrees out and do the calibration then later redo the WAS 0 out cpd and Ackermann won't be proper.

I liked the way Topcon does this calibration:
Do a left / right turn measure the max counts value then starts to drive straight and tries to figure the middle value. Once we have these raw values I think the rest of it ( zero, cpd, Ackermann, max limit) could be calculated properly.

Integral we could fix at 10 imho. Max pwm could be let to the user (150 is quite a good value to start with)

As for the minimum and proportional settings we should be looking at the steer chart. If it starts to move late then needs more minimum pwm. I would start with a lower value just like the Wizard suggests and raise it until it starts to react well.

As for proportional it needs to come down if it overshoots then the values. If the 2 bar doesn't follow each other close enough then the proportional needs to be higher. 

A great help for this is when we move the guidance line 15-30cm to the side which mimics the line acquisition 5-10 cm errors. Based on my experience if we're able to find back to the line from 10-20 cm then the 0-5 cm aspect will be good as well.

Maybe this could be fed back to the AI tool?

Do a proper WAS calibration.
Start to drive on the line and jump +/- 20cm to see how steering reacts (3-5 km/H)
Then once these rough values are estimated we should ask the user to do the testing at the bottom/top of the speed range. Once we're happy with the way it behaves we should be able to have a good idea how the acquisition factor, speed gain , look ahead values look. 

Keep in mind that short look ahead might mean we're never able to get on the line as the steering might not be quick enough for it.

richardklasens:
I changed the Base branch to Develop, since that is the place to create PR’s on, but as you can see, it’s now far behind and needs a rebase first to solve merge conflicts

PR #858, opened 2025-09-03
Add BingMap for easy loading and saving field background pictures

J005t67:


richardklasens:
If it’s okay I will hold this until my PR is merged into develop with the new Loader/Saver classes. The PR is approved but I want to do a full roundtrip test first

J005t67:
> If it’s okay I will hold this until my PR is merged into develop with the new Loader/Saver classes. The PR is approved but I want to do a full roundtrip test first

I know you prefer not to have any interference when you are working on the Loader/Saver classes. That's the reason why this PR only contains code that is not yet used anywhere, so this PR can be completed without any risk.

My next PR, starts using the new BingMap and BingMapStreamer. It will replace the old code in WorldGrid and FormMap. So, I propose to approve this one, and wait with approving my next PR.

PR #877, opened 2025-09-17
Let new base class DisposableObject take care of Dispose pattern

J005t67:
Let new base class DisposableObject take care of Dispose pattern
Subclasses only need to override the method OnDispose to implement the Dispose pattern
and let the class DisposableObject takes care of the rest.

### DIFF
--- a/SourceCode/AgOpenGPS.Core/Base/DisposableObject.cs
+++ b/SourceCode/AgOpenGPS.Core/Base/DisposableObject.cs
@@ -1,0 +1,43 @@
+﻿using System;
+
+namespace AgOpenGPS.Core.Base
+{
+    // Base class that makes implementing the Dispose pattern easy.
+    public abstract class DisposableObject : IDisposable
+    {
+        protected bool IsDisposed { get; private set; }
+
+        protected DisposableObject()
+        {
+        }
+
+        ~DisposableObject()
+        {
+            Dispose(false);
+        }
+
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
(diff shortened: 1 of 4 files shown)


FantasyTeddy:
I don't think this is a good idea. The Dispose pattern is concise enough such that it can be implemented without the need of such a base class. Having another layer of abstraction around that just increases the chance that it won't be recognized by anyone new to the project.

J005t67:
> I don't think this is a good idea. The Dispose pattern is concise enough such that it can be implemented without the need of such a base class. Having another layer of abstraction around that just increases the chance that it won't be recognized by anyone new to the project.

I think many people are (a bit) puzzled by the Dispose pattern and the new class really helps to simplify things for them. Also why would you clutter up your classes with the same code for the Dispose pattern over and over again, when you only have to override one method?

FantasyTeddy:
Yes, the Dispose pattern (and `IDisposable` in general) are not that easy to grasp.
However, in its most basic form, implementing `IDisposable` should only add one method:
```csharp
public void Dispose()
{
    // Dispose managed resources
}
```
The other parts are very specific for what use-case needs to be fulfilled:
- Is this class supposed to be a base class? The add the `protected virtual void Dispose(bool disposing)` method.
- Do you need to dispose unmanaged resources? Then add a finalizer (although that is actually discouraged).

My point is just: disposing objects is not easy and making it seem easy can be dangerous...

richardklasens:
I’m not sure if I’m a fan of this…

richardklasens:
Lets close this PR since it seems it doesn’t make it better?

PR #878, opened 2025-09-17
Bugfix/restore mousewheel zoom

J005t67:
- Restore original code for zooming  with mouse wheel.
- Move zoom code to Camera class and make things a bit more robust.


FantasyTeddy:
@J005t67 and @richardklasens: I'm not sure if this was correct: you merged changes from the `develop` branch into the `Fixes-6.8` branch...

richardklasens:
The fix needs to be in the upcoming release

FantasyTeddy:
Yes, but not all the other changes that happened on `develop` since we branched off...

Issue #871, opened 2025-09-14
Roll icon too small


gunicsba:
Not sure what happened but the bottom right roll marker used to be bigger.

Is this intentional?

richardklasens:
I haven’t noticed it….

gunicsba:
Old:
<img width="960" height="1280" alt="Image" src="https://github.com/user-attachments/assets/ef07679b-5749-4ddf-b1a8-a5928d845558" />
New:

<img width="960" height="1280" alt="Image" src="https://github.com/user-attachments/assets/127ce021-34eb-451c-904b-dea9cd67390f" />

J005t67:
Most likely, I did it unintentionally during the introducing of the Texture classes. I assigned this issue to myself.

richardklasens:
Solved in PR #879

PR #879, opened 2025-09-17
Bugfix/rollindicator too small

J005t67:
The RollIndicator became half its size on 2025-02-17 after introduction of new Texture functions. Corrected now.


richardklasens:
This wil fix #871

PR #884, opened 2025-09-19
Adds overwrite confirmation for profile creation

richardklasens:
Prompts the user for confirmation before overwriting an existing profile.
Adds an extra confirmation for resetting to an empty profile, warning the user about potential data loss.
Handles empty profile names gracefully, preventing errors.

### DIFF
--- a/SourceCode/GPS/Forms/Profiles/FormNewProfile.cs
+++ b/SourceCode/GPS/Forms/Profiles/FormNewProfile.cs
@@ -81,41 +81,51 @@ namespace AgOpenGPS.Forms.Profiles
         private void buttonOK_Click(object sender, EventArgs e)
         {
             string newProfileName = SanitizeFileName(textBoxName.Text.Trim()).Trim();
-            if (!string.IsNullOrEmpty(newProfileName))
+            if (string.IsNullOrEmpty(newProfileName))
+                return;
+
+            string newProfilePath = Path.Combine(RegistrySettings.vehiclesDirectory, newProfileName + ".xml");
+
+            // Bestaat al? -> bevestigen om te overschrijven
+            if (File.Exists(newProfilePath))
             {
-                string newProfilePath = Path.Combine(RegistrySettings.vehiclesDirectory, newProfileName + ".xml");
+                var overwrite = FormDialog.Show(
+                    gStr.gsSaveAndReturn,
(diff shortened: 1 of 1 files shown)


SourceCode/GPS/Forms/Profiles/FormNewProfile.cs
```diff
@@ -81,42 +81,52 @@ private void textBoxName_TextChanged(object sender, EventArgs e)
...
-                        $"Profile '{newProfileName}' already exists. Overwrite?",
-                        MessageBoxButtons.YesNo);
-
-                    if (result != DialogResult.OK)
-                    {
-                        return;
-                    }
-                }
+                if (overwrite != DialogResult.Yes)
+                    return;
```
FantasyTeddy: Shouldn't that be `DialogResult.OK`? Otherwise I think it will always return here.
```suggestion
                if (overwrite != DialogResult.OK)
                    return;
```

SourceCode/GPS/Forms/Profiles/FormNewProfile.cs
```diff
@@ -81,42 +81,52 @@ private void textBoxName_TextChanged(object sender, EventArgs e)
...
-                {
-                    CreateNewProfileFromCurrent(newProfileName);
-                }
-                else
-                {
-                    CreateNewProfileFromExisting(newProfileName, existingProfileName);
-                }
+            if (existingProfileName.Equals(EmptyProfile))
+            {
+                // EXTRA BEVESTIGING voor 'Empty Profile'
```
FantasyTeddy: Dutch comment

SourceCode/GPS/Forms/Profiles/FormNewProfile.cs
```diff
@@ -81,42 +81,52 @@ private void textBoxName_TextChanged(object sender, EventArgs e)
...
+                var confirmReset = FormDialog.Show(
+                    "!! WARNING !!",
+                    "This will reset all Tractor measurements and control, Are you Sure??",
+                    MessageBoxButtons.YesNo);
+
+                if (confirmReset != DialogResult.Yes)
```
FantasyTeddy: Again, I think this should be `DialogResult.OK`.

FantasyTeddy (approved):

PR #885, opened 2025-09-20
Replace `Accord.Video.DirectShow` with `OpenCvSharp4`

FantasyTeddy:
The changes in #876 were still not enough to make the webcam work with .NET 9.

This is another attempt to use `OpenCvSharp4` instead.


FantasyTeddy:
Oh no, I just realized that this increases the size of the application by about 10x...
That still might not be a suitable solution :(

richardklasens (commented):
So keep this open to probably find a better (another) solution?

FantasyTeddy:
> So keep this open to probably find a better (another) solution?

Yes, I think so.

PR #891, opened 2025-09-23
Add Math Helpers and SensorInputMapper

richardklasens:


SourceCode/AgOpenGPS.Core/Math/Distance2D.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+    /// <summary>
+    /// Helpers to compute Euclidean distances in the local plane (meters).
+    /// </summary>
+    public static class Distance2D
+    {
+        /// <summary>
+        /// Euclidean distance between two (easting, northing) points (meters).
+        /// </summary>
+        public static double Distance(double easting1, double northing1, double easting2, double northing2)
```
J005t67, line 17: General comment: Please do not add another Distance method. We already have so many.
Specific comment:
We should avoid code that needs to calculate the easting and northing part of coordinates in separate statements.
GeoCoord, GeoDelta and GeoDir have oparators that allow to make calculations without bothering about the individual parts. As a conseqeunce, we do not need this method and we should not want it.

SourceCode/AgOpenGPS.Core/Math/Distance2D.cs
```diff
@@ -0,0 +1,46 @@
...
+            double dx = easting1 - easting2;
+            double dy = northing1 - northing2;
+            return dx * dx + dy * dy;
+        }
+
+        /// <summary>
+        /// Euclidean distance between two GeoCoord points (meters).
+        /// Delegates to GeoCoord.Distance to avoid duplicated logic.
+        /// </summary>
+        public static double Distance(GeoCoord a, GeoCoord b) => a.Distance(b);
```
J005t67, line 38: In my opinion, we don't need this. Instead of calling this small wrapper function, it is better to call the Distance method of GeoCoord directly.

Issue #571, opened 2025-01-11
Remove unused code/resources


FantasyTeddy:
Because AgOpenGPS has a long history and has seen many changes and additions, there are quite a few things that are not used anymore.
In order to keep the codebase clean and maintainable, unused code or resources should be removed.

- [ ] ~Remove commented out code~ (see comments below)
- [x] Remove code that is never called (dead code)
- [x] Remove forms that are never displayed
- [x] Remove resources (images, sounds, ...) that are not used anymore
- [x] Remove settings that are not used anymore
- [x] Remove texts/translations that are not used anymore
- [x] Remove tools that are not used anymore (e.g. OGL.exe, UDP_SimTool.exe)

J005t67:
Commented out code sometimes contains usefull comments. I would keep the commented out code.

FantasyTeddy:
Can you give an example of where this is the case?
Because in my experience, commented out code has almost always been something that someone tried and then forgot about.

J005t67:
I agree that most of the commented out has no value anymore, but I propose to keep it in for now. And I don't want to dig in the code for hours to see which comments are still valuable and which ones are  not. I propose to remove commented out code l ater when somebody is working on a file anyway.

FantasyTeddy:
Okay, but this means that we also cannot do any of the other proposed cleanup steps: anything could be referenced from within commented out code and if we remove it, the commented out code might become useless...

Henny20:
To be honest. Images in Resources will not work on other platforms. There is no System.Drawing. That is only supported on Windows. We load images in C# or Xaml code.

richardklasens:
I filled the checkboxes :)

FantasyTeddy:
> I filled the checkboxes :)

What do the checked boxes represent for you? The things we should clean up?
Can you explain your reasoning for why these should be done and the others not?

richardklasens:
> What do the checked boxes represent for you? The things we should clean up? Can you explain your reasoning for why these should be done and the others not?

I think, based on my second reading, everything is ok but I’m with Joost, for understanding other parts it’s not a problem to leave commented code for now.

FantasyTeddy:
Okay, thank you for clarifying.
I guess we just need to be careful to not delete anything that is referenced from any commented out code.

FantasyTeddy:
I updated the description of this issue accordingly and removed the checkmarks again, such that they can be checked once they are actually done.

PR #937, opened 2025-10-06
Remove unused commented out code blocks

Xyntexx:
Remove 700+ lines of commented-out code. All this code is preserved 
in git history if needed.

Files cleaned:
- SourceCode\GPS\Classes\CContour.cs
- SourceCode\AgIO\Source\Forms\SerialComm.Designer.cs
- SourceCode\GPS\Classes\CABCurve.cs

No functional changes.

### DIFF
--- a/SourceCode/AgIO/Source/Forms/SerialComm.Designer.cs
+++ b/SourceCode/AgIO/Source/Forms/SerialComm.Designer.cs
@@ -32,16 +32,12 @@ namespace AgIO
         public static string portNameMachineModule = "***";
         public static int baudRateMachineModule = 38400;
 
-        //public  static string portNameModule3 = "***";
-        //public  static int baudRateModule3 = 38400;
-
         //used to decide to autoconnect section arduino this run
         public string recvGPSSentence = "GPS";
         public string recvGPS2Sentence = "GPS2";
         public string recvIMUSentence = "IMU";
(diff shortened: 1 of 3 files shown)


FantasyTeddy:
We had a discussion about removing commented-out code in issue #571 a while back.

@J005t67: Maybe you can elaborate if this is still valid or we should actually remove the "dead code"?

Xyntexx:
IMO, when these classes will be modified or refactored the approach will be copied from AOG_Dev and the comments won't have any value. 

And then there are the drawbacks of leaving commented-out code.

Xyntexx:
Also in some cases the commented out code is older or newer than the surrounding code so it's hard to know for someone to know if it's still relevant or what its purpose

FantasyTeddy:
I am totally with you, but I think @J005t67 or @richardklasens should decide if this is the right thing to do.

richardklasens (approved):

J005t67:
> We had a discussion about removing commented-out code in issue #571 a while back.
> 
> @J005t67: Maybe you can elaborate if this is still valid or we should actually remove the "dead code"?

In general it is my personal preference to keep the commented out code a bit longer, but I can also see the advantages of removing it. If someone knows the code well and he knows that the commented code is outdated, it is defenitvely better to remove it.

Issue #916, opened 2025-10-05
Tram Width imperial unit display


WildBuckwheat:
v6.3.0b3

<img width="768" height="581" alt="Image" src="https://github.com/user-attachments/assets/250069b6-b386-4a0c-9d25-19a50b441d53" />

- Tram width not displayed correctly when units set to imperial (12 inches in a foot, display should be 90')
- Historically this has been an issue with implement widths through AgOpenGPS also. It appears to be fixed for implement widths?

PR #944, opened 2025-10-07
Bugfix/imperial tramwidth

J005t67:
Fix bug #916: Correctly display imperial tram width.

First round to inches before splitting feets and inches. (Splitting first and round later results in things like 89'12")

Fix same problem in display of Tool Width at bottom row of ConfigTool


richardklasens (approved):
Code wise it seems promising! With this you can probably make a deep dive in the whole program :)

Issue #893, opened 2025-09-24
System.AccessViolationException on close


lansalot:
Develop branch at fcc148a99a25bbe78194c419bfa4999206fecaeb

Open app, opened configuration, closed, time for a beer, goodbye.. exception thrown. Value of _textureId is 13 at time of death.

<img width="1149" height="287" alt="Image" src="https://github.com/user-attachments/assets/9a4fff66-1413-465a-9afa-3a63e0140e29" />

Call stack:

<img width="892" height="106" alt="Image" src="https://github.com/user-attachments/assets/2ef92d15-fe6b-4402-8d06-5483283bcd0f" />

https://github.com/AgOpenGPS-Official/AgOpenGPS/blob/fcc148a99a25bbe78194c419bfa4999206fecaeb/SourceCode/AgOpenGPS.Core/DrawLib/Texture2D.cs#L94

System.AccessViolationException
  HResult=0x80004003
  Message=Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
  Source=<Cannot evaluate the exception source>
  StackTrace:
<Cannot evaluate the exception stack trace>

richardklasens:
@J005t67 it seems it tries to delete texture if the program is already passed closing it

Xyntexx:
Fixed in #947

PR #947, opened 2025-10-07
Fix AccessViolationException when closing application

Xyntexx:
- Prevents AccessViolationException crashes when closing the application by only calling OpenGL texture deletion during explicit disposal, not during finalization
- OpenGL calls require an active context and must be on the correct thread, which cannot be guaranteed during garbage collection finalization


Xyntexx:
Fixes #893

richardklasens (approved):

Xyntexx:
@FantasyTeddy Let's add this to the milestone?

FantasyTeddy:
Doesn't this problem only exist in "develop"?

Xyntexx:
You're correct, never mind

Issue #926, opened 2025-10-06
U turn crashes program near inner boundaries


WildBuckwheat:
v6.8.0b3

https://github.com/user-attachments/assets/296d3c6c-5a5f-48f5-971b-257cbb91d02f

- U turn generation near inner boundaries can result AgOpenGPS to crash

PR #949, opened 2025-10-07
Fix critical U-turn crashes near inner boundaries

Xyntexx:
Fixed two critical bugs that caused crashes when generating U-turns near inner boundaries:

1. Infinite loop bug: Four instances of incorrect loop iteration where 'cnt2--' was used instead of 'a++', causing infinite loops and array index out of bounds exceptions when the counter went negative.

2. Missing bounds validation: Added checks to ensure ytList and ytList2 are not empty before accessing array elements, preventing crashes when turn generation fails to produce valid paths.

These fixes prevent the application from freezing or crashing when U-turns are generated near small inner boundaries.

Testing help needed: These changes have not been extensively tested for U-turn performance and behavior. Please test U-turn generation in various scenarios, especially near inner boundaries, to ensure correct functionality.

Addresses #926


richardklasens:
Same here, please use fixes 6.8 as base branch

richardklasens (approved):
Lets merge this to fixes to we can test this also with other stuff

PR #954, opened 2025-10-07
Fix U-turn failing when offset equals half implement width

### LINKED ISSUE #935: U turn fails when offset is half implement width
v6.8.0b3

https://github.com/user-attachments/assets/45f8e5f5-f883-49b0-b7c3-2e9ece3e3904

- U turn fails when offset is set to **exactly** half of implement width

Xyntexx:
When tool offset is set to exactly half the implement width, the U-turn start and end positions become identical (turnOffset = 0). This caused two issues:

1. Path disappearing: The path would disappear immediately when the tractor reached the start position because the code couldn't distinguish between being at the start versus completing the turn.

2. Wrong direction after turn: When the straight approach and straight exit overlap at the same position, the closest point search would find the approach points instead of the exit points, causing the tractor to turn the wrong direction after completing the turn.

Fixes:
- Track progress through the U-turn path using maxProgressIndexReached and only allow completion when the vehicle has progressed through at least 50% of the path points
- Start closest point search from maxProgressIndexReached - 5 instead of from index 0, preventing the search from jumping back to overlapping start points when the vehicle should be following exit points

Fixes #935

### DIFF
--- a/SourceCode/GPS/Classes/CYouTurn.cs
+++ b/SourceCode/GPS/Classes/CYouTurn.cs
@@ -2452,6 +2452,7 @@ namespace AgOpenGPS
             mf.p_239.pgn[mf.p_239.uturn] = 0;
             isOutSameCurve = false;
             isGoingStraightThrough = false;
+            maxProgressIndexReached = 0;
         }
 
         public void FailCreate()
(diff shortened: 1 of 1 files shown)


farmerbriantee:
that doesn't actually fix the dubins problem which happens in multiple places in the application

Xyntexx:
What is the Dubins problem?

richardklasens (approved):
Until we have a better solution, let’s try rhis

Issue #906, opened 2025-10-05
Delete field path not displayed correctly


WildBuckwheat:
When deleting a field the user is prompted with a "Delete For Sure?" message

<img width="1790" height="1017" alt="Image" src="https://github.com/user-attachments/assets/599137d3-472b-4843-afbc-5f5d6b537560" />

Issue 1.
The message box attempts to display the entire path to the field files, but is unable to do so. The path name could be much longer depending on field name and local folder names. Perhaps only the field name should be displayed?
 
I'm nitpicking but
Issue 2.
"Delete For Sure?"
- uses mid sentence capitalization that is inconsistent with the rest of the program.
- Standard OS type wording would be something like "Are you sure you want to delete?" or "Are you sure you want to delete this field?"

farmerbriantee:
Is fixed

PR #965, opened 2025-10-08
Field path must fit in 'Delete For Sure' Dialog

J005t67:
When deleting a field, a confirmation dialog is shown with the path of the field.
This change splits the path of the field in two lines if it does not fit.
It solves the first part of issue # 906, but it does not change the texts.

### DIFF
--- a/SourceCode/GPS/Forms/Pickers/FormFilePicker.cs
+++ b/SourceCode/GPS/Forms/Pickers/FormFilePicker.cs
@@ -333,9 +333,19 @@ namespace AgOpenGPS
                 return;
 
             string dir2Delete = Path.Combine(RegistrySettings.fieldsDirectory, fieldName);
-
+            int maxLength = 45;
+            string multiLineDir = dir2Delete;
+            // Split directory name if it is too long for 1 line
+            if (maxLength < dir2Delete.Length)
+            {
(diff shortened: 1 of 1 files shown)


FantasyTeddy:
Wouldn't it make more sense to do it as suggested by @WildBuckwheat in #906: only display the field name instead of the entire path?

J005t67:
> Wouldn't it make more sense to do it as suggested by @WildBuckwheat in #906: only display the field name instead of the entire path?

Personally, I like the explicit way it displays the complete path, but I'm not a very typical user nor an usability expert. Which user or usability expert has an opinion on this?

WildBuckwheat:
Personally I prefer the complete path as well. If there’s a way to display the complete path that would be my preference.

FantasyTeddy:
Maybe we could do the following:
- Use the question as the title instead of the path.
- Change the message from a `Label` to a `TextBox` with `WordWrap` enabled.

That way we don't have to write our own wrapping logic. (However, I'm not sure if the automatic word wrapping works with file paths...)

richardklasens:
> Personally I prefer the complete path as well. If there’s a way to display the complete path that would be my preference.

I would definitely not use a path here. The field name is enough.. you picked it from the list and ticked Delete.. the complete path is the same for every field. So it doesn’t make sense to me

PR #970, opened 2025-10-09
Avoid saving and showing duplicated flags

Rob2029:
If a new flag is less than around 1mm next to an existing flag it will not be saved in flags.txt or shown on screen. To avoid loading duplicates from a file or by pressing the flag icon multiple times.


SourceCode/GPS/IO/FlagFiles.cs
```diff
@@ -47,18 +74,31 @@ public static List<CFlag> Load(string fieldDirectory)
 
         public static void Save(string fieldDirectory, IReadOnlyList<CFlag> flags)
         {
+
             var filename = Path.Combine(fieldDirectory, "Flags.txt");
 
+            // Prevent saving duplicates based on latitude and longitude
+            var distinctFlags = new List<CFlag>();
+            foreach (var f in flags ?? new List<CFlag>())
```
J005t67: Why iterate over an empty flag list?   You intentions are much more clear if you simply put the whole foreach inside an 
if (flags != null)

SourceCode/GPS/IO/FlagFiles.cs
```diff
@@ -1,20 +1,39 @@
...
+            }
+            return distinctFlags;
+        }
         public static List<CFlag> Load(string fieldDirectory)
         {
             var result = new List<CFlag>();
             var path = Path.Combine(fieldDirectory, "Flags.txt");
             if (!File.Exists(path)) return result;
 
+            var seen = new HashSet<string>();
```
J005t67: Is this used anywhere?

SourceCode/GPS/IO/FlagFiles.cs
```diff
@@ -1,20 +1,39 @@
...
+                    Math.Abs(d.longitude - f.longitude) < 1e-8);
+
+                if (!duplicate)
+                {
+                    distinctFlags.Add(f);
+                }
+            }
+            return distinctFlags;
+        }
         public static List<CFlag> Load(string fieldDirectory)
```
J005t67, line 30: Add empty line between functions

PR #972, opened 2025-10-09
Fix headland distance display

### LINKED ISSUE #956: Distance to Headland
v6.8.0b3
- Distance to headland in imperial units is inches, in metric is meters
- Inches not really an appropriate unit for such a measurement, should be feet (display as ft)
- Number display not vertically centered in yellow bubble
- A neat thing to display also might be area to headland?

Image

richardklasens:
Refines the headland distance display for better readability.

Adjusts spacing and formatting for clarity, and switches to feet for imperial units.
Fixes #956


SourceCode/GPS/Forms/OpenGL.Designer.cs
```diff
@@ -2647,19 +2648,20 @@ private void DrawHeadlandDistance()
...
 
-                // m -> inch = 39.3700787
-                double displayValue = imperial ? (meters * 39.3700787) : meters;
+                // m -> ft = 3.2808399
+                double displayValue = imperial ? (meters * 3.2808399) : meters;
 
-                // Formatting: keep one decimal for meters, no decimals for inches
-                string fmt = imperial ? "0" : "0.0";
-                string unit = imperial ? " IN" : " m";
+                // Formatting: two decimals for feet, one decimal for meters
```
J005t67: Strange, a meter is more than 3 times larger than a foot, so you might need one decimal more for meters for the same accuracy, not less.

SourceCode/GPS/Forms/OpenGL.Designer.cs
```diff
@@ -2647,19 +2648,20 @@ private void DrawHeadlandDistance()
...
+                string fmt = imperial ? "0.00" : "0.0";
+                string unit = imperial ? " ft" : " m";
 
-                label = displayValue.ToString(fmt) + unit;
+                label = displayValue.ToString(fmt, CultureInfo.InvariantCulture) + unit;
```
J005t67: CultureInfo.InvariantCulture is not correct. It will always use a '.' as decimal separator, which is incorrect in languages like Dutch that require a ','

SourceCode/GPS/Forms/OpenGL.Designer.cs
```diff
@@ -2647,19 +2648,20 @@ private void DrawHeadlandDistance()
...
-                // m -> inch = 39.3700787
-                double displayValue = imperial ? (meters * 39.3700787) : meters;
+                // m -> ft = 3.2808399
+                double displayValue = imperial ? (meters * 3.2808399) : meters;
 
-                // Formatting: keep one decimal for meters, no decimals for inches
-                string fmt = imperial ? "0" : "0.0";
-                string unit = imperial ? " IN" : " m";
+                // Formatting: two decimals for feet, one decimal for meters
+                string fmt = imperial ? "0.00" : "0.0";
```
J005t67: As far as I know, the preferred format is feet and inches, not feet with a decimal part.

SourceCode/AgOpenGPS.Core/Models/Base/Units.cs
```diff
@@ -84,7 +84,20 @@ public static string MediumDistanceString(bool isMetric, double distanceInMeters
                 return feet.ToString() + "' " + inches.ToString() + '"';
             }
         }
+        public static string ScalarDistanceString(bool isImperial, double distanceInMeters,
```
J005t67: Add an empty line between functions

J005t67: Keep things a bit consistent. All other functions use a boolean argument 'isMetric', so better avoid 'isImperial' in the new function.

J005t67: The 'Scalar' part in the function name does not really help. Distances are always a scalar. I would prefer something like MediumBigDistanceString. (And reserve the name LongDistanceString for display of values in km and miles.)

Issue #931, opened 2025-10-06
Some variables in All Settings not populating


WildBuckwheat:
v6.8.0b3

<img width="330" height="312" alt="Image" src="https://github.com/user-attachments/assets/72ba1892-8c20-4cff-b3ce-67cea55b8f96" />

<img width="444" height="356" alt="Image" src="https://github.com/user-attachments/assets/668b823d-f0b7-4acd-9084-ca86139318b4" />

<img width="274" height="345" alt="Image" src="https://github.com/user-attachments/assets/4392566c-08ec-4907-892a-0bbb5f4febca" />

Issue #909, opened 2025-10-05
Reference to AOG in launch window


WildBuckwheat:
v6.8.0 beta 3

<img width="773" height="516" alt="Image" src="https://github.com/user-attachments/assets/0b607e21-7198-44a5-bfa3-25f31ee9cfd6" />

To quote someone, "AOG != AgOpenGPS"

FantasyTeddy:
I do not agree with that statement. This is only one of many places, where "AOG" is used as an abbreviation for "AgOpenGPS".

As long as the "other AOG" is not something official, I don't think we should confuse people with statements like "AOG != AgOpenGPS".

farmerbriantee:
It still is best not to confuse the user by being schizophrenic.

FantasyTeddy:
What do you mean by schizophrenic?

I think it is totally okay to abbreviate a product.

Examples:
- WordPress - WP
- Photoshop - PS
- PowerPoint - PPT

farmerbriantee:
<img width="1067" height="729" alt="Image" src="https://github.com/user-attachments/assets/466a6154-b93e-4dc1-bf85-124fe996f38b" />

Issue #908, opened 2025-10-05
Hamburger -> About ->Enable Terms Form


WildBuckwheat:
6.8.3 Beta 3

<img width="768" height="519" alt="Image" src="https://github.com/user-attachments/assets/e76b134e-0894-42d6-a4b4-d7f9176d6097" />

The enable terms form button doesn't seem to do anything. 
Display for version not suited for longer version names.

FantasyTeddy:
I am not sure about the legal implications of hiding the terms form.
@farmerbriantee do you know if it's okay to allow users to hide that form?

lansalot:
Prefer to leave it there, it used to be hideable, but never hurts to remind
people.

richardklasens:
So.. delete the button means solved?

PR #969, opened 2025-10-09
Consolidate duplicate forms into AgLibrary

Xyntexx:
Moves FormYes and FormTimedMessage from GPS/AgIO/ModSim into shared AgLibrary.

**Changes:**
- Created AgLibrary/Forms/ with unified implementations
- FormYes: Combined dynamic width calc + optional Cancel button
- FormTimedMessage: Clean implementation with proper disposal
- Removed image dependencies (resource-agnostic)
- Added AgLibrary reference to ModSim project

**Impact:**
- 20 files changed: 23 insertions(+), 603 deletions(-)
- Net reduction of 580 lines of code
- Single source of truth for dialog forms

**Test plan:**
- [ ] Build solution successfully
- [ ] Test FormYes in GPS/AgIO/ModSim applications
- [ ] Test FormTimedMessage in GPS/AgIO/ModSim applications
- [ ] Verify Cancel button works in FormYes when enabled
- [ ] Verify timer auto-close works in FormTimedMessage

PR #986, opened 2025-10-13
Small fixes for FormTimedMessage,FormDialog,FormFirst,FormAllSetting

### LINKED ISSUE #953: Profile messages don't fit
Profile warning message texts do not fit boxes.

Image

Image

richardklasens:
Fixes multiple issues
Closes #953 #931 #908 #909

### DIFF
--- a/SourceCode/GPS/Forms/FormTimedMessage.cs
+++ b/SourceCode/GPS/Forms/FormTimedMessage.cs
@@ -12,16 +12,26 @@ namespace AgOpenGPS
         {
             InitializeComponent();
 
-            //get copy of the calling main form
-            //mf = callingForm as FormGPS;
-
+            // Set text first
             lblMessage.Text = titleString;
             lblMessage2.Text = messageString;
 
-            timer1.Interval = timeInMsec;
+            // Force labels to measure their content
+            lblMessage.AutoSize = true;
+            lblMessage2.AutoSize = true;
 
-            int messWidth = messageString.Length;
-            Width = messWidth * 15 + 120;
+            // Calculate required size based on actual label sizes
+            int maxLabelWidth = Math.Max(lblMessage.PreferredWidth, lblMessage2.PreferredWidth);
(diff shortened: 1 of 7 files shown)


SourceCode/GPS/Forms/FormTimedMessage.cs
FantasyTeddy, line 1: Just as a heads-up: this will probably cause a conflict with #969 when we merge the release back into develop.

SourceCode/GPS/Forms/Settings/FormAllSettings.cs
```diff
@@ -94,9 +94,9 @@ private void LoadLabels()
             label126.Text = Properties.Settings.Default.setTool_trailingToolToPivotLength.ToString();
             label128.Text = Properties.Settings.Default.setVehicle_hitchLength.ToString();
             label130.Text = Properties.Settings.Default.setVehicle_hydraulicLiftLookAhead.ToString();
-            label132.Text = "*****";
+            label132.Text = "N/A";
             label134.Text = Properties.Settings.Default.setVehicle_isStanleyUsed.ToString();
-            label136.Text = "*****";
+            label136.Text = "N/A";
```
FantasyTeddy, line 99: Do we know what these settings should be set to? If we don't shouldn't we just remove them?

richardklasens: they can be removed but then you need to do a big overhaul on the page.

SourceCode/GPS/Forms/FormTimedMessage.cs
richardklasens: We will see :)

SourceCode/GPS/Forms/Settings/FormAllSettings.cs
FantasyTeddy: Okay, so maybe something to do in develop then?

PR #987, opened 2025-10-14
Add AgShare Bulk Uploader 

richardklasens:
I push this to fixes to have the opportunity to test it with more people


SourceCode/GPS/Forms/Field/FormAgShareUploader.cs
```diff
@@ -0,0 +1,323 @@
...
+using System.Windows.Forms;
+using AgLibrary.Logging;
+using AgOpenGPS.Core.Models;
+using AgOpenGPS.IO;
+
+namespace AgOpenGPS.Forms
+{
+    public partial class FormAgShareUploader : Form
+    {
+        private readonly AgShareClient agShareClient;
```
FantasyTeddy, line 16: It is rather unfortunate, that we cannot use `CAgShareUploader` here instead of the `AgShareClient` directly. (For downloading the `CAgShareDownloader` is not `static` and encapsulates everything.)

SourceCode/GPS/Forms/Field/FormAgShareUploader.cs
```diff
@@ -0,0 +1,323 @@
...
+        private List<FieldInfo> availableFields;
```
FantasyTeddy, line 17: I don't think we need this field. It is only used inside of `LoadAvailableFields` and only for counting the number of fields.

SourceCode/GPS/Forms/Field/FormAgShareUploader.cs
```diff
@@ -0,0 +1,323 @@
...
+            if (snapshot == null)
+            {
+                throw new Exception("Failed to load field data");
+            }
+
+            // Use existing upload logic
+            await CAgShareUploader.UploadAsync(snapshot, agShareClient, null);
+        }
+
+        private async Task<FieldSnapshot> LoadFieldSnapshot(FieldInfo fieldInfo)
```
FantasyTeddy, line 242: I think this method should be part of the `CAgShareUploader`.

SourceCode/GPS/Forms/Field/FormAgShareUploader.cs
```diff
@@ -0,0 +1,323 @@
...
+            // Load field data from directory
+            var snapshot = await LoadFieldSnapshot(fieldInfo);
+
+            if (snapshot == null)
+            {
+                throw new Exception("Failed to load field data");
+            }
+
+            // Use existing upload logic
+            await CAgShareUploader.UploadAsync(snapshot, agShareClient, null);
```
FantasyTeddy, line 239: Passing `null` as the parameter for `FormGPS` only works, because it is not actually used within the method. Can we remove it?

SourceCode/GPS/Forms/Field/FormAgShareUploader.cs
```diff
@@ -0,0 +1,323 @@
...
+                    catch (Exception ex)
+                    {
+                        failCount++;
+                        lblStatus.Text = $"Failed: {fieldInfo.Name} ✗ - {ex.Message}";
+                        Log.EventWriter($"Failed to upload field {fieldInfo.Name}: {ex.Message}");
+                    }
+
+                    progressBar.Value++;
+                    Application.DoEvents(); // Keep UI responsive
+                    await Task.Delay(500); // Small delay to show status
```
FantasyTeddy, line 208: I'm not sure it is a good idea to wait here. Imagine that you want to upload 20 fields... now you have to wait 10 seconds in total...

SourceCode/GPS/Forms/Field/FormAgShareUploader.cs
richardklasens: Not important for now. The CAgShareUploader is unchanged.

SourceCode/GPS/Forms/Field/FormAgShareUploader.cs
richardklasens: Will do in develop after merge

SourceCode/GPS/Forms/Field/FormAgShareUploader.cs
richardklasens: Will do in develop after merge

SourceCode/GPS/Forms/Field/FormAgShareUploader.cs
richardklasens: Idem

PR #1008, opened 2025-10-21
Improves application shutdown process

richardklasens:
Improves the application shutdown process by addressing data loss and race conditions, ensuring settings and field data are saved correctly.

- Adds logging for application closing.
- Implements improved exception handling during the saving process.
- Fixes potential race conditions by using a shutdown flag.
- Ensures settings are saved and the system log is written even when no job is started.

Fixes: Critical Shutdown Data Loss & Race Conditions

### DIFF
--- a/SourceCode/GPS/Forms/OpenGL.Designer.cs
+++ b/SourceCode/GPS/Forms/OpenGL.Designer.cs
@@ -1393,7 +1393,10 @@ namespace AgOpenGPS
             if (fileSaveAlwaysCounter > 60)
             {
                 fileSaveAlwaysCounter = 0;
-                //if (sbMissedSentence.Length > 0) FileSaveMissedEvents();
+                if (Log.sbEvents.Length > 0)
+                {
+                    Log.FileSaveSystemEvents();
+                }
             }
 
             //if a minute has elapsed save the field in case of crash and to be able to resume            


SourceCode/GPS/Forms/FormGPS.cs
```diff
@@ -564,77 +564,130 @@ private void FormGPS_FormClosing(object sender, FormClosingEventArgs e)
...
 
 
         private async Task ShowSavingFormAndShutdown(int choice)
         {
-            using (FormSaving savingForm = new FormSaving())
+            FormSaving savingForm = null;
+
+            try
             {
+                savingForm = new FormSaving();
```
FantasyTeddy, line 611: Why do it this way and not with the `using` statement as before?

richardklasens: The issue with `using`:
  When the user clicks "No" on the save error dialog (line 670-674), we need to cancel shutdown and return early.
  With a using statement, this would cause:
  1. Line 673: savingForm.Close() starts the dispose process
  2. return exits the using scope
  3. Compiler-generated Dispose() is called again → double dispose risk

  This can lead to ObjectDisposedException or race conditions in Windows Forms event handling.

  Why try/finally is better here:
  - The finally block always executes, even with early returns
  - We have explicit control over disposal timing
  - ?.Dispose() is safe and won't conflict with previous Close() calls
  - The form stays visible during exception handling for user feedback

  For complex async operations with multiple exit points and user interaction, explicit try/finally is more
  predictable than using.

PR #1016, opened 2025-10-23
Improves AgShare field download process

richardklasens:
Adds validation and error handling to the AgShare field download process to ensure data integrity and provide better feedback to the user.

Includes checks for null or invalid field data, coordinate ranges, and boundary/AB line availability.  Also provides more detailed logging and handles potential exceptions during download and parsing.

Adds a "Failed" counter to the bulk download process and displays it to the user.

### DIFF
--- a/SourceCode/GPS/Classes/AgShare/AgShareDownloader.cs
+++ b/SourceCode/GPS/Classes/AgShare/AgShareDownloader.cs
@@ -90,17 +90,9 @@ namespace AgOpenGPS
                     return null;
                 }
 
-                // Validation is now done in Parse method, but we'll do a quick check here
-                // since Parse throws exceptions and we want to return null for preview failures
-                try
-                {
-                    AgShareFieldParser.Parse(dto);
-                }
-                catch (ArgumentException ex)
-                {
-                    Log.EventWriter($"[AgShare] Preview download failed for fieldId={fieldId}: {ex.Message}");
-                    return null;
-                }
+                // Validation is done in Parse method
+                // If validation fails, the outer catch will handle it
+                AgShareFieldParser.Parse(dto);
 
                 return dto;
             }


SourceCode/GPS/Classes/AgShare/AgShareDownloader.cs
```diff
@@ -25,14 +25,85 @@ public CAgShareDownloader()
...
                 var dto = JsonConvert.DeserializeObject<AgShareFieldDto>(json);
+
+                // Validate DTO
+                if (dto == null)
+                {
+                    Log.EventWriter($"[AgShare] Download failed for fieldId={fieldId}: Failed to deserialize field data");
+                    return false;
+                }
+
+                if (!ValidateFieldDto(dto))
```
FantasyTeddy: Maybe the `ValidateFieldDto` could be part of the `AgShareFieldParser.Parse` method? That way you wouldn't have to do the same checks twice.

SourceCode/GPS/Classes/AgShare/AgShareDownloader.cs
```diff
@@ -54,51 +125,123 @@ public async Task<List<AgShareGetOwnFieldDto>> GetOwnFieldsAsync()
...
+                        {
+                            var id = File.ReadAllText(agsharePath).Trim();
+                            alreadyExists = Guid.TryParse(id, out Guid guid) && guid == field.Id;
+                        }
+                        catch { }
+                    }
+
+                    if (alreadyExists && !forceOverwrite)
+                    {
+                        skipped++;
```
FantasyTeddy, line 148: Maybe also add a log statement here, that this field was skipped?

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
```diff
@@ -21,59 +42,124 @@ public static LocalFieldModel Parse(AgShareFieldDto dto)
...
+                        if (point == null)
+                        {
+                            continue; // Skip null points
+                        }
+
+                        // Validate coordinates
+                        if (point.Latitude < -90 || point.Latitude > 90 ||
+                            point.Longitude < -180 || point.Longitude > 180)
+                        {
+                            continue; // Skip invalid coordinates
```
FantasyTeddy, line 81: Maybe we should add some log statements in these cases? Otherwise there might be strange issues that may be hard to debug without that information.

SourceCode/GPS/Forms/Field/FormAgShareDownloader.cs
```diff
@@ -171,8 +178,12 @@ private async void BtnDownloadAll_Click(object sender, EventArgs e)
             chkForceOverwrite.Enabled = true;
 
             // Show result
-            gps.TimedMessageBox(3000, "AgShare",
-                $"Downloaded {result.Downloaded} new field(s).\nSkipped {result.Skipped} existing.");
+            string message = $"Downloaded {result.Downloaded} new field(s).\nSkipped {result.Skipped} existing.";
```
FantasyTeddy: Maybe skipped fields could be added conditionally like failed ones?

SourceCode/GPS/Classes/AgShare/AgShareDownloader.cs
```diff
@@ -25,14 +25,85 @@ public CAgShareDownloader()
             client = new AgShareClient(Settings.Default.AgShareServer, Settings.Default.AgShareApiKey);
         }
 
+        /// <summary>
+        /// Validates that a field DTO contains the minimum required data to be processed safely
+        /// </summary>
+        private bool ValidateFieldDto(AgShareFieldDto dto)
+        {
+            if (dto == null)
+                return false;
```
J005t67: Add curly braces around the if block

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
```diff
@@ -21,59 +42,124 @@ public static LocalFieldModel Parse(AgShareFieldDto dto)
...
-                    PtB = new LocalPoint(vB.Easting, vB.Northing),
-                    CurvePoints = new List<LocalPoint>()
-                };
-
-                if (ab.Coords.Count > 2)
-                {
-                    for (int i = 0; i < ab.Coords.Count; i++)
+                    if (ab.Coords[0].Latitude < -90 || ab.Coords[0].Latitude > 90 ||
+                        ab.Coords[0].Longitude < -180 || ab.Coords[0].Longitude > 180 ||
+                        ab.Coords[1].Latitude < -90 || ab.Coords[1].Latitude > 90 ||
```
J005t67: You added these range test 4 times, please implement it only once in a separation function. Why not use the type Wgs84 (in stead of your own CoordDto) and add a method IsValid to Wgs84

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
```diff
@@ -21,59 +42,124 @@ public static LocalFieldModel Parse(AgShareFieldDto dto)
...
+                    var vA = converter.ToLocal(ab.Coords[0].Latitude, ab.Coords[0].Longitude);
+                    var vB = converter.ToLocal(ab.Coords[1].Latitude, ab.Coords[1].Longitude);
+                    double heading = GeoConverter.HeadingFromPoints(vA, vB);
 
-                        if (i < ab.Coords.Count - 1)
+                    var abLine = new AbLineLocal
+                    {
+                        Name = ab.Name ?? "Unnamed",
+                        Heading = heading,
+                        PtA = new LocalPoint(vA.Easting, vA.Northing),
```
J005t67, line 127: Why do you use LocalPoint here (and in other parts of the code)? GeoCoord and GeoCoordDir have everthing you need.

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
FantasyTeddy: I think having a separate `CoordinateDto` does make sense, because as its name implies, it is used as a _data transfer object_ for the AgShare API. Using `Wgs84` would couple our internal representation of a coodinate to this API, which is usually not desirable.

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
```diff
@@ -21,59 +42,124 @@ public static LocalFieldModel Parse(AgShareFieldDto dto)
...
+                        }
+
+                        // Validate coordinates
+                        if (point.Latitude < -90 || point.Latitude > 90 ||
+                            point.Longitude < -180 || point.Longitude > 180)
+                        {
+                            continue; // Skip invalid coordinates
+                        }
+
+                        var local = converter.ToLocal(point.Latitude, point.Longitude);
```
J005t67, line 84: Why does ToLocal return a Vec2? We already have GeoCoord for this? Why can't we use LocalPlane to do the conversions?

richardklasens:
Address review comments from FantasyTeddy and J005t67 at commit https://github.com/AgOpenGPS-Official/AgOpenGPS/commit/d6ac513c5091585b65dfc745015e40b5e8e1209e

- Move ValidateFieldDto logic into AgShareFieldParser.Parse method
- Add log statement when field is skipped in DownloadAllAsync
- Add log statements for skipped invalid coordinates
- Show skipped count conditionally like failed count
- Add curly braces around single-line if statement
- Create separate IsValidCoordinate function to avoid code duplication

SourceCode/GPS/Classes/AgShare/AgShareDownloader.cs
```diff
@@ -54,51 +72,130 @@ public async Task<List<AgShareGetOwnFieldDto>> GetOwnFieldsAsync()
...
+
+                if (dto == null)
+                {
+                    Log.EventWriter($"[AgShare] Preview download failed for fieldId={fieldId}: Failed to deserialize field data");
+                    return null;
+                }
+
+                // Validation is now done in Parse method, but we'll do a quick check here
+                // since Parse throws exceptions and we want to return null for preview failures
+                try
```
FantasyTeddy: We already have a `try`-`catch` statement around the entire logic. Do we need this additional one?

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
```diff
@@ -21,59 +56,123 @@ public static LocalFieldModel Parse(AgShareFieldDto dto)
...
+                            Log.EventWriter($"[AgShare] Skipping invalid boundary coordinate in field '{dto.Name}': Lat={point.Latitude}, Lon={point.Longitude}");
+                            continue; // Skip invalid coordinates
+                        }
+
+                        var local = converter.ToLocal(point.Latitude, point.Longitude);
+                        ringList.Add(new LocalPoint(local.Easting, local.Northing));
+                    }
+
+                    // Only add rings with at least 3 points (minimum for a valid polygon)
+                    if (ringList.Count >= 3)
```
FantasyTeddy, line 89: Do we also need some log statements for cases where we don't have a valid polygon?

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
```diff
@@ -21,59 +56,123 @@ public static LocalFieldModel Parse(AgShareFieldDto dto)
...
+                    {
+                        Name = ab.Name ?? "Unnamed",
+                        Heading = heading,
+                        PtA = new LocalPoint(vA.Easting, vA.Northing),
+                        PtB = new LocalPoint(vB.Easting, vB.Northing),
+                        CurvePoints = new List<LocalPoint>()
+                    };
+
+                    if (ab.Coords.Count > 2)
+                    {
```
FantasyTeddy, line 133: Maybe same here: do we need a log statement if an AB line does not have enough coordinates?

FantasyTeddy (approved):

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
richardklasens: I can never be a not valid polygon, AgShare wont except it either

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
richardklasens: Same here..

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
FantasyTeddy: Sure, but that is true for most of the other safety checks as well. Why even bother doing them if it cannot happen?

PR #1018, opened 2025-10-24
Display of the angle and inverse angle of the current abline

chrisriteco:
Display of the angle and inverse angle of the current abline in the currentField line (top main window). Accuracy of 5 decimal places. 1. to visualize the abline angle when working with abline names; 2. to verify the direction of travel—for example, when driving with a Dual system; 3. for exchange with other drivers/systems.

abheadingandreverse

### DIFF
--- a/SourceCode/GPS/Forms/GUI.Designer.cs
+++ b/SourceCode/GPS/Forms/GUI.Designer.cs
@@ -198,11 +198,12 @@ namespace AgOpenGPS
                                     oppositeAbAngle = oppositeAbAngle - 360;
                                 }
 
-                                lblCurrentField.Text = "Line: " + trk.gArr[trk.idx].name + " - " + glm.toDegrees(trk.gArr[trk.idx].heading).ToString("N3") + ", " + oppositeAbAngle.ToString("N3");
+                                GeoDir headingDir = new GeoDir(trk.gArr[trk.idx].heading);
+                                lblCurrentField.Text = gStr.gsABline + ": " + trk.gArr[trk.idx].name + "  " + headingDir.HeadingString("N3") + ", " + headingDir.Inverted.HeadingString("N3");
                             }
                             else
                             {
-                                lblCurrentField.Text = "Line: " + gStr.gsNoGuidanceLines;
+                                lblCurrentField.Text = gStr.gsABline + ": " + gStr.gsNoGuidanceLines;
                             }
                             break;
 


richardklasens (commented):
Hi Chris.

I understand this, but do you really need 5 decimals? 2 should be enough for visual check?
Other then that, where is your LatLon for Point A/B?

chrisriteco:
Deviation of 0.001° over 100 meters gives a lateral deviation of approximately 0.17 cm, every 0.01° a deviation of 1.75 cm. If there is visual space, I would go for the 3 decimal places - then there is also value in exchanging with another system.
Display of the LatLon for Point A/B is in the section where 'isJobStarted' is not 'true'. Nothing has changed here.

SourceCode/GPS/Forms/GUI.Designer.cs
```diff
@@ -189,7 +189,15 @@ private void tmrWatchdog_tick(object sender, EventArgs e)
 
                         case 2:
                             if (trk.idx > -1)
-                                lblCurrentField.Text = "Line: " + trk.gArr[trk.idx].name;
+                            {
+                                double turnedAbAngle = 0;
```
J005t67: No need to  assign 0 first. Simply assign the proper value from the start.

SourceCode/GPS/Forms/GUI.Designer.cs
```diff
@@ -189,7 +189,15 @@ private void tmrWatchdog_tick(object sender, EventArgs e)
...
+                                turnedAbAngle = glm.toDegrees(trk.gArr[trk.idx].heading) + 180;
+                                if (turnedAbAngle > 360)
+                                    turnedAbAngle = turnedAbAngle - 360;
```
J005t67: Use { and } for single line if statements.

SourceCode/GPS/Forms/GUI.Designer.cs
```diff
@@ -189,7 +189,15 @@ private void tmrWatchdog_tick(object sender, EventArgs e)
...
+
+                                turnedAbAngle = Math.Round(turnedAbAngle, 5);
+                                lblCurrentField.Text = "Line: " + trk.gArr[trk.idx].name + " - " + Convert.ToString(Math.Round(glm.toDegrees(trk.gArr[trk.idx].heading), 5)) + ", " + turnedAbAngle;
+                            }
                             else
                                 lblCurrentField.Text = "Line: " + gStr.gsNoGuidanceLines;
```
J005t67: Use { and } for the else block too.

SourceCode/GPS/Forms/GUI.Designer.cs
```diff
@@ -189,7 +189,15 @@ private void tmrWatchdog_tick(object sender, EventArgs e)
...
                             if (trk.idx > -1)
-                                lblCurrentField.Text = "Line: " + trk.gArr[trk.idx].name;
+                            {
+                                double turnedAbAngle = 0;
+                                turnedAbAngle = glm.toDegrees(trk.gArr[trk.idx].heading) + 180;
+                                if (turnedAbAngle > 360)
+                                    turnedAbAngle = turnedAbAngle - 360;
+
+                                turnedAbAngle = Math.Round(turnedAbAngle, 5);
+                                lblCurrentField.Text = "Line: " + trk.gArr[trk.idx].name + " - " + Convert.ToString(Math.Round(glm.toDegrees(trk.gArr[trk.idx].heading), 5)) + ", " + turnedAbAngle;
```
J005t67: Do not use Math.Round to force the number of decimals. It won't work. Numbers like 4.000 will be displayed without any decimals and that makes it impossible for the user to determine to accuracy.  In stead, you should pass a format string to the function ToString();
Write something like glm.toDegrees(trk.gArr[trk.idx].heading).ToString("N5")

SourceCode/GPS/Forms/GUI.Designer.cs
J005t67: I think the name oppositeAbAngle would be more accurate.

SourceCode/GPS/Forms/GUI.Designer.cs
J005t67: The word Line needs to be translated too.

SourceCode/GPS/Forms/GUI.Designer.cs
```diff
@@ -192,16 +192,18 @@ private void tmrWatchdog_tick(object sender, EventArgs e)
...
-                                    turnedAbAngle = turnedAbAngle - 360;
+                                double oppositeAbAngle = glm.toDegrees(trk.gArr[trk.idx].heading) + 180;
+                                if (oppositeAbAngle > 360)
+                                {
+                                    oppositeAbAngle = oppositeAbAngle - 360;
+                                }
 
-                                turnedAbAngle = Math.Round(turnedAbAngle, 5);
-                                lblCurrentField.Text = "Line: " + trk.gArr[trk.idx].name + " - " + Convert.ToString(Math.Round(glm.toDegrees(trk.gArr[trk.idx].heading), 5)) + ", " + turnedAbAngle;
+                                lblCurrentField.Text = "Line: " + trk.gArr[trk.idx].name + " - " + glm.toDegrees(trk.gArr[trk.idx].heading).ToString("N3") + ", " + oppositeAbAngle.ToString("N3");
```
J005t67: I overlooked it during previous review, but the degree symol "\u00B0" is still missing. So we still need some rework. To make things a bit more easy, I have added a new method HeadingString to the struct GeoDir. The review is stil pending, but when it is finished, you could write things like:

GeoDir headingDir = new GeoDir(trk.gArr[trk.idx].heading);
lblCurrentField.Text = "Line: " + trk.gArr[trk.idx].name + " - " + headingDir .HeadingString("N3") + ", " + headingDir .Inverse.HeadingString("N3");

Note that a GeoDir only takes 8 bytes of memory, which is the size of the only element AngleInRadians that it contains. So GeoDir can be used to replace any variable of type double that represents a direction, without any overhead, while adding a lot of nice features that a normal double lacks.

By the way, I find the dash (or is it a minus sign) in front of the number very confusing because it looks like a minus sign.

PR #1032, opened 2025-10-29
Fix for UpdateLabels in FormFlags

Rob2029:
A fix for the exception that happens when the `mf.flagPts.Count = 1`  and `mf.flagNumberPicked` was set to 0

### DIFF
--- a/SourceCode/GPS/Forms/Field/FormFlags.cs
+++ b/SourceCode/GPS/Forms/Field/FormFlags.cs
@@ -25,9 +25,7 @@ namespace AgOpenGPS
         {
             if (mf.flagNumberPicked > mf.flagPts.Count)
             {
-                if (mf.flagPts.Count == 1) mf.flagNumberPicked = 1;
-                else
-                    mf.flagNumberPicked = mf.flagPts.Count - 1;
+                mf.flagNumberPicked = mf.flagPts.Count;
             }
             lblLatStart.Text = mf.flagPts[mf.flagNumberPicked - 1].latitude.ToString();
             lblLonStart.Text = mf.flagPts[mf.flagNumberPicked - 1].longitude.ToString();


SourceCode/GPS/Forms/Field/FormFlags.cs
```diff
@@ -25,7 +25,9 @@ private void UpdateLabels()
         {
             if (mf.flagNumberPicked > mf.flagPts.Count)
             {
-                mf.flagNumberPicked = mf.flagPts.Count - 1;
+                if (mf.flagPts.Count == 1) mf.flagNumberPicked = 1;
```
J005t67: A flag count of 1 should not be treated as a special case. 
A flag count of zero would be a special case, but currently the window closes when the last flag is deleted and we do not need to take care of 'zero flags'
So, for  all flag counts above zero , you can simply write the following without the if statement.
mf.flagNumberPicked = mf.flagPts.Count;

SourceCode/GPS/Forms/Field/FormFlags.cs
```diff
@@ -25,7 +25,9 @@ private void UpdateLabels()
...
+                else
+                    mf.flagNumberPicked = mf.flagPts.Count - 1;
```
J005t67: The cause of the problem is the following function in Controls.Designer.cs.
On the first line it sets nextFlag, but this value  becomes incorrect when  
DeduplicateFlags removes one or more flags. So nextFlag must be set after calling DepulicateFlags()

private void btnFlag_Click(object sender, EventArgs e)
        {
            int nextflag = flagPts.Count + 1;      \\ This value becomes incorrect after DeduplicateFlags
            CFlag flagPt = new CFlag(
                AppModel.CurrentLatLon.Latitude,
                AppModel.CurrentLatLon.Longitude,
                pn.fix.easting, pn.fix.northing,
                fixHeading, flagColor, nextflag, nextflag.ToString());
            flagPts.Add(flagPt);
            flagPts = FlagsFiles.DeduplicateFlags(flagPts);
            FileSaveFlags();

Rob2029: I tried calling DeduplicateFlags before nextFlag, but that did not solved the issue 
When there is just 1 flag and Deduplicate is called first is still possible to have 2 flags in the same spot until DeduplicateFlags is called again
So I just set mf.flagNumberPicked = mf.flagPts.Count;
And that seem to solve the issue, but now I am not really sure if that is a solid fix

FantasyTeddy:
@J005t67: Is there anything stopping us from merging this?
Should it maybe go into the `fixes/6.8` branch?

J005t67:
> @J005t67: Is there anything stopping us from merging this? Should it maybe go into the `fixes/6.8` branch?

This bug fix is related to the new function DeduplicateFlags, which on its turn is related to the relatively new import/export of flags. I just cheked: Import/export of flags is in 6.8, but the function DeduplicateFlags is not (yet). It seems that I approved the PR for DeduplicateFlags for the development branch, but it should have been merged to the fixes/6.8 branch.

FantasyTeddy:
> This bug fix is related to the new function DeduplicateFlags, which on its turn is related to the relatively new import/export of flags. I just cheked: Import/export of flags is in 6.8, but the function DeduplicateFlags is not (yet). It seems that I approved the PR for DeduplicateFlags for the development branch, but it should have been merged to the fixes/6.8 branch.

You mean #970, correct? If so, I would argue that it is fine to get this in the next release and not in 6.8.

PR #1038, opened 2025-11-12
Adding the wheel angle of keya to the SIM value

gunicsba:
Making a PR as i have problems with the local build...


Pat-I:
Has this been tested without Keya and without board before been merged?

PR #1040, opened 2025-11-13
Adds track copying functionality

richardklasens:
Adds the ability to copy tracks between fields, including coordinate system conversion.

This change introduces a new `TrackCopier` class with methods to:

- Convert tracks between fields, handling coordinate system transformations.
- Copy tracks from a source field to a target field, saving and reloading as needed.

It also adds a `FormCopyTracks` form that allows the user to select a source field and tracks to copy to the current field.

### DIFF
--- a/SourceCode/GPS/Forms/Field/FormCopyTracks.cs
+++ b/SourceCode/GPS/Forms/Field/FormCopyTracks.cs
@@ -235,8 +235,22 @@ namespace AgOpenGPS.Forms.Field
 
                 // Clear and reload tracks to ensure all indices and references are correct
                 mf.trk.gArr?.Clear();
-                mf.trk.idx = -1;
                 mf.FileLoadTracks();
+
+                // Set index to first visible track if available, otherwise -1
+                if (mf.trk.gArr.Count > 0)
+                {
+                    mf.trk.idx = 0;
+                    // Find first visible track
+                    for (int i = 0; i < mf.trk.gArr.Count; i++)
+                    {
+                        if (mf.trk.gArr[i].isVisible)
+                        {
+                            mf.trk.idx = i;
+                            break;
+                        }
+                    }
+                }
 
                 lblStatus.Text = $"Successfully imported {copiedCount} track(s)";
 


SourceCode/GPS/Forms/FormGPS.Designer.cs
```diff
@@ -1976,14 +1978,23 @@ private void InitializeComponent()
...
             this.recordedPathStripMenu.Click += new System.EventHandler(this.recordedPathStripMenu_Click);
-            // 
+            //
+            // copyTracksToolStripMenuItem
+            //
+            this.copyTracksToolStripMenuItem.Font = new System.Drawing.Font("Tahoma", 26.25F, System.Drawing.FontStyle.Bold, System.Drawing.GraphicsUnit.Point, ((byte)(0)));
+            this.copyTracksToolStripMenuItem.Image = global::AgOpenGPS.Properties.Resources.FileNew;
+            this.copyTracksToolStripMenuItem.Name = "copyTracksToolStripMenuItem";
+            this.copyTracksToolStripMenuItem.Size = new System.Drawing.Size(445, 70);
+            this.copyTracksToolStripMenuItem.Text = "Copy Tracks";
```
FantasyTeddy: Is "Copy" the correct word here? I got the impression that I could "copy" any track from any field into any other field. But that is not quite what this does: you can copy any track from any field **into the current field**. For me personally, that is more like an "Import" functionality.

SourceCode/GPS/Forms/Field/FormCopyTracks.cs
```diff
@@ -0,0 +1,279 @@
...
+    /// <summary>
+    /// Form that allows the user to copy tracks from other fields to the current field.
+    /// Handles coordinate conversion between different field origins.
+    /// </summary>
+    public partial class FormCopyTracks : Form
+    {
+        private readonly FormGPS mf;
+        private string selectedFieldDirectory;
+        private List<CTrk> availableTracks;
+        private List<FieldDescription> availableFields;
```
FantasyTeddy: These fields are only used in one method each (`availableTracks` => `LoadTracksFromField`, `availableFields` => `LoadFieldList`). Therefore, they can/should be local variables instead of fields.

SourceCode/GPS/Forms/Field/FormCopyTracks.cs
```diff
@@ -0,0 +1,279 @@
...
+                        availableFields.Add(fieldDescription);
+
+                        var item = new ListViewItem(fieldDescription.Name);
+                        item.Tag = fieldDescription;
+                        lbFields.Items.Add(item);
+                    }
+                }
+
+                lbFields.EndUpdate();
+                lblStatus.Text = $"Found {availableFields.Count} field(s) with tracks";
```
FantasyTeddy: It looks like `availableFields` is only used to count the number of fields? Couldn't that be done differently, like looking at the number of fields in the list (`lbFields.Items.Count`)?

SourceCode/GPS/Forms/Field/FormCopyTracks.cs
```diff
@@ -0,0 +1,279 @@
...
+                {
+                    // Check if field has Field.txt and TrackLines.txt
+                    string fieldFile = Path.Combine(fieldDir, "Field.txt");
+                    string trackFile = Path.Combine(fieldDir, "TrackLines.txt");
+
+                    if (File.Exists(fieldFile) && File.Exists(trackFile))
+                    {
+                        // Don't show the current field in the list
+                        if (mf.currentFieldDirectory != null &&
+                            Path.GetFullPath(fieldDir) == Path.GetFullPath(currentFieldFullPath))
```
FantasyTeddy: Do we need to compare the absolute path of the fields? Wouldn't it be enough to compare the field name (i.e. directory name)?

SourceCode/GPS/Forms/Field/FormCopyTracks.cs
```diff
@@ -0,0 +1,279 @@
...
+
+                string currentFieldFullPath = Path.Combine(fieldsDir, mf.currentFieldDirectory ?? "");
+
+                foreach (var fieldDir in fieldDirs)
+                {
+                    // Check if field has Field.txt and TrackLines.txt
+                    string fieldFile = Path.Combine(fieldDir, "Field.txt");
+                    string trackFile = Path.Combine(fieldDir, "TrackLines.txt");
+
+                    if (File.Exists(fieldFile) && File.Exists(trackFile))
```
FantasyTeddy: Maybe do an early return here:
```suggestion
                    if (!File.Exists(fieldFile) || !File.Exists(trackFile))
                        continue;
```

SourceCode/GPS/Forms/Field/FormCopyTracks.cs
```diff
@@ -0,0 +1,279 @@
...
+
+                        // Try to load field info
+                        Wgs84? wgs84Start = null;
+                        try
+                        {
+                            wgs84Start = FieldPlaneFiles.LoadOrigin(fieldDir);
+                        }
+                        catch { }
+
+                        var fieldDescription = new FieldDescription(fieldDirInfo, wgs84Start, null);
```
FantasyTeddy: I don't think the `FieldDescription` class should be used here. You are only interested in the `DirectoryInfo` of the field, so maybe just store that?

SourceCode/GPS/Forms/Field/FormCopyTracks.cs
```diff
@@ -0,0 +1,279 @@
...
+
+                        var fieldDirInfo = new DirectoryInfo(fieldDir);
+
+                        // Try to load field info
+                        Wgs84? wgs84Start = null;
+                        try
+                        {
+                            wgs84Start = FieldPlaneFiles.LoadOrigin(fieldDir);
+                        }
+                        catch { }
```
FantasyTeddy: I don't think the `Wgs84Start` in the `FieldDescription` is actually being used. So this is not necessary.

SourceCode/GPS/Forms/Field/FormCopyTracks.cs
```diff
@@ -0,0 +1,279 @@
...
+                mf.trk.gArr?.Clear();
+                mf.trk.idx = -1;
+                mf.FileLoadTracks();
+
+                lblStatus.Text = $"Successfully copied {copiedCount} track(s)";
+
+                mf.TimedMessageBox(2000, "Copy Tracks",
+                    $"Successfully copied {copiedCount} track(s) to current field.");
+
+                Close();
```
FantasyTeddy: Are you sure that we want to close the form after copying? What if I want to copy some more tracks from a different field?

SourceCode/GPS/Classes/TrackCopier.cs
```diff
@@ -0,0 +1,192 @@
...
+    {
+        /// <summary>
+        /// Convert tracks from one field to another, handling coordinate system transformations.
+        /// </summary>
+        /// <param name="tracks">Tracks to convert</param>
+        /// <param name="sourceFieldDirectory">Source field directory path</param>
+        /// <param name="targetFieldDirectory">Target field directory path</param>
+        /// <param name="sharedFieldProperties">Shared field properties for coordinate conversion</param>
+        /// <returns>List of converted tracks</returns>
+        public static List<CTrk> ConvertTracks(
```
FantasyTeddy:
```suggestion
        private static List<CTrk> ConvertTracks(
```

SourceCode/GPS/Forms/FormGPS.Designer.cs
richardklasens: changed to Import Tracks

SourceCode/GPS/Forms/Field/FormCopyTracks.cs
richardklasens: changed to local variables

SourceCode/GPS/Forms/Field/FormCopyTracks.cs
richardklasens: changed

SourceCode/GPS/Forms/Field/FormCopyTracks.cs
richardklasens: removed

SourceCode/GPS/Classes/TrackCopier.cs
richardklasens: changed

SourceCode/GPS/Forms/Field/FormCopyTracks.cs
richardklasens: removed

SourceCode/GPS/Forms/Field/FormCopyTracks.cs
richardklasens: changed

FantasyTeddy (approved):

PR #1043, opened 2025-11-16
Add GeoLineSegment

J005t67:
Another small step for the implementation of KmlFieldImporter.

The method IntersectionPoint() will be used to correct self-intersecting boundaries.

### DIFF
--- a/SourceCode/AgOpenGPS.Core.Tests/Models/Base/GeoLineSegmentTests.cs
+++ b/SourceCode/AgOpenGPS.Core.Tests/Models/Base/GeoLineSegmentTests.cs
@@ -5,15 +5,6 @@ namespace AgOpenGPS.Core.Tests.Models
 {
     public class GeoLineSegmentTests
     {
-        const double _minNorthing = -1.0;
-        const double _maxNorthing = 3.0;
-        const double _minEasting = 2.0;
-        const double _maxEasting = 4.0;
-        readonly GeoCoord _neCoord = new GeoCoord(_maxNorthing, _maxEasting);
-        readonly GeoCoord _seCoord = new GeoCoord(_minNorthing, _maxEasting);
-        readonly GeoCoord _swCoord = new GeoCoord(_minNorthing, _minEasting);
-        readonly GeoCoord _nwCoord = new GeoCoord(_maxNorthing, _minEasting);
-
 
         [Test]
         public void Test_Intersects()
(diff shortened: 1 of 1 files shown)


SourceCode/AgOpenGPS.Core.Tests/Models/Base/GeoLineSegmentTests.cs
```diff
@@ -0,0 +1,100 @@
...
+    public class GeoLineSegmentTests
+    {
+        const double _minNorthing = -1.0;
+        const double _maxNorthing = 3.0;
+        const double _minEasting = 2.0;
+        const double _maxEasting = 4.0;
+        readonly GeoCoord _neCoord = new GeoCoord(_maxNorthing, _maxEasting);
+        readonly GeoCoord _seCoord = new GeoCoord(_minNorthing, _maxEasting);
+        readonly GeoCoord _swCoord = new GeoCoord(_minNorthing, _minEasting);
+        readonly GeoCoord _nwCoord = new GeoCoord(_maxNorthing, _minEasting);
```
FantasyTeddy: Is there a particular reason why these values should be fields and not just variables inside their respective tests?

SourceCode/AgOpenGPS.Core.Tests/Models/Base/GeoLineSegmentTests.cs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            GeoCoord coordA = new GeoCoord(13.0, -10.0);
+            GeoCoord coordB = new GeoCoord(12.0, -19.0);
+            GeoCoord otherCoordA = new GeoCoord(14.0, -18.0);
+            GeoCoord otherCoordB = new GeoCoord(-8, -5);
+            GeoLineSegment lineSegment = new GeoLineSegment(coordA, coordB);
+            GeoLineSegment otherLineSegment = new GeoLineSegment(otherCoordA, otherCoordB);
+            GeoCoord? interSectionPoint = lineSegment.IntersectionPoint(otherLineSegment);
+
+            Assert.That(interSectionPoint.HasValue, Is.EqualTo(true));
```
FantasyTeddy, line 20: Shouldn't we also test the resulting `GeoCoord` for correctness?

SourceCode/AgOpenGPS.Core.Tests/Models/Base/GeoLineSegmentTests.cs
```diff
@@ -0,0 +1,100 @@
+﻿using AgOpenGPS.Core.Models;
+using NUnit.Framework;
+
+namespace AgOpenGPS.Core.Tests.Models
+{
+    public class GeoLineSegmentTests
```
FantasyTeddy, line 6: Maybe we should also have a test that verifies that two line segments that are not parallel, but do not intersect actually don't intersect.

SourceCode/AgOpenGPS.Core.Tests/Models/Base/GeoLineSegmentTests.cs
J005t67: These GeoCoords are used in three different test cases

SourceCode/AgOpenGPS.Core.Tests/Models/Base/GeoLineSegmentTests.cs
J005t67: Done

SourceCode/AgOpenGPS.Core.Tests/Models/Base/GeoLineSegmentTests.cs
J005t67: Done

FantasyTeddy: > These GeoCoords are used in three different test cases

I know, but what is the benefit of sharing these GeoCoords between tests?
I would argue that it makes the tests harder to read, because you have to jump back up to see what these values are.

SourceCode/AgOpenGPS.Core.Tests/Models/Base/GeoLineSegmentTests.cs
```diff
@@ -27,6 +27,30 @@ public void Test_Intersects()
...
+            // Intersection point must lie on first segment
+            Assert.That(
+                coordA.Distance(interSectionPoint.Value) + interSectionPoint.Value.Distance(coordB),
+                Is.EqualTo(coordA.Distance(coordB))
+            );
+            // Intersection point must lie on other segment too
+            Assert.That(
+                otherCoordA.Distance(interSectionPoint.Value) + interSectionPoint.Value.Distance(otherCoordB),
+                Is.EqualTo(otherCoordA.Distance(otherCoordB))
+            );
```
FantasyTeddy, line 30: Why not test for the exact intersection point value?

J005t67: Althoug the assertions do not litterly mention the correct value, the assertions are only true if the intersectionPoint is correct, so they do test for the exact intersection point value.
I lthink the asserts are an elegant way to test the correctness, without the need to recompute the exact intersection point in the test. (That would incur the risk of making the same bug in the test and the tested code, and as a result overlook a bug.)

SourceCode/AgOpenGPS.Core.Tests/Models/Base/GeoLineSegmentTests.cs
J005t67: OK, will change it

SourceCode/AgOpenGPS.Core.Tests/Models/Base/GeoLineSegmentTests.cs
FantasyTeddy: I get your point. However, by including calculations instead of the exact intersection point, two things can happen:
- The test will fail if the implementation of `Distance` is broken. This is not critical, but unnecessary.
- It makes the test harder to understand, even though the solution is pretty elegant.

But we can leave it as is. It's a detail.

FantasyTeddy (approved):

PR #1061, opened 2025-11-24
Enrich GLW

J005t67:
- Split GLW before it becomes too big
 - Remove file GeoLine.cs (merged from 6.8.1) Use GeoLineSegment in stead
 - Let LineStyle and PointStyle use ColorRgba

### DIFF
--- a/SourceCode/GPS/Classes/CABLine.cs
+++ b/SourceCode/GPS/Classes/CABLine.cs
@@ -1,6 +1,6 @@
+using AgOpenGPS.Core.Drawing;
 using AgOpenGPS.Core.DrawLib;
 using AgOpenGPS.Core.Models;
-using OpenTK.Graphics.OpenGL;
 using System;
 using System.Collections.Generic;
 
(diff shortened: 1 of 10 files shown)


SourceCode/GPS/Classes/CABLine.cs
richardklasens, line 1: I should definitely first move all guidance methods into one class like Brian did for his project and remove CABLine.cs and CABCurve.cs

SourceCode/AgOpenGPS.Core/Drawing/Colors.cs
```diff
@@ -5,38 +5,38 @@ namespace AgOpenGPS.Core.Drawing
...
 
-        static public ColorRgb HitchColor = new ColorRgb(0.765f, 0.76f, 0.32f);
-        static public ColorRgb HitchTrailingColor = new ColorRgb(0.7f, 0.4f, 0.2f);
-        static public ColorRgb HitchRigidColor = new ColorRgb(0.237f, 0.037f, 0.0397f);
+        static public readonly ColorRgba HitchColor = new ColorRgba(0.765f, 0.76f, 0.32f);
+        static public readonly ColorRgba HitchTrailingColor = new ColorRgba(0.7f, 0.4f, 0.2f);
+        static public readonly ColorRgba HitchRigidColor = new ColorRgba(0.237f, 0.037f, 0.0397f);
 
-        static public ColorRgb SvenArrowColor = new ColorRgb(0.95f, 0.95f, 0.10f);
+        static public readonly ColorRgba SvenArrowColor = new ColorRgba(0.95f, 0.95f, 0.10f);
```
FantasyTeddy, line 31: Why use `ColorRgba` if we are just describing a color without alpha value?
I think it would be much cleaner if we had for example some implicit conversions from `ColorRgb` to `ColorRgba`.

SourceCode/GPS/Classes/CABLine.cs
J005t67: Don't worry to much about my changes to CABLine, this PR is about making the OpenGL wrapper code more mature.
In general, wrapping external libraries is a good practice. More specificly for GLW the benifits are (a.o.):
-it helps to reduce the dependencies to the specific OpenGL toolkit that we are using now. This might become important one day if we want to switch to another OpenGL toolkit one day.
- the wraper is a nice central place to start implementing error checking on the OpenGL calls
- the wrapper is more geared towards the rest of the code in the sense that it supports GeoCoords in its interface

The sole reason that I changed CABLine was that I wanted to test my new wrapper functions. If CABLine does not survive 2026, that is OK with me.

SourceCode/AgOpenGPS.Core/Drawing/Colors.cs
J005t67: LineStyle and PointStyle become more easy if they only have to deal with RgbaColor. Also, I don't want to do color conversions for each frame, but only once, during initialization.

PR #1063, opened 2025-11-24
parsed wrong field for roll. (check with unicore, bynav, comnav)

aortner:
unicore:
image

bynav:
image

comnav:
image

### DIFF
--- a/SourceCode/AgIO/Source/Forms/NMEA.Designer.cs
+++ b/SourceCode/AgIO/Source/Forms/NMEA.Designer.cs
@@ -948,8 +948,9 @@ namespace AgIO
                 float.TryParse(words[2], NumberStyles.Float, CultureInfo.InvariantCulture, out headingTrueDual);
                 headingTrueDualData = headingTrueDual;
 
+
                 //  Console.WriteLine(HeadingForced);
-                float.TryParse(words[3], NumberStyles.Float, CultureInfo.InvariantCulture, out rollK);
+                float.TryParse(words[4], NumberStyles.Float, CultureInfo.InvariantCulture, out rollK);
                 // Console.WriteLine(nRoll);
 
                 int.TryParse(words[5], NumberStyles.Float, CultureInfo.InvariantCulture, out int trasolution);


SourceCode/AgIO/Source/Forms/NMEA.Designer.cs
```diff
@@ -948,8 +948,9 @@ private void ParseSTI036()
                 float.TryParse(words[2], NumberStyles.Float, CultureInfo.InvariantCulture, out headingTrueDual);
                 headingTrueDualData = headingTrueDual;
 
+
                 //  Console.WriteLine(HeadingForced);
-                float.TryParse(words[3], NumberStyles.Float, CultureInfo.InvariantCulture, out rollK);
+                float.TryParse(words[4], NumberStyles.Float, CultureInfo.InvariantCulture, out rollK);
```
FantasyTeddy, line 953: Are you sure that the `rollK` variable actually refers to "roll" as the NMEA sentences do?
For example if I look at the KSXT message, it also takes "pitch" from the message for the `rollK` variable.

Pat-I:
why would you parse TRA instead HPR? Is one of them more generic? I ask because the Teensy firmware for the UM982 use HPR

aortner:
I use a ComNav A100Pro Antenna. Has build in IMU and provides Tilt Roll compensated output position with precise heading in one antenna.

so message is tra for heading pitch and roll.

https://www.comnavtech.com/product/receiver/a100.html

Optional IMU for A100Pro allows 
automatic tilt compensation,
enabling smooth operation under variable terrain and
environmental conditions.

for bynav ksxt i think the reason because antenna are side by side and not front /back (usual) so it outputs picht.

Pat-I:
TRA is **for IMU output?** I still don't have my answer about the difference between TRA and HPR

Is TRA a standard or manufacturer specific?(I guess it's at least somewhat standard if 3 manufacturers use it) If its too specific just build a PAOGI on an MCU.

Pat-I:
When was TRA parsing added to AOG?

SourceCode/AgIO/Source/Forms/NMEA.Designer.cs
farmerbriantee: Correct, it will overwrite rollK with pitch value.

farmerbriantee (approved):

Issue #1054, opened 2025-11-22
Boundary tool not accurate.


Pat-I:
### AgOpenGPS Version

v6.8.0

### Bug Description

When using the boundary tool, the boundary doesn't always follow the section coverage.

<img width="1280" height="691" alt="Image" src="https://github.com/user-attachments/assets/3a1b3877-b0cf-465a-aac2-41e883c19c37" />

### Steps to Reproduce

Just build boundaries from sections

### Expected Behavior

check the result.

### Severity and Impact

Boundary inaccuracy

### Operating System

windows11

### Hardware Setup

<img width="1178" height="887" alt="Image" src="https://github.com/user-attachments/assets/96dc7a87-3fbf-4011-a83c-ee35b62e4b74" />

### Additional Information

```text
Brian suggest to change this line (fromBndTool 825) from 0.05 to 0.02, which would change the behavior about how long it make a line straight before bending.
```

farmerbriantee:
delta for boundary point exclusion set to 0.02 from 0.05 in branch FormBndToolUpdate

PR #1067, opened 2025-11-26
Form bnd tool update

farmerbriantee:
Fix overall operation to much simpler 2 button push rather then 10 million. Distance and Smooth are saved properties. Boundary can be both edited and created.

### DIFF
--- a/SourceCode/GPS/GlobalSuppressions.cs
+++ b/SourceCode/GPS/GlobalSuppressions.cs
@@ -1,8 +1,0 @@
-﻿// This file is used by Code Analysis to maintain SuppressMessage
-// attributes that are applied to this project.
-// Project-level suppressions either have no target or are given
-// a specific target and scoped to a namespace, type, member, etc.
-
-using System.Diagnostics.CodeAnalysis;
-
-[assembly: SuppressMessage("Style", "IDE1006:Naming Styles", Justification = "<Pending>", Scope = "member", Target = "~M:AgOpenGPS.FormBndTool.btnZoomIn_Click(System.Object,System.EventArgs)")]


SourceCode/GPS/Forms/Field/FormBndTool.cs
```diff
@@ -30,11 +32,10 @@ public partial class FormBndTool : Form
 
         private double zoom = 1, sX = 0, sY = 0;
 
-        public vec3 pint = new vec3(0.0, 1.0, 0.0);
-
         public List<vec3> secList = new List<vec3>();
         public List<vec3> bndList = new List<vec3>();
         public List<vec3> smooList = new List<vec3>();
+        public List<vec3> tempList = new List<vec3>();
```
FantasyTeddy: Is this actually being used somewhere?

SourceCode/GPS/GlobalSuppressions.cs
```diff
@@ -0,0 +1,8 @@
+﻿// This file is used by Code Analysis to maintain SuppressMessage
+// attributes that are applied to this project.
+// Project-level suppressions either have no target or are given
+// a specific target and scoped to a namespace, type, member, etc.
+
+using System.Diagnostics.CodeAnalysis;
+
+[assembly: SuppressMessage("Style", "IDE1006:Naming Styles", Justification = "<Pending>", Scope = "member", Target = "~M:AgOpenGPS.FormBndTool.btnZoomIn_Click(System.Object,System.EventArgs)")]
```
FantasyTeddy: Do we need this suppression?

SourceCode/GPS/Forms/Field/FormBndTool.cs
farmerbriantee: Not sure, Was in twol so came across.

farmerbriantee: I'll check

farmerbriantee: Not needed

farmerbriantee: What is the best way to delete the 2 lines with tempList?

SourceCode/GPS/GlobalSuppressions.cs
farmerbriantee: It should be throughout the entire project. Winforms makes small pre camel case, then proceeeds to call them an IDE message - hundreds of them

SourceCode/GPS/Forms/Field/FormBndTool.cs
farmerbriantee: Doing recheck

FantasyTeddy: > What is the best way to delete the 2 lines with tempList?

Just like you did: add a new commit with the correction.

SourceCode/GPS/GlobalSuppressions.cs
FantasyTeddy: You are right, but it is only reported as a suggestion (not an error or a warning).
Therefore, I don't think it is worth adding hundreds of suppressions, do you?

SourceCode/GPS/Forms/Field/FormBndTool.cs
farmerbriantee: Also there is the delta change required from issue https://github.com/AgOpenGPS-Official/AgOpenGPS/issues/1054 That Pat opened. Can fix it here now - if that is ok

SourceCode/GPS/GlobalSuppressions.cs
farmerbriantee: I don't, but I thought there used to be a global way to remove IDE 1006. Messages are useful  but they get buried with 1006

farmerbriantee: i pushed the templist removal and the delta fix

FantasyTeddy: > I don't, but I thought there used to be a global way to remove IDE 1006. Messages are useful but they get buried with 1006

There is. Maybe it was lost when I added some analyzer stuff.
Can you add it to the `.editorconfig` file?

farmerbriantee: I think you best do that. I seem to screw it up.

FantasyTeddy: Okay. But can you please remove this file?

farmerbriantee: yes

farmerbriantee: [assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Style", "IDE1006:Naming Styles", Justification = "TBD", Scope = "member", Target = "*")]

This used to work - but doesn't anymore either. Deleted Global Sup

farmerbriantee: Ready to go

FantasyTeddy: Perfect, thank you!

Issue #1053, opened 2025-11-20
Guidelines short of boundary


farmerbriantee:
Noticed on several fields the guidelines for curve don't reach the field boundary

<img width="1397" height="970" alt="Image" src="https://github.com/user-attachments/assets/18e75c6e-afa7-4d82-b2d4-b6c68a9dabb6" />

PR #1071, opened 2025-11-27
Curve guidelines extended

farmerbriantee:
Issue #1053 
GuideLines adding extensions
Simplifying of object initializations
Make Static several functions in Curve

### DIFF
--- a/SourceCode/GPS/Classes/CABCurve.cs
+++ b/SourceCode/GPS/Classes/CABCurve.cs
@@ -584,19 +584,7 @@ namespace AgOpenGPS
                     if (newGuideList == null || newGuideList.Count == 0)
                         continue;
 
-                    vec3 endA = new vec3
-                    {
-                        easting = newGuideList[0].easting - (Math.Sin(newGuideList[0].heading) * 2000.0),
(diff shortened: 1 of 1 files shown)


SourceCode/GPS/Classes/CABCurve.cs
```diff
@@ -679,6 +692,26 @@ private List<List<vec3>> BuildCurveGuidelines(double distAway, int _passes, CTrk
...
+                            northing = newGuideList[0].northing - (Math.Cos(newGuideList[0].heading) * 2000.0)
+                        };
+                        newGuideList.Insert(0, endA);
+
+                        endA = new vec3
+                        {
+                            easting = newGuideList[newGuideList.Count - 1].easting + (Math.Sin(newGuideList[newGuideList.Count - 1].heading) * 2000.0),
+                            northing = newGuideList[newGuideList.Count - 1].northing + (Math.Cos(newGuideList[newGuideList.Count - 1].heading) * 2000.0)
+                        };
+                        newGuideList.Add(endA);
```
FantasyTeddy: Isn't that the same as line 587-599? Could we create a separate method for that to not duplicate the code?

FantasyTeddy (changes requested):

FantasyTeddy (approved):

PR #1072, opened 2025-11-28
Better zoom pan in boundary tool form

Pat-I:
(Small) changes in the Boundary Tool Form. The zoom an Pan buttons are improved and  always usable now.
image
So you can zoom at all steps to add points.
I have tested quite extensively, I thing it works real good.

### DIFF
--- a/SourceCode/GPS/Forms/Field/FormBndTool.cs
+++ b/SourceCode/GPS/Forms/Field/FormBndTool.cs
@@ -31,6 +31,7 @@ namespace AgOpenGPS
         private int bndSelect = 0, smPtsChoose = 1, smPts = 4;
 
         private double zoom = 1, sX = 0, sY = 0;
+        private const double panStep = 0.15;
 
         public List<vec3> secList = new List<vec3>();
         public List<vec3> bndList = new List<vec3>();
@@ -767,35 +768,35 @@ namespace AgOpenGPS
             if (zoom > 1) zoom = 1;
         }
 
+        private void btnZoomIn_Click(object sender, EventArgs e)
+        {
+            zoom *= 0.5;
+            if (zoom < 0.015625) zoom = 0.015625;
+        }
+
         private void btnMoveDn_Click(object sender, EventArgs e)
         {
-            sY += 0.15 * zoom;
+            sY += panStep * zoom;
         }
 
         private void btnMoveUp_Click(object sender, EventArgs e)
         {
-            sY -= 0.15 * zoom;
+            sY -= panStep * zoom;
+        }
+
+        private void btnMoveLeft_Click(object sender, EventArgs e)
+        {
+            sX += panStep * zoom;
+        }
+
+        private void btnMoveRight_Click(object sender, EventArgs e)
+        {
+            sX -= panStep * zoom;
         }
 
         private void btnResetReduce_Click_1(object sender, EventArgs e)
         {
             Reset();
-        }
-
-        private void btnMoveLeft_Click(object sender, EventArgs e)
-        {
-            sX += 0.15 * zoom;
-        }
-
-        private void btnMoveRight_Click(object sender, EventArgs e)
-        {
-            sX -= 0.15 * zoom;
-        }
-
-        private void btnZoomIn_Click(object sender, EventArgs e)
-        {
-            zoom *= 0.5;
-            if (zoom < 0.015625) zoom = 0.015625;
         }
 
         private void btnSlice_Click(object sender, EventArgs e)


Pat-I:
I should add that the green check mark quite annoy me. Because it save nothing after pressing build your temped to click it to save but you have to press "Save" instead.

It should be changed for either a red X or a return arrow IMO 
image
Or even better: A red X until the save button is pressed THEN the green check mark.
But I wont do it. If @richardklasens you think it's a good idea maybe assign someone.

SourceCode/GPS/Forms/Field/FormBndTool.cs
```diff
@@ -778,20 +763,18 @@ private void cboxPointDistance_SelectedIndexChanged(object sender, EventArgs e)
 
         private void btnZoomOut_Click(object sender, EventArgs e)
         {
-            zoom += 0.1;
+            zoom *= 2;
```
J005t67, line 767: Move this method closer to the ZoomIn method

SourceCode/GPS/Forms/Field/FormBndTool.cs
```diff
@@ -778,20 +763,18 @@ private void cboxPointDistance_SelectedIndexChanged(object sender, EventArgs e)
...
             if (zoom > 1) zoom = 1;
         }
 
         private void btnMoveDn_Click(object sender, EventArgs e)
         {
-            if (zoom == 0.1)
-                sY += 0.01;
+            sY += 0.15 * zoom;
```
J005t67: I would prefer to use one constant named panStep in steadof a hardcoded value of 0.15 on four different places

J005t67 (commented):
Only a few minor comments

J005t67:
> I should add that the green check mark quite annoy me. Because it save nothing after pressing build your temped to click it to save but you have to press "Save" instead.
> 
> It should be changed for either a red X or a return arrow IMO Or even better: A red X until the save button is pressed THEN the green check mark. But I wont do it. If @richardklasens you think it's a good idea maybe assign someone.



> I should add that the green check mark quite annoy me. Because it save nothing after pressing build your temped to click it to save but you have to press "Save" instead.

I understand that the current panel is confusing, but Both options you mention are not harmonized with the other panels, so we should not implement like that. I would prefer the commonly used red cross next to the green checkmark. The red cross cancels the green checkmark saves and closes the panel.

SourceCode/GPS/Forms/Field/FormBndTool.cs
Pat-I: I made some changes but messed all up by trying other stuff before commiting. I will try later again.

You would use "private const double panStep = 0.15;"? Because zoom is a double. Now you see my low coding skill :smile:

Pat-I:
@J005t67 if you are happy with the "panStep" addition I think this could be merged.

The same changes are also in PR #1078 but it's maybe better to add here in case 1078 would be dropped or need to be reverted.

The functionality of the two are not related at all.
Just saying.

richardklasens:
> @J005t67 if you are happy with the "panStep" addition I think this could be merged.
> 
> The same changes are also in PR #1078 but it's maybe better to add here in case 1078 would be dropped or need to be reverted.
> 
> The functionality of the two are not related at all. Just saying.

That’s not how Git works..
If the addition is already in another PR, it shouldn’t be a part of this PR. That will result in merge conflicts

Pat-I:
> That’s not how Git works..

Ok so I would propose **to reject** this PR. these two commits are the first two of PR #1078 

If something is really wrong in subsequent commits you could remove them and keep just the first two of a PR?
Anyway it would not be a big deal, this commit is only a few line.

Pat-I:
Closed to avoid future comflic with PR #1078

FantasyTeddy:
> That’s not how Git works.. If the addition is already in another PR, it shouldn’t be a part of this PR. That will result in merge conflicts

I'm not sure that this statement is correct. If the commits are identical, there shouldn't be an issue with conflicts in the other PR.
I would strongly suggest to reopen this PR and merge it. That way the other PR stays smaller and more focused.

richardklasens:
> > That’s not how Git works.. If the addition is already in another PR, it shouldn’t be a part of this PR. That will result in merge conflicts
> 
> I'm not sure that this statement is correct. If the commits are identical, there shouldn't be an issue with conflicts in the other PR. I would strongly suggest to reopen this PR and merge it. That way the other PR stays smaller and more focused.

I understand but the changes are already in the other PR. Means they will be merged when that PR is approved and pulled. So closing this one is fine for now.

Pat-I:
@richardklasens I reopen it.
Because
1- it should be added regardless what will happen with #1078 
2- it target two completely different issues.
3- The commits are identicals

PR #1077, opened 2025-11-29
BndTool defaults fix

farmerbriantee:
Spacing and Smooth set by index not value.


FantasyTeddy (approved):
Just out of curiosity: why are these even settings? Is that really something you want to keep when opening the form again?

Pat-I:
Some people (like me) want 1 as spacing or an other value than 4 for smooth. Having it as saved setting is nice IMO

Issue #1075, opened 2025-11-29
Better behaviour from the "build from boundary" form


Pat-I:
### Problem or Use Case

The "build from boundary" form had a modification in PR #1067 but it still needs huge improvements.

Less steps is not always better.

### Proposed Solution

<img width="1010" height="735" alt="Image" src="https://github.com/user-attachments/assets/0168496a-0792-4005-8e78-1a04d4c38638" />

First, it should not edit the current boundary or headland list. It should work with a temporary list (data)

The form should have a cancel button and an OK button (grayed out until a new boundary in ready to use).
At any point, the cancel button would discard any work done in the form keeping the previous boundary.
One a usable boundary would be drawn the OK button would become active, the user can now chose the new one or keep the old one.

2- The zoom and Pan buttons should always be active, like in PR #1072 

3- the workflow:

- User modify the spacing if desired
- User press "Build, the temporary boundary is now displayed **with** the displayed smoothing
- At this point, the user can change the smooth, **it is recalculated at each selection**
- When happy he press Save. The OK button is now active.
- The user can now edit the boundary, Close and discard, Close and accept the new one. At any point he can also press "reset", this will bring it a the beginning to the spacing choice

This way the user should **never** lose his current boundary if he press some button

### Priority and Impact

The big impact: currently an user can lose his current boundary without warning just by pressing the "reset" button: there is no warning message anymore

farmerbriantee:
From the comments in international forum most people use it the same way. They do the outside round, use 2m and smooth 4, hit start and save and continue on with their work. I know that's how i use it as well and reflects the current setup. 

Certainly can also be improved. Everything can. 

Actually there is a problem with default settings. In Settings.cs line 256/7 the values should be 1, 1 not 2,4

Pat-I:
@farmerbriantee 
`At any point, the cancel button would discard any work done in the form keeping the previous boundary.`
Probably just change it to work on tempory data, keeping the actual one until the green check mark is pressed.

On complex boundary some would try to improve it but can messup. Being able to just discard all by pressing the red arrow would be a real nice feature.
<img width="149" height="69" alt="Image" src="https://github.com/user-attachments/assets/eee47b62-3b79-4eac-b2c8-4ef47cb94bde" />

The green check mark would be active if a boundary is already there or been drawn.

All the other steps I wrote could be discarded, it's just my preference, probably not how a "common" user would use it.

Pat-I:
[...]

PR #1078, opened 2025-11-30
Boundary form works on temporary boundary

Pat-I:
This all to go back to your existing boundary in case you mess up with your attemp to modify the boundary.

It work well but will probably need some changes to be considered as well written code.

Especialy, I use a tempBndList derived from bndList which is maybe not what a coder would want to see.
@J005t67 
@FantasyTeddy


Pat-I:
The two first commits are the same than PR #1072

SourceCode/GPS/Forms/Field/FormBndTool.cs
```diff
@@ -614,6 +645,63 @@ private void BuildBnd()
...
+                for (int j = 0; j < mf.bnd.tempBndList[i].fenceLine.Count; j++)
+                {
+                    b.fenceLine.Add(mf.bnd.tempBndList[i].fenceLine[j]);
+                }
+                // Compute area and ear
+                b.CalculateFenceArea(result.Count);
+                if (b.fenceLineEar != null) b.fenceLineEar.Clear();
+
+                double delta = 0;
+                for (int k = 0; k < b.fenceLine.Count; k++)
```
J005t67, line 669: I see lots of duplicate code. We should realy stop adding more and more duplicate code. Please put duplicate parts in a separate function. This will also help to bring down the nesting level and make the code more readable.

SourceCode/GPS/Forms/Field/FormBndTool.cs
```diff
@@ -614,6 +645,63 @@ private void BuildBnd()
...
+        {
+            secList?.Clear();
+
+            DeleteBoundary();
+
+            var result = new List<CBoundaryList>();
+            for (int i = 0; i < mf.bnd.tempBndList.Count; i++)
+            {
+                var b = new CBoundaryList();
+                b.isDriveThru = mf.bnd.tempBndList[i].isDriveThru;
```
J005t67, line 658: Do not repeat the phrase 'mf.bnd.tempBndList[i]' so many times. Asign it to a variable and reuse that variable.

SourceCode/GPS/Forms/Field/FormBndTool.cs
```diff
@@ -954,14 +1064,14 @@ private void oglSelf_MouseDown(object sender, MouseEventArgs e)
...
                 {
-                    for (int j = 0; j < mf.bnd.bndList.Count; j++)
+                    for (int j = 0; j < mf.bnd.tempBndList.Count; j++)
                     {
-                        for (int i = 0; i < mf.bnd.bndList[j].fenceLine.Count; i++)
+                        for (int i = 0; i < mf.bnd.tempBndList[j].fenceLine.Count; i++)
                         {
-                            double dist = ((pint.easting - mf.bnd.bndList[j].fenceLine[i].easting) * (pint.easting - mf.bnd.bndList[j].fenceLine[i].easting))
-                                            + ((pint.northing - mf.bnd.bndList[j].fenceLine[i].northing) * (pint.northing - mf.bnd.bndList[j].fenceLine[i].northing));
+                            double dist = ((pint.easting - mf.bnd.tempBndList[j].fenceLine[i].easting) * (pint.easting - mf.bnd.tempBndList[j].fenceLine[i].easting))
```
J005t67, line 1073: Do not repeat the part             mf.bnd.tempBndList[j] over and over again. This is very error prone and hard to read. It is escpecially combersome and risky because a very similar name mf.bnd.BndList[j] also exist.

J005t67, line 1073: Do not write expressions that manipulate the easting and northing parts separately. We have GeoCoord, GeoDelta and GeoDir available that can be used instead. In this particular case you should use one of the many distance functions in glm.

SourceCode/GPS/Forms/Field/FormBndTool.cs
```diff
@@ -1043,22 +1153,37 @@ private void oglSelf_Paint(object sender, PaintEventArgs e)
...
-                for (int i = 0; i < mf.bnd.bndList[0].fenceLine.Count; i++)
+                for (int i = 0; i < mf.bnd.tempBndList[0].fenceLine.Count; i++)
                 {
-                    GL.Vertex3(mf.bnd.bndList[0].fenceLine[i].easting, mf.bnd.bndList[0].fenceLine[i].northing, 0);
+                    GL.Vertex3(mf.bnd.tempBndList[0].fenceLine[i].easting, mf.bnd.tempBndList[0].fenceLine[i].northing, 0);
                 }
                 GL.End();
 
                 GL.PointSize(4);
                 GL.Begin(PrimitiveType.Points);
```
J005t67, line 1166: Also in this case, there is no need to write code that mentions the easting and northing part individually. Use GLW.DrawPointsPrimitive() to replace the for loop. It will help to reduce the number of direct dependencies to the OpenTK toolkit, and as such it will make AgOpenGps more portable in the future. Also it will help to enhance performance (in the future), because I will optimize GLW.DrawPointsPrimitive() in the future.

SourceCode/GPS/Forms/Field/FormBndTool.cs
```diff
@@ -1043,22 +1153,37 @@ private void oglSelf_Paint(object sender, PaintEventArgs e)
...
-                for (int i = 0; i < mf.bnd.bndList[0].fenceLine.Count; i++)
+                for (int i = 0; i < mf.bnd.tempBndList[0].fenceLine.Count; i++)
                 {
-                    GL.Vertex3(mf.bnd.bndList[0].fenceLine[i].easting, mf.bnd.bndList[0].fenceLine[i].northing, 0);
+                    GL.Vertex3(mf.bnd.tempBndList[0].fenceLine[i].easting, mf.bnd.tempBndList[0].fenceLine[i].northing, 0);
```
J005t67, line 1169: No need to pass zeros around (third argument.) Please use GLW.PrawPointsPrimitve in stead to replace the whole loop

SourceCode/GPS/Forms/Field/FormBndTool.cs
```diff
@@ -1043,22 +1153,37 @@ private void oglSelf_Paint(object sender, PaintEventArgs e)
...
                 }
                 GL.End();
+
+                if (mf.bnd.tempBndList.Count > 1)
```
J005t67, line 1173: No need to test for the Count > 1.  The for loop already does that before its first iteration.

SourceCode/GPS/Forms/Field/FormBndTool.cs
```diff
@@ -1043,22 +1153,37 @@ private void oglSelf_Paint(object sender, PaintEventArgs e)
...
+                {
+                    for (int i = 1; i < mf.bnd.tempBndList.Count; i++)
+                    {
+                        GL.Color3(0.95f, 0.95f, 0.35f);
+                        GL.Begin(PrimitiveType.LineLoop); //Todo: change color if possible
```
J005t67, line 1178: Replace for loop with GLW.DrawLineLoopPrimitive()

SourceCode/GPS/Forms/Field/FormBndTool.cs
```diff
@@ -1130,9 +1255,9 @@ private void oglSelf_Paint(object sender, PaintEventArgs e)
                         GL.Color3(0.90f, 0.5f, 0.25f);
                         GL.Begin(PrimitiveType.LineStrip);
```
J005t67, line 1256: Use GLW.DrawLineStripPrimitive() to replace whole loop with one call

SourceCode/GPS/Forms/Field/FormBndTool.cs
```diff
@@ -1160,21 +1285,21 @@ private void DrawABTouchPoints()
             GL.PointSize(24);
             GL.Begin(PrimitiveType.Points);
 
-            if (mf.bnd.bndList.Count != 0)
+            if (mf.bnd.tempBndList.Count != 0)
             {
                 GL.Color3(0, 0, 0);
-                if (start != 99999) GL.Vertex3(mf.bnd.bndList[bndSelect].fenceLine[start].easting, mf.bnd.bndList[bndSelect].fenceLine[start].northing, 0);
-                if (end != 99999) GL.Vertex3(mf.bnd.bndList[bndSelect].fenceLine[end].easting, mf.bnd.bndList[bndSelect].fenceLine[end].northing, 0);
+                if (start != 99999) GL.Vertex3(mf.bnd.tempBndList[bndSelect].fenceLine[start].easting, mf.bnd.tempBndList[bndSelect].fenceLine[start].northing, 0);
```
J005t67, line 1291: Do not reapeat the phrase mf.bnd.tempBndList[bndSelect] so many times. Assign it to a variable once and reuse that variable.

SourceCode/GPS/Forms/Field/FormBndTool.cs
```diff
@@ -1160,21 +1285,21 @@ private void DrawABTouchPoints()
             GL.PointSize(24);
             GL.Begin(PrimitiveType.Points);
```
J005t67, line 1286: Do not put the GL.Begin and GL.End on a different indenation level. In this case you can easily solve it by using one of the Draw*Primitive functions in GLW.

SourceCode/GPS/Forms/Field/FormBndTool.cs
```diff
@@ -985,13 +1095,13 @@ private void oglSelf_MouseDown(object sender, MouseEventArgs e)
...
+                if (mf.bnd.tempBndList.Count != 0)
                 {
 
-                    for (int i = 0; i < mf.bnd.bndList[j].fenceLine.Count; i++)
+                    for (int i = 0; i < mf.bnd.tempBndList[j].fenceLine.Count; i++)
                     {
-                        double dist = ((pint.easting - mf.bnd.bndList[j].fenceLine[i].easting) * (pint.easting - mf.bnd.bndList[j].fenceLine[i].easting))
-                                        + ((pint.northing - mf.bnd.bndList[j].fenceLine[i].northing) * (pint.northing - mf.bnd.bndList[j].fenceLine[i].northing));
+                        double dist = ((pint.easting - mf.bnd.tempBndList[j].fenceLine[i].easting) * (pint.easting - mf.bnd.tempBndList[j].fenceLine[i].easting))
+                                        + ((pint.northing - mf.bnd.tempBndList[j].fenceLine[i].northing) * (pint.northing - mf.bnd.tempBndList[j].fenceLine[i].northing));
```
J005t67, line 1104: Use a proper distance function (distance squared in this case) to replace this lengthy, risky and hard to read expression.
I will extent GeoLineSegment with a Length and LengthSquared property, so if you wait a few days, you could write something like:

                            GeoLineSegment segment = GeoRefactorHelper.GetLineSegment(mf.bnd.bndList[j].fenceLine, i);
                            double distanceSquared = segment.LengthSquared;

SourceCode/GPS/Forms/Field/FormBndTool.cs
J005t67, line 1169: No need to repeat the same variable 3 times. Please use GLW.DrawPointsPrimitives here.

SourceCode/GPS/Classes/CBoundary.cs
```diff
@@ -11,6 +11,7 @@ public partial class CBoundary
 
 
         public List<CBoundaryList> bndList = new List<CBoundaryList>();
+        public List<CBoundaryList> tempBndList = new List<CBoundaryList>();
```
FantasyTeddy, line 14: Why is the temporary list a property of the `CBoundary` class? Shouldn't that just be a variable inside the `FormBndTool`?

Pat-I: I thought I removed it. It should not be used

Pat-I:
@J005t67 Thank you for your inputs, indeed I think this should all be applied.
But I have some more basic questions first:

1- The goal is to not modify any data from mf.bnd.bndList before the user quit the form by clicking the green check mark.

So the simplest for me was to create tempBndList in CBoundary.cs
**Does someone want to create something more local?** It is only used while the form is open.

Note: I have just seen that "Build from AB line" could also benefit from a temporary storage. Should the related functions (point 2 and 3 here) be stored outside "FormBndTool" to be reused?

2-When the form open I want a "complete copy from bndList to tempBndList.
**Are the lines 119 to 152 in FormBndTool.cs OK?**
Because in the construtor in CBounday.cs line 17 there are variables turnSelected, isHeadlandOn, isSectionControlledByHeadland; I don't know if they are (or need to be) copied over.

Same for fenceLine. "isDriveThru" is copied but I don't know about "area"
fenceLineEar is recalculated, which I don't know if it's good, could it be copied over? I don't know about "hdLine" and "turnLine" from CBoundaryList.cs line 16

Note that I copied this code from the boundary file reader in SaveOpen.

**Headlands:**
From my testing headland are removed (probably not copied over) when a modification is made to any boundary (outer or inner).
So some stuff to study:
**Should the headland for untouched boundary be carried over?** 
Hint: The older version just seemed to no touch the headland pts even if the boundaries are modified.

Or could we rebuild the headland automatically? I don't think so because the parameters (how many meter from which side) are not saved.

3-Copy back: Save and exit.
Fonction "BuildBnd()" in FormBndTool.cs
Here again some copy/paste from me.
**A review is needed about the "CalcutateFenceArea" function:** it is used 2 times, probably just need once

When a new outer boundary is draw it delete all "bndList" data so no problem there.

**When you press the OK button without changing anything the headlands are lost: to repair!**

When you press the OK with modified boundary pts the best solution (for now) would be to just copy back the headland. 

Clonclusion:
Where to put those functions? (could be reused in others forms) and resolve the headland problem (copy it over each list) before thinking merging it.

@J005t67 @FantasyTeddy I will wait for your inputs and/or if one of you want to work on it.

SourceCode/GPS/Classes/CBoundary.cs
Pat-I: Oh you added this line? Because I tried exactly this at the beginning.
The reason I did't used this:
You can modify points in any "fenceLine" List that are in the "bndList" so I think you really need to copy over the whole thing.

Se my comment below about this and how it should be made. I have also discovered that the "build from ABline form" could also benefit from a temporary copy.

Pat-I: Aie. You can now see my low coding skills. 

I did not saw that this comment is in CBoundary.cs, I thought it was in FormBndTool, where I tried Just a  simple list. But off course this didn't worked.

Pat-I:
I'm closing it now because I want a complete rewrite.
I'm also letting @J005t67 do some refactor before working on it again. I will continue to discussion on issue #1075

PR #1084, opened 2025-12-02
Refactor AgShare download to use static IO classes

richardklasens:
- Replace LocalFieldModel/LocalPoint/AbLineLocal with ParsedField using domain types (CBoundaryList, CTrk) directly
- FieldFileWriter now uses BoundaryFiles.Save(), TrackFiles.Save(), FieldPlaneFiles.Save() instead of manual file writing
- Parser creates domain types directly, eliminating intermediate conversion step
- Remove unused BoundaryHelper.WithHeadings() from GeoConverter
- Update FormAgShareDownloader to render using domain types

### DIFF
--- a/SourceCode/AgOpenGPS.Core/DrawLib/GLW.GeoCoordPrimitives.cs
+++ b/SourceCode/AgOpenGPS.Core/DrawLib/GLW.GeoCoordPrimitives.cs
@@ -18,7 +18,6 @@ namespace AgOpenGPS.Core.DrawLib
         {
             DrawPrimitive(PrimitiveType.Lines, vertices);
         }
-
         public static void DrawLineLoopPrimitive(GeoCoord[] vertices)
         {
             DrawPrimitive(PrimitiveType.LineLoop, vertices);
@@ -29,9 +28,25 @@ namespace AgOpenGPS.Core.DrawLib
             DrawPrimitive(PrimitiveType.LineStrip, vertices);
         }
 
+        public static void DrawPointsPrimitive(GeoCoord[] vertices)
+        {
+            DrawPrimitive(PrimitiveType.Points, vertices);
+        }
+
         public static void DrawTriangleFanPrimitive(GeoCoord[] vertices)
         {
             DrawPrimitive(PrimitiveType.TriangleFan, vertices);
+        }
+
+        public static void DrawPointLayered(
+            PointStyle[] pointStyles,
+            GeoCoord coord)
+        {
+            foreach (PointStyle pointStyle in pointStyles)
+            {
+                SetPointStyle(pointStyle);
+                DrawPoint(coord);
+            }
         }
 
         public static void DrawLinesPrimitiveLayered(
--- a/SourceCode/AgOpenGPS.Core/Translations/gStr.ja.resx
+++ b/SourceCode/AgOpenGPS.Core/Translations/gStr.ja.resx
@@ -59,7 +59,7 @@
             : using a System.ComponentModel.TypeConverter
             : and then encoded with base64 encoding.
     -->
-  <xsd:schema id="root" xmlns="" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:msdata="urn:schemas-microsoft-com:xml-msdata">
+  <xsd:schema xmlns="" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:msdata="urn:schemas-microsoft-com:xml-msdata" id="root">
     <xsd:import namespace="http://www.w3.org/XML/1998/namespace" />
     <xsd:element name="root" msdata:IsDataSet="true">
       <xsd:complexType>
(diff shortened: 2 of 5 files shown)


SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
```diff
@@ -1,23 +1,24 @@
...
+    /// <summary>
+    /// Parses AgShare field DTOs into domain types ready for file writing.
+    /// </summary>
     public static class AgShareFieldParser
     {
-        /// <summary>
-        /// Validates if coordinates are within valid WGS84 ranges
-        /// </summary>
+        // Validates if coordinates are within valid WGS84 ranges
         private static bool IsValidCoordinate(double latitude, double longitude)
```
J005t67: Move this function the Wgs84. That way, you don't take the risk that th earguments are passed in the wrong order

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
```diff
@@ -29,13 +30,11 @@ public static LocalFieldModel Parse(AgShareFieldDto dto)
                 throw new ArgumentException("Field name cannot be null or empty", nameof(dto));
```
J005t67: The text in the exception does not match the condition.

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
```diff
@@ -44,134 +43,119 @@ public static LocalFieldModel Parse(AgShareFieldDto dto)
...
-                        Name = ab.Name ?? "Unnamed",
-                        Heading = heading,
-                        PtA = new LocalPoint(vA.Easting, vA.Northing),
-                        PtB = new LocalPoint(vB.Easting, vB.Northing),
-                        CurvePoints = new List<LocalPoint>()
+                        name = ab.Name ?? "Unnamed",
+                        mode = isCurve ? TrackMode.Curve : TrackMode.AB,
+                        ptA = new vec2(vA.Easting, vA.Northing),
+                        ptB = new vec2(vB.Easting, vB.Northing),
+                        heading = GeoConverter.HeadingFromPoints(vA, vB),
```
J005t67: Remove the function HeadingFromPoints and use

GeoDir(GeoCoord fromCoord, GeoCoord toCoord) and AngleInRadians

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
```diff
@@ -44,134 +43,119 @@ public static LocalFieldModel Parse(AgShareFieldDto dto)
...
+                    var trk = new CTrk
                     {
-                        Name = ab.Name ?? "Unnamed",
-                        Heading = heading,
-                        PtA = new LocalPoint(vA.Easting, vA.Northing),
-                        PtB = new LocalPoint(vB.Easting, vB.Northing),
-                        CurvePoints = new List<LocalPoint>()
+                        name = ab.Name ?? "Unnamed",
+                        mode = isCurve ? TrackMode.Curve : TrackMode.AB,
+                        ptA = new vec2(vA.Easting, vA.Northing),
```
J005t67: I know that we have more than 1200 lines of code currently, that mention the easting and northing part individually, but we should stop doing it that way. It bloats the code size and every time, you take the (tiny) risk of mixing up the order of the arguments. Use a GeoCoord and use a function ToGeoCoord() that does the conversion to GeoCoord for you.

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
```diff
@@ -44,134 +43,119 @@ public static LocalFieldModel Parse(AgShareFieldDto dto)
...
                                     var nextLocal = converter.ToLocal(next.Latitude, next.Longitude);
-
-                                    // Correct volgorde: Northing, Easting
-                                    var localCoord = new GeoCoord(local.Northing, local.Easting);
-                                    var nextCoord = new GeoCoord(nextLocal.Northing, nextLocal.Easting);
-                                    localHeading = new GeoDir(new GeoDelta(localCoord, nextCoord)).AngleInRadians;
+                                    var delta = new GeoDelta(
+                                        new GeoCoord(local.Northing, local.Easting),
+                                        new GeoCoord(nextLocal.Northing, nextLocal.Easting));
+                                    heading = new GeoDir(delta).AngleInRadians;
```
J005t67: GeoDir has improved a bit. You can now use a constructor that takes two GeoCoord, so you don't need the GeoDelta anymore.

public GeoDir(GeoCoord fromCoord, GeoCoord toCoord)

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
```diff
@@ -44,134 +43,119 @@ public static LocalFieldModel Parse(AgShareFieldDto dto)
...
                             }
 
-                            abLine.CurvePoints.Add(new LocalPoint(local.Easting, local.Northing, localHeading));
+                            trk.curvePts.Add(new vec3(local.Easting, local.Northing, heading));
+                        }
+
+                        // Update ptA/ptB to first/last curve points
+                        if (trk.curvePts.Count >= 2)
+                        {
+                            trk.ptA = new vec2(trk.curvePts[0].easting, trk.curvePts[0].northing);
```
J005t67: Use the method ToVec2() in stead of passing the arguments individually

SourceCode/GPS/Forms/Field/FormAgShareDownloader.cs
```diff
@@ -273,41 +273,42 @@ private void RenderField(LocalFieldModel field)
...
             if (boundaries != null)
             {
-                foreach (var ring in boundaries)
+                foreach (var bnd in boundaries)
                 {
-                    foreach (var pt in ring)
+                    foreach (var pt in bnd.fenceLine)
                     {
-                        bb.Include(pt.ToGeoCoord());
+                        bb.Include(new GeoCoord(pt.northing, pt.easting));
```
J005t67: Use ToGeoCoord() for the conversion

SourceCode/GPS/Forms/Field/FormAgShareDownloader.cs
```diff
@@ -273,41 +273,42 @@ private void RenderField(LocalFieldModel field)
...
+                foreach (var trk in tracks)
                 {
-                    if (ab.CurvePoints != null && ab.CurvePoints.Count > 0)
+                    if (trk.mode == TrackMode.Curve && trk.curvePts != null && trk.curvePts.Count > 0)
                     {
-                        foreach (var pt in ab.CurvePoints)
+                        foreach (var pt in trk.curvePts)
                         {
-                            bb.Include(pt.ToGeoCoord());
+                            bb.Include(new GeoCoord(pt.northing, pt.easting));
```
J005t67: Use ToGeoCoord()

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
```diff
@@ -78,6 +78,9 @@ public static ParsedField Parse(AgShareFieldDto dto)
 
                     if (bnd.fenceLine.Count >= 3)
                     {
+                        // First ring is outer boundary, subsequent rings are holes (drive-through)
+                        bnd.isDriveThru = boundaryIndex > 0;
```
J005t67: Will this set isDriveThru to true for al inner boundaries? Looks like a bug to me.

richardklasens: Changed to False as default since AgShare doesnt have a property yet for IsDriveThru

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
```diff
@@ -27,10 +20,11 @@ public static ParsedField Parse(AgShareFieldDto dto)
 
             if (string.IsNullOrWhiteSpace(dto.Name))
             {
-                throw new ArgumentException("Field name cannot be null or empty", nameof(dto));
+                throw new ArgumentException("Field name cannot be null, empty or whitespace", nameof(dto));
             }
 
-            if (!IsValidCoordinate(dto.Latitude, dto.Longitude))
+            var origin = new Wgs84(dto.Latitude, dto.Longitude);
```
J005t67, line 26: Let dto use a Wgs84 internally, in stead of indiviual Latitude and Longitude. That way, you don't need to do the conversion to Wgs84 here.

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
```diff
@@ -47,10 +41,10 @@ public static ParsedField Parse(AgShareFieldDto dto)
             {
                 FieldId = dto.Id,
                 Name = dto.Name,
-                Origin = new Wgs84(dto.Latitude, dto.Longitude)
+                Origin = origin
             };
 
-            var converter = new GeoConverter(dto.Latitude, dto.Longitude);
+            var converter = new GeoConverter(origin.Latitude, origin.Longitude);
```
J005t67, line 47: Let the GeoConverter take a Wgs84 as contructor argument. It avoids the risk of passing the arguments in the wrong order.

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
```diff
@@ -66,20 +60,22 @@ public static ParsedField Parse(AgShareFieldDto dto)
...
-                        if (!IsValidCoordinate(point.Latitude, point.Longitude))
+                        var wgs = new Wgs84(point.Latitude, point.Longitude);
+                        if (!wgs.IsValid)
                         {
                             Log.EventWriter($"[AgShare] Skipping invalid boundary coordinate: Lat={point.Latitude}, Lon={point.Longitude}");
                             continue;
                         }
 
-                        var local = converter.ToLocal(point.Latitude, point.Longitude);
+                        var local = converter.ToLocal(wgs.Latitude, wgs.Longitude);
```
J005t67, line 70: Let the function ToLocal take a Wgs84 as argument

J005t67, line 70: Replace 'var' with the explicit type here. In this specific case (conversions) it enhances readability, otherwise I would not care much.

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
richardklasens: GeoConverter isn’t a part of this PR

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
richardklasens: Same here

richardklasens: To me it’s better to begin with Var.. mean there is a local variable.

SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
richardklasens: Not sure if that is really the thing. AgShare comms is an external project from outside so I won’t lean on internal stuff in this case

PR #1094, opened 2025-12-06
Add method Include(GeoBoundingBox) to GeoBoundingBox

J005t67:
The new method
Include(GeoBoundingBox bb) 

is a small preparation step to start using QuadStrip in the future. QuadStrip keeps track of its own GeoBoundingBox and will help to simplify the calulation of the bounding box of the worked area. It also might speed up rendering, because it could help to skip rendering  of QuadStrip that are outside the Viewport

### DIFF
--- a/SourceCode/AgOpenGPS.Core.Tests/Models/Base/GeoBoundingBoxTests.cs
+++ b/SourceCode/AgOpenGPS.Core.Tests/Models/Base/GeoBoundingBoxTests.cs
@@ -20,5 +20,76 @@ namespace AgOpenGPS.Core.Tests.Models
             Assert.That(centerCoord, Is.EqualTo(correctCenterCoord));
         }
 
+        [Test]
+        public void Test_IncludeBoundingBox_NotEmpty_NotEmpty()
+        {
+            GeoCoord coordA1 = new GeoCoord(-10.0, 20.0);
+            GeoCoord coordA2 = new GeoCoord(-30.0, 40.0);
+            GeoBoundingBox bbA = GeoBoundingBox.CreateEmpty();
+            bbA.Include(coordA1);
+            bbA.Include(coordA2);
+
+            GeoCoord coordB1 = new GeoCoord(-20.0, 30.0);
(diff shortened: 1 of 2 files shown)


SourceCode/AgOpenGPS.Core.Tests/Models/Base/GeoBoundingBoxTests.cs
```diff
@@ -20,5 +20,76 @@ public void Test_CenterCoord()
             Assert.That(centerCoord, Is.EqualTo(correctCenterCoord));
         }
 
+        [Test]
+        public void Test_IncludeBoundingBox_NotEmpty_NotEmpty()
```
FantasyTeddy, line 24: Maybe we should put some conventions for test naming into place? For me it is not obvious, what "NotEmpty" refers to for example.

SourceCode/AgOpenGPS.Core.Tests/Models/Base/GeoBoundingBoxTests.cs
```diff
@@ -20,5 +20,76 @@ public void Test_CenterCoord()
...
+            GeoBoundingBox bbA = GeoBoundingBox.CreateEmpty();
+            bbA.Include(coordA1);
+            bbA.Include(coordA2);
+
+            GeoCoord coordB1 = new GeoCoord(-20.0, 30.0);
+            GeoCoord coordB2 = new GeoCoord(-40.0, 50.0);
+            GeoBoundingBox bbB = GeoBoundingBox.CreateEmpty();
+            bbA.Include(coordB1);
+            bbA.Include(coordB2);
+            bbA.Include(bbB);
```
FantasyTeddy, line 37: We should seriously consider to have some separate "Arrange", "Act", "Assert" sections, for example with comments:
```csharp
// Arrange
...
bbA.Include(coordB1);
bbA.Include(coordB2);

// Act
bbA.Include(bbB);

// Assert
Assert.That(bbA.IsEmpty, Is.False);
...
```
This would help to see which step is actually being tested. (Here it is the `Include(GeoBoundingBox)` method, but that is not very obvious)

PR #1108, opened 2025-12-11
Performance 2: Use GL.SetColor(Byte[]) in stead of GL.SetColor(byte, byte, byte, byte)

J005t67:
The assmblycode of the old code that calls GL.SetColor(byte, byte, byte, byte), needs 8 instructions to set 4 different register before calling,

Most likely, the new code only needs 2 instruction to set 1 register before calling GL.SetColor(Byte[]).
I wanted to check this in the assembly code by setting a breakpoint in GLW.SetColor(), but the breakpoint was not hit anymore.
This means that the method GLW.SetColor() has become so small that the compiler inlines it, which is even better for performance.

### DIFF
--- a/SourceCode/GPS/Classes/CABCurve.cs
+++ b/SourceCode/GPS/Classes/CABCurve.cs
@@ -1020,15 +1020,18 @@ namespace AgOpenGPS
                 GL.Color3(0.95f, 0.42f, 0.750f);
                 GL.LineWidth(4.0f);
                 GL.Begin(PrimitiveType.LineStrip);
-                for (int h = 0; h < desList.Count; h++) GL.Vertex3(desList[h].easting, desList[h].northing, 0);
+                for (int h = 0; h < desList.Count; h++)
+                {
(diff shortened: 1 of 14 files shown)


SourceCode/AgOpenGPS.Core/Models/Base/ColorRgba.cs
```diff
@@ -27,16 +21,35 @@ public ColorRgba(float red, float green, float blue, float alpha = 1.0f)
...
-        public byte Green { get; }
-        public byte Blue { get; }
-        public byte Alpha { get; }
+        // For better performance in GLW.SetColor()
+        public byte[] ByteArray { get; private set; }
+
+        public byte Red
+        {
+            get { return ByteArray[0]; }
+            set { ByteArray[0] = value; }
```
FantasyTeddy, line 33: Do we even need the setters here? Shouldn't we treat the `ColorRgba` more like some immutable data structure?

PR #1114, opened 2025-12-16
Refactor: reduce duplicate code with new class Viewport

J005t67:
The new Viewport class achieves better separation of concerns between the viewport related functionality and the functionality of the form that contains the viewport. (In other words: the forms become a lot more simple.)
Most of the code in the Viewport class is simply copied from the original code of FormBndTool. The three biggest differences:
1) The viewports can be used to display any region that can be specified by SetBoundingBox(). (The original code always displayed the field.) Note that this change also remove a dependency to the main form.
2) The original code only took the largest size of the field into account to make the filed fit in the viewport. (Which works well, because most viewports are square or almost square.) The nieuw viewport class takes the aspect ratio of both viewport and boundingBox into account, which will give a better fit if for example both the viewport and BoundingBox have a 'Landscape' aspect ratio.
3) The original code used sX and sY to store the  amount of panning. Both are a fraction of the size of the field. To calculate the center of the viewport, this fraction must be multiplied by the field size and added to the center of the field. The new code, simply stores the amount of panning by storing the GeoCoord that is in the center of the viewport.

### DIFF
--- a/SourceCode/AgOpenGPS.Core/DrawLib/GLW.Viewport.cs
+++ b/SourceCode/AgOpenGPS.Core/DrawLib/GLW.Viewport.cs
@@ -45,7 +45,7 @@ namespace AgOpenGPS.Core.DrawLib
 
         public static void SetClearColor(float red, float green, float blue, float alpha)
         {
-            GL.ClearColor(0.0f, 0.0f, 0.0f, 1.0f);
+            GL.ClearColor(red, green, blue, alpha);
         }
 
         public static void Flush()


SourceCode/AgOpenGPS.Core/DrawLib/GLW.Viewport.cs
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        public static void ClearColorAndDepthBuffer()
+        {
+            GL.Clear(ClearBufferMask.DepthBufferBit | ClearBufferMask.ColorBufferBit);
+        }
+
+        public static void SetClearColor(float red, float green, float blue, float alpha)
+        {
+            GL.ClearColor(0.0f, 0.0f, 0.0f, 1.0f);
```
FantasyTeddy:
```suggestion
            GL.ClearColor(red, green, blue, alpha);
```

J005t67: Thanks again! Fixed it.

PR #1115, opened 2025-12-17
Performance: ColorRgba: add constructor that only needs one UInt32 argument

J005t67:
The new constructor only needs one argument , (so less instructions on the caller side)
Also, it avoids the costly conversion from floats to bytes. (less instructions in the function itself.)
It only needs 4 byte of data, in stead of 3 or 4 times size of float, (12 or 16 bytes). This also reduces code size a bit and so it reduces the pressure on the cache a bit.
None of the above is very important, because this is not where the performance bottlenecks are.

Anyway, I do like the following improvements:
- the comment that says not to recreate the colors repeatedly for each frame.
- the fact that ColorRgba is now more ready to replace the use of System.Windows.Drawing.Color that ties us to windows and makes the current code less portable.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,15 @@
+# These are supported funding model platforms
+
+github: # Replace with up to 4 GitHub Sponsors-enabled usernames e.g., [user1, user2]
+patreon: # Replace with a single Patreon username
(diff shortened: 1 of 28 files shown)


richardklasens:
@codex please review

PR #1126, opened 2025-12-27
Add DTOs for all relevant AgShare queries

FantasyTeddy:


SourceCode/AgOpenGPS.Core/AgShare/Models/GuidanceTrackDto.cs
richardklasens, line 1: Should we rename this to GuidanceTrackDto? It’s more then just AbLine

SourceCode/AgOpenGPS.Core/AgShare/Models/CoordinateDto.cs
richardklasens, line 1: Should we able to use the Core models instead?

SourceCode/AgOpenGPS.Core/AgShare/Models/PolygonDto.cs
richardklasens, line 1: Maybe we should rename “Holes” to “Inner”

SourceCode/GPS/Classes/AgShare/AgShareUploader.cs
```diff
@@ -92,35 +92,32 @@ public async Task UploadAsync(FieldSnapshot snapshot, FormGPS gps)
                     }
                 }
 
-                List<AbLineUploadDto> abLines = ConvertAbLines(snapshot.Tracks, snapshot.Converter);
+                List<AbLineDto> abLines = ConvertAbLines(snapshot.Tracks, snapshot.Converter);
```
richardklasens: If we change to GuidanceTrackDto we need to rename this also

SourceCode/AgOpenGPS.Core/AgShare/Models/GuidanceTrackDto.cs
FantasyTeddy: Good idea, I changed it.

SourceCode/AgOpenGPS.Core/AgShare/Models/CoordinateDto.cs
FantasyTeddy: I don't think we should. Everything that is part of the AgShare API should stay separate and be converted later. That way the internals of AgOpenGPS are not entangled with the AgShare API.

SourceCode/AgOpenGPS.Core/AgShare/Models/PolygonDto.cs
FantasyTeddy: Would make sense, but we can't. "Holes" is the name of the AgShare API that we cannot break (or we will break older clients).

SourceCode/GPS/Classes/AgShare/AgShareUploader.cs
FantasyTeddy: Done

PR #1132, opened 2025-12-28
Add some tests for the `AgShareClient`

FantasyTeddy:
I am not entirely sure about the usefulness of these tests. Please let me know if they should be different.


FantasyTeddy:
I don't think that this is a good way to test this.

PR #1133, opened 2025-12-28
Small AgShare fixes caused by previous refactorings

FantasyTeddy:


SourceCode/GPS/Forms/Field/FormAgShareDownloader.cs
```diff
@@ -48,6 +48,12 @@ private async void FormAgShareDownloader_Load(object sender, EventArgs e)
                 // Get user's own fields from the AgShare server
                 var fields = await downloader.GetOwnFieldsAsync();
 
+                if (fields == null)
+                {
+                    gps.TimedMessageBox(1000, "AgShare", "Failed to load field list.");
```
gunicsba, line 53: Couldn't we try to load these from resources?
would be so nice to have everything in a translatable format

FantasyTeddy: Sure, but maybe in a later PR? (Doesn't really have anything to do with my changes)

richardklasens (approved):

PR #1139, opened 2025-12-31
DrawWorldGrid must take grid rotation into account

J005t67:
The function DrawWorldgrid did not take the grid orientation into account during the computation of the offset of the first and last grid lines.

Also simplified the code a bit by simply starting at the origin

### DIFF
--- a/SourceCode/AgOpenGPS.Core/AgShare/AgShareClient.cs
+++ b/SourceCode/AgOpenGPS.Core/AgShare/AgShareClient.cs
@@ -15,36 +15,33 @@ namespace AgOpenGPS.Core.AgShare
     /// </summary>
     public class AgShareClient
     {
-        private readonly HttpClient _client;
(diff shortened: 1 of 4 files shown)


SourceCode/GPS/Classes/CTram.cs
```diff
@@ -130,6 +130,7 @@ public void DrawTram()
                     }
                     GL.End();
                     GL.Begin(PrimitiveType.LineStrip);
+                    DrawTramInner();
```
FantasyTeddy: Shouldn't that replace the loop below?

J005t67: Oops, this is unrelated code and I should not have commited it. I have reverted it. (I wanted to check te generated assembly code for this loop and inspecting the assembly code becomes more easy if you move it to a separate function. )

PR #1142, opened 2026-01-02
Move part of WorldGrid code to FieldGrid and FieldGridVisual.

J005t67:
Move part of WorldGrid code to FieldGrid and FieldGridVisual to make it reusable.
Let FormGrid show old and new FieldGrid

### DIFF
--- a/SourceCode/AgOpenGPS.Core/Translations/gStr.cs.resx
+++ b/SourceCode/AgOpenGPS.Core/Translations/gStr.cs.resx
@@ -924,7 +924,7 @@ Objednávku, kterou je tlačíte, bude objednávka, která se objeví v menu od 
     <value>Uložit a vrátit</value>
   </data>
   <data name="gsSaveBeerTime" xml:space="preserve">
-    <value>Čas na pivo! Čeština!</value>
+    <value>Uvidíme se příště!</value>
   </data>
   <data name="gsSaveField" xml:space="preserve">
     <value>Ukládám pole</value>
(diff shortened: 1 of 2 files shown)


SourceCode/AgOpenGPS.Core/Models/WorldGrid.cs
```diff
@@ -1,18 +1,17 @@
...
 using OpenTK.Graphics.OpenGL;
 using System;
-using System.Collections.Generic;
 using System.Drawing;
 
 namespace AgOpenGPS.Core
 {
     public class WorldGrid
     {
+        private readonly FieldGridVisual _fieldGridVisual;
```
FantasyTeddy, line 14: Should the `WorldGrid` model really know about the `FieldGridVisual`? I thought the dependency should be the other way around: Visual -> Model

SourceCode/AgOpenGPS.Core/Visuals/Field/FieldGridVisual.cs
```diff
@@ -0,0 +1,69 @@
...
+using AgOpenGPS.Core.Models;
+using System;
+using System.Collections.Generic;
+
+namespace AgOpenGPS.Core.Visuals
+{
+    public class FieldGridVisual
+    {
+        static public readonly ColorRgba FieldGridDayColor = Colors.Gray012;
+        static public readonly ColorRgba FieldGridNightColor = Colors.Gray025;
```
FantasyTeddy: Can we use this order for the modifiers?
```suggestion
        public static readonly ColorRgba FieldGridDayColor = Colors.Gray012;
        public static readonly ColorRgba FieldGridNightColor = Colors.Gray025;
```

SourceCode/GPS/Forms/Guidance/FormGrid.cs
```diff
@@ -101,6 +112,8 @@ private void oglSelf_Paint(object sender, PaintEventArgs e)
                 bool isSelected = j == 0;
                 FenceLineVisual.DrawFenceLine(fenceLineEar, isSelected);
             }
+            _fieldGridVisual.Draw(Colors.White, mf.FieldBoundingBox);
+            _fieldGridCopyVisual.Draw(Colors.Yellow, mf.FieldBoundingBox);
```
FantasyTeddy, line 116: Why do we need to draw both grids?

SourceCode/AgOpenGPS.Core/Visuals/Field/FieldGridVisual.cs
J005t67: Changed. And also changed in Colors.cs where this code fragment was copied from. I will make another small PR to fix al other occurences of 'static public'

SourceCode/AgOpenGPS.Core/Models/WorldGrid.cs
J005t67: Yes! You are exactly right and I also implemented it that way for FieldGrid and FieldGridVisual. WorldGrid, on the other hand is still a bit of everything, that still needs to be split in layers for 'Model-code' and 'Visual-code'. Moving out the FieldGrid part (FieldGrid and FieldGridVisual) is one of the steps to finally split WorldGrid completely in 'Model-code' and 'Visual-code'. By the way: in the end I  want to connect the visuals to the models using the MVVM pattern.

SourceCode/GPS/Forms/Guidance/FormGrid.cs
J005t67: This change is for better usability, and maybe I should have discussed it first in a separate issue or I should have made a separate PR for it.
Anyway:
The white grid shows the situation when the panel was opened. The yellow grid is updated when the user clicks two points or clicks the 'AB-alignement button', (and the white grid remains unchanged)
After pressing 'OK' the yellow grid is accepted. After pressing 'Cancel' the yellow grid is ignored.

FantasyTeddy: I see. Makes sense to me, thanks for explaining.

PR #1145, opened 2026-01-03
SMART Was0 logic

gunicsba:
I had this earlier but that time it didn't make it into the main branch.

Giving it a second try:
image

What it does?
Check for the WAS values and try to figure out where the middle point should be. All you do is open the steering settings and wait. Then it visually tells the offset you want to apply.

### DIFF
--- a/SourceCode/GPS/Forms/Settings/FormSteer.cs
+++ b/SourceCode/GPS/Forms/Settings/FormSteer.cs
@@ -1369,7 +1369,7 @@ namespace AgOpenGPS
                            $"{gStr.gsConfidence}: {stats.ConfidenceLevel:F1}%\n\n" +
                            (stats.HasValidRecommendation ? "✓" : "✗");
 
-            FormDialog.Show(gStr.gsCalibrationDetails, message, MessageBoxButtons.OK);
+            FormDialog.Show(gStr.gsCalibrationDetails, message, DialogSeverity.Info);
 
             // Also log to file
             mf.smartWASCalibration.LogAnalysisReport();


richardklasens:
@codex please review

richardklasens:
I would prefer to create a function for it with a dedicated “activate” button instead of doing it directly while opening the form. Maybe add an automatic Counts per degree estimation function as well together with the Smart WAS.

SourceCode/GPS/Classes/SmartWASCalibration.cs
```diff
@@ -0,0 +1,329 @@
...
+        public DateTime LastCollectionTime { get; private set; }
+
+        // Statistics
+        public double Mean { get; private set; }
+        public double StandardDeviation { get; private set; }
+        public double Median { get; private set; }
+
+        #endregion
+
+        #region Constructor
```
richardklasens, line 45: Maybe split the logic in a separate class so it can be disconnected from the UI. Now everything leads to FormGPS

SourceCode/GPS/Classes/CSmartWASCalibration.cs
```diff
@@ -0,0 +1,329 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace AgOpenGPS
+{
+    /// <summary>
+    /// Smart WAS Calibration system that collects and analyzes guidance line steer angle data
+    /// to determine the optimal WAS zero point based on statistical distribution analysis
+    /// </summary>
+    public class CSmartWASCalibration
```
FantasyTeddy: Please don't use the "C" prefix for new code. We should remove that eventually on all other classes too.

SourceCode/GPS/Classes/CSmartWASCalibration.cs
```diff
@@ -0,0 +1,329 @@
...
+                    steerAngleHistory[i] += appliedOffsetDegrees;
+                }
+
+                // Recalculate statistics with the adjusted data
+                if (SampleCount >= MIN_SAMPLES_FOR_ANALYSIS)
+                {
+                    PerformStatisticalAnalysis();
+                }
+
+                AgLibrary.Logging.Log.EventWriter($"Smart WAS: Applied {appliedOffsetDegrees:F2}° offset to {steerAngleHistory.Count} collected samples");
```
FantasyTeddy: Maybe consider adding a `using AgLibrary.Logging;` at the top to simplify this line:
```suggestion
                Log.EventWriter($"Smart WAS: Applied {appliedOffsetDegrees:F2}° offset to {steerAngleHistory.Count} collected samples");
```

SourceCode/GPS/Classes/CSmartWASCalibration.cs
```diff
@@ -0,0 +1,329 @@
...
+        /// Called from the main GPS update loop
+        /// </summary>
+        /// <param name="guidanceSteerAngle">Current guidance line steer angle in degrees</param>
+        /// <param name="currentSpeed">Current vehicle speed in km/h</param>
+        public void AddSteerAngleSample(double guidanceSteerAngle, double currentSpeed)
+        {
+            if (!IsCollectingData) return;
+
+            // Check if we should collect this sample
+            if (!ShouldCollectSample(guidanceSteerAngle, currentSpeed)) return;
```
FantasyTeddy: Maybe the check for `IsCollectingData` should also be part of the `ShouldCollectSample` method?

SourceCode/GPS/Classes/CSmartWASCalibration.cs
```diff
@@ -0,0 +1,329 @@
...
+        // Analysis parameters
+        private const double NORMAL_DISTRIBUTION_THRESHOLD = 0.8;  // How much of data should be within 1 std dev
+
+        // Public properties for UI updates
+        public bool IsCollectingData { get; private set; }
+        public int SampleCount { get; private set; }
+        public double RecommendedWASZero { get; private set; }
+        public double ConfidenceLevel { get; private set; }
+        public bool HasValidRecommendation { get; private set; }
+        public DateTime LastCollectionTime { get; private set; }
```
FantasyTeddy: What is the `LastCollectionTime` being used for?

SourceCode/GPS/Classes/CSmartWASCalibration.cs
```diff
@@ -0,0 +1,329 @@
...
+        private const int MIN_SAMPLES_FOR_ANALYSIS = 200;  // Minimum samples for reliable analysis
+        private const double MIN_SPEED_THRESHOLD = 2.0;  // km/h - minimum speed for data collection
+        private const double MAX_ANGLE_THRESHOLD = 15.0;  // degrees - maximum angle to consider valid
+
+        // Analysis parameters
+        private const double NORMAL_DISTRIBUTION_THRESHOLD = 0.8;  // How much of data should be within 1 std dev
+
+        // Public properties for UI updates
+        public bool IsCollectingData { get; private set; }
+        public int SampleCount { get; private set; }
```
FantasyTeddy: Maybe `SampleCount` could be a computed property to always be up-to-date:
```suggestion
        public int SampleCount => steerAngleHistory.Count;
```

SourceCode/GPS/Classes/CSmartWASCalibration.cs
```diff
@@ -0,0 +1,329 @@
...
+                // Limit the number of samples to prevent memory issues
+                if (steerAngleHistory.Count > MAX_SAMPLES)
+                {
+                    steerAngleHistory.RemoveAt(0);  // Remove oldest sample
+                }
+
+                SampleCount = steerAngleHistory.Count;
+
+                // Perform analysis if we have enough samples
+                if (SampleCount >= MIN_SAMPLES_FOR_ANALYSIS)
```
FantasyTeddy: Since `PerformStatisticalAnalysis` already checks if we have enough samples, there is no need to do that here too.

SourceCode/GPS/Classes/CSmartWASCalibration.cs
```diff
@@ -0,0 +1,329 @@
...
+            if (SampleCount == 0) return "No data collected yet.";
+
+            var report = $"Smart WAS Calibration Analysis Report:\n" +
+                        $"Samples Collected: {SampleCount}\n" +
+                        $"Mean Angle: {Mean:F3}°\n" +
+                        $"Median Angle: {Median:F3}°\n" +
+                        $"Std Deviation: {StandardDeviation:F3}°\n" +
+                        $"Recommended WAS Zero: {RecommendedWASZero:F3}°\n" +
+                        $"Confidence Level: {ConfidenceLevel:F1}%\n" +
+                        $"Valid Recommendation: {HasValidRecommendation}";
```
FantasyTeddy: You mentioned in another PR, that we should try to keep everything translatable. Should we do that here too?

(Ideally we would not perform the translation here, but return just all the necessary information to do so. The translation should happen in the utilizing form.)

SourceCode/GPS/Classes/CSmartWASCalibration.cs
```diff
@@ -0,0 +1,329 @@
...
+        {
+            // Only collect data when moving at reasonable speed
+            if (speed < MIN_SPEED_THRESHOLD) return false;
+
+            // Ignore extreme angles that might be outliers
+            if (Math.Abs(steerAngle) > MAX_ANGLE_THRESHOLD) return false;
+
+            // Only collect when autosteer is active and we have valid guidance
+            if (!mf.isBtnAutoSteerOn) return false;
+            if (Math.Abs(mf.guidanceLineDistanceOff) > 15000) return false;  // > 15m off line
```
FantasyTeddy: Should we maybe also introduce a constant for the `15000` maximum distance?

gunicsba:
> I would prefer to create a function for it with a dedicated “activate” button instead of doing it directly while opening the form. Maybe add an automatic Counts per degree estimation function as well together with the Smart WAS.

Reason I didn't want an extra button / constantly running function is because based on my experience we only have to do WAS 0 once. And this is usually done while driving. In fact the best is to roughly set +/- 25-35 degrees on the steer limit with CPD / Ackermann then figure out the WAS 0 before you do the turn calibration. So for these reasons I believe this should be part of the steer form.

As for the separate function that might be useful if we want to run this from the steer wizard. Can we run it from there? Do we have an AB line for it to drive on?

SourceCode/GPS/Classes/CSmartWASCalibration.cs
```diff
@@ -0,0 +1,329 @@
...
+                // Log error but don't crash
+                AgLibrary.Logging.Log.EventWriter($"Error in Smart WAS analysis: {ex.Message}");
+                HasValidRecommendation = false;
+            }
+        }
+
+        /// <summary>
+        /// Calculate median value from sorted data
+        /// </summary>
+        private double CalculateMedian(List<double> sortedData)
```
FantasyTeddy: Maybe it would make sense to put the purely math related methods (`CalculateMedian` and `CalculateStandardDeviation`) into a separate helper class (e.g. `MathHelpers`)? That way they could also be tested quite easily.

SourceCode/GPS/Classes/CSmartWASCalibration.cs
```diff
@@ -0,0 +1,329 @@
...
+
+            double sumSquaredDifferences = data.Sum(x => Math.Pow(x - mean, 2));
+            return Math.Sqrt(sumSquaredDifferences / (data.Count - 1));
+        }
+
+        /// <summary>
+        /// Calculate confidence level based on how well the data fits a normal distribution
+        /// centered around the guidance line
+        /// </summary>
+        private void CalculateConfidenceLevel(List<double> sortedData)
```
FantasyTeddy: I think it would be cleaner if we passed `StandardDeviation` as an argument as well and maybe return the calculated confidence level instead of assigning it to `ConfidenceLevel` directly.
Otherwise it is a little bit weird that we pass the list of data, but take all the other information from somewhere else.

SourceCode/GPS/Forms/Position.designer.cs
```diff
@@ -999,6 +999,13 @@ public void UpdateFixPosition()
             //out serial to autosteer module  //indivdual classes load the distance and heading deltas 
             SendPgnToLoop(p_254.pgn);
 
+            // Smart WAS Calibration data collection
+            if (smartWASCalibration != null && Math.Abs(guidanceLineDistanceOff) < 500) // Within 50cm of guidance line
```
FantasyTeddy: Can the `smartWASCalibration` even be `null` at this point? (Same question goes for all the other `null` checks)

SourceCode/GPS/btnImages/Steer/ST_SmartWAS.png
FantasyTeddy, line 1: It took me a while to realize that this is supposed to be a magic wand. Maybe that should be a different image? (Or at least make the wand stand out more, such that it is recognizable)

FantasyTeddy:
> So for these reasons I believe this should be part of the steer form.

I think the problem is not that it is part of the steer form (that is perfectly fine in my opinion). The question is, if it has to run constantly when the steer form is open, or if it should be a button that needs to be activated in order to run the algorithm.

SourceCode/GPS/btnImages/Steer/ST_SmartWAS.png
gunicsba: Changed the icon I think it's more different this way.

SourceCode/GPS/Classes/CSmartWASCalibration.cs
gunicsba: I did it as 50cm (and it was 50cm in a different place) now they both use the same constant.

SourceCode/GPS/Classes/CSmartWASCalibration.cs
gunicsba: wasn't aware it's possible to do such, neatly looking solution it's in the code

SourceCode/GPS/Classes/CSmartWASCalibration.cs
gunicsba: It wasn't used at all, so I've removed that! Thanks for catching it.

SourceCode/GPS/Classes/CSmartWASCalibration.cs
gunicsba: Did the refactor as you asked.

SourceCode/GPS/Classes/CSmartWASCalibration.cs
gunicsba: Refactored all the logging lines.

SourceCode/GPS/Classes/CSmartWASCalibration.cs
gunicsba: Changed the class name

SourceCode/GPS/Classes/CSmartWASCalibration.cs
gunicsba: the main messages / popups are using gStr this is more like a debug information

SourceCode/GPS/Classes/CSmartWASCalibration.cs
FantasyTeddy: That's what code reviews are good for; learning new stuff :)

SourceCode/GPS/Classes/CSmartWASCalibration.cs
FantasyTeddy: Ah I see. But now you changed it to only log the values?

SourceCode/GPS/btnImages/Steer/ST_SmartWAS.png
FantasyTeddy: It is certainly better now, yes.

SourceCode/GPS/Classes/SmartWASCalibration.cs
```diff
@@ -0,0 +1,323 @@
...
+            if (!HasValidRecommendation) return 0;
+
+            // Convert degrees to counts
+            return (int)Math.Round(RecommendedWASZero * currentCPD);
+        }
+
+        /// <summary>
+        /// Get statistics and analysis results for UI display
+        /// </summary>
+        public CalibrationStats GetCalibrationStats()
```
FantasyTeddy, line 164: This is exactly what I had in mind for getting that information to the UI. But I think you are not using it?

SourceCode/GPS/Helpers/MathHelpers.cs
```diff
@@ -0,0 +1,48 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace AgOpenGPS.Helpers
+{
+    /// <summary>
+    /// Mathematical helper functions for statistical calculations
+    /// </summary>
+    public static class MathHelpers
```
FantasyTeddy, line 10: Do you think you could add some tests for these methods?

gunicsba: consider it done

SourceCode/GPS/Classes/SmartWASCalibration.cs
gunicsba: image
click the highlighted areas and it pops up

SourceCode/AgOpenGPS.Tests/Helpers/MathHelpersTests.cs
```diff
@@ -0,0 +1,280 @@
...
+
+            // Assert
+            Assert.That(result, Is.EqualTo(5.0));
+        }
+
+        [Test]
+        public void CalculateMedian_EvenNumberOfValues_ReturnsAverage()
+        {
+            // Arrange
+            var data = new List<double> { 1.0, 2.0, 3.0, 4.0 };
```
FantasyTeddy: This list is not optimal: it also has `2.5` as its average (mean).

SourceCode/AgOpenGPS.Tests/Helpers/MathHelpersTests.cs
```diff
@@ -0,0 +1,280 @@
...
+
+            // Act
+            double result = MathHelpers.CalculateMedian(data);
+
+            // Assert
+            Assert.That(result, Is.EqualTo(2.5));
+        }
+
+        [Test]
+        public void CalculateMedian_UnsortedData_ReturnCorrectMedian()
```
FantasyTeddy: I don't think this test is necessary. `CalculateMedian_OddNumberOfValues_ReturnsMiddleValue` already tests with an unsorted list.

SourceCode/AgOpenGPS.Tests/Helpers/MathHelpersTests.cs
```diff
@@ -0,0 +1,280 @@
...
+
+            // Act
+            double result = MathHelpers.CalculateStandardDeviation(data, mean);
+
+            // Assert
+            Assert.That(result, Is.EqualTo(0.0));
+        }
+
+        [Test]
+        public void CalculateStandardDeviation_NullList_ReturnsZero()
```
FantasyTeddy: I'm actually not sure if a `null` list should be even allowed as an argument. I think it would be safer if we threw an `ArgumentNullException` in that case.

SourceCode/AgOpenGPS.Tests/Helpers/MathHelpersTests.cs
```diff
@@ -0,0 +1,280 @@
...
+            // Expected std dev = sqrt((2450.25+2450.25)/1) = sqrt(4900.5) ≈ 70.0
+            Assert.That(result, Is.EqualTo(70.0).Within(0.1));
+        }
+
+        [Test]
+        public void CalculateStandardDeviation_RealWorldWASData_ReturnsCorrectValue()
+        {
+            // Arrange - Simulated WAS calibration data (steering angles in degrees)
+            var data = new List<double> { -2.1, -1.8, -2.0, -1.9, -2.2, -2.0, -1.7, -2.1, -1.9, -2.0 };
+            double mean = -1.97; // Approximately
```
FantasyTeddy: The fact that we have to know the mean value beforehand is a little bit unfortunate. What do you think if the `CalculateStandardDeviation` would return both the result (standard deviation) and the mean value (maybe as a tuple)?

SourceCode/GPS/Forms/Settings/FormSteer.cs
```diff
@@ -1352,7 +1352,27 @@ private void UpdateSmartCalibrationStatus()
 
         private void SmartCalLabel_Click(object sender, EventArgs e)
         {
-            // Log and show detailed analysis report
+            // Get calibration stats
+            var stats = mf.smartWASCalibration.GetCalibrationStats();
+
+            if (stats.SampleCount == 0)
+            {
+                mf.TimedMessageBox(2000, gStr.gsCalibrationDetails, "No data collected yet.");
```
FantasyTeddy: Maybe this should be translated too?

SourceCode/GPS/Forms/Settings/FormSteer.cs
```diff
@@ -1352,7 +1352,27 @@ private void UpdateSmartCalibrationStatus()
...
+            // Build translatable message
+            string message = $"{gStr.gsSamples}: {stats.SampleCount}\n" +
+                           $"{gStr.gsMean}: {stats.Mean:F3}°\n" +
+                           $"{gStr.gsMedian}: {stats.Median:F3}°\n" +
+                           $"{gStr.gsStandardDeviation}: {stats.StandardDeviation:F3}°\n" +
+                           $"{gStr.gsRecommendedOffset}: {stats.RecommendedWASZero:F3}°\n" +
+                           $"{gStr.gsConfidence}: {stats.ConfidenceLevel:F1}%\n\n" +
+                           (stats.HasValidRecommendation ? "✓" : "✗");
+
+            mf.TimedMessageBox(5000, gStr.gsCalibrationDetails, message);
```
FantasyTeddy: I am not a huge fan of using the `TimedMessageBox`. This is a good example why: you only give the user 5 seconds to look at all this data. Why? Shouldn't he be able to choose when he has read everything that he needs to?

This is not just an issue here, it is an issue through most of AgOpenGPS, where many `TimedMessageBox`es should probably be replaced by a dialog that can be manually closed.

SourceCode/GPS/Classes/SmartWASCalibration.cs
FantasyTeddy: Great!

SourceCode/GPS/Helpers/MathHelpers.cs
FantasyTeddy: Perfect, thank you!

SourceCode/AgOpenGPS.Tests/Helpers/MathHelpersTests.cs
gunicsba: I've changed it to unsorted list where median and average doesn't equal

SourceCode/AgOpenGPS.Tests/Helpers/MathHelpersTests.cs
gunicsba: Removed

SourceCode/GPS/Forms/Settings/FormSteer.cs
gunicsba: I think we should refactor TimedMessageBox itself and also put the messages in a log.
Then we should be able to show those logs with the Log Viewer. 

But that shouldn't be part of this. PR

FantasyTeddy: I agree. Maybe you can add your thoughts in this discussion I started: https://github.com/AgOpenGPS-Official/AgOpenGPS/discussions/1147

SourceCode/AgOpenGPS.Tests/Helpers/MathHelpersTests.cs
gunicsba: I overlooked this will do it tonight

SourceCode/AgOpenGPS.Tests/Helpers/MathHelpersTests.cs
FantasyTeddy: You are still checking for `null`. I guess you disagree with my suggestion?

gunicsba: sorry @FantasyTeddy  I thought I have removed that test... Seems not.

gunicsba:
Added the exception for Null arguments

FantasyTeddy:
This looks good to me now.

The question that remains is what @richardklasens brought up: shouldn't it be possible to "activate"/"deactivate" the functionality (even if the steering form is open)?

gunicsba:
In my opinion figuring out the WAS0 / CPD / Ackermann is a one time job. 

Maybe in the future it would be necessary to automatically zero the WAS for certain algorithms / IMU based WAS solutions?

FantasyTeddy:
> In my opinion figuring out the WAS0 / CPD / Ackermann is a one time job.

Which is why it is probably not a good idea to start it automatically everytime the steer form is open.
Have you considered making it some kind of wizard that could be run initially?

SourceCode/GPS/Forms/Settings/FormSteer.cs
```diff
@@ -1266,5 +1283,114 @@ private void btnVehicleReset_Click(object sender, EventArgs e)
...
+
+            if (Math.Abs(offsetCounts) > 50)
+            {
+                string msg = string.Format(gStr.gsExceededRangeMsg, offsetCounts);
+                mf.TimedMessageBox(3000, gStr.gsExceededRange, msg);
+                return;
+            }
+
+            // Apply the offset
+            hsbarWasOffset.Value += offsetCounts;
```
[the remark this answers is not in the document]
richardklasens: Was this solved @gunicsba

richardklasens:
I would just add a button to activate it in the extend form. Hit the button again to apply the auto calibrated value.

gunicsba:
We all agreed that having the WAS 0 being a true perfect zero is one of the most important steps in AgOpenGPS.
If we add an extra button then they don't know if the WAS 0 value needs adjustment.

Right now the the button is not active until you could apply the value. (And we also display the value we plan to apply before the button is pressed.)

If our current values are good it should display 0 as the required offset. If not then it'll display values.

gunicsba:
@richardklasens  not sure why but there was a compliation error ... I've fixed that now.

lansalot:
> We all agreed that having the WAS 0 being a true perfect zero is one of the most important steps in AgOpenGPS. If we add an extra button then they don't know if the WAS 0 value needs adjustment.
> 
> Right now the the button is not active until you could apply the value. (And we also display the value we plan to apply before the button is pressed.)
> 
> If our current values are good it should display 0 as the required offset. If not then it'll display values.

I think it has been mentioned before that sometimes you will need a different WAS offset in each configuration too?

What about perhaps it runs anyway, but if the user hasn't engaged the "auto tune" button, then it tells them once to perhaps consider engaging it? If not, it doesn't annoy them any further, and also it doesn't keep trying to optimise or send them changes to make?

gunicsba:
@lansalot I think first it would be better to have more people trying it and adding feedback. Maybe this could be auto applied every 30s or so?
I often have to tell people to get the WAS0 pressed until the tractor calms down

lansalot:
Will have a look and try it sometimes, maybe tmrw on Tuesday. Cheers!

richardklasens:
Added to the latest Betafixes branch to inject it in the upcoming release, i close the PR

PR #1146, opened 2026-01-03
Viewport must not strech content, but must display area that matches aspect ratio of viewport.

J005t67:
In contrast to earlier claims, the viewport class did not yet correctly handle viewports or bounding boxes with aspect ratio != 1.

When this PR is finished, I will make a PR to simplify the resize related code in FormGrid and FormBndTool.
 The resize related code will become more simple, because it no longer needs to keep the viewport square.
This will also enhance usability, because the user has more freedom to resize the panels.

### DIFF
--- a/SourceCode/AgOpenGPS.Core/DrawLib/GLW.DrawArrays.cs
+++ b/SourceCode/AgOpenGPS.Core/DrawLib/GLW.DrawArrays.cs
@@ -1,0 +1,37 @@
+﻿using OpenTK.Graphics.OpenGL;
+
+namespace AgOpenGPS.Core.DrawLib
+{
+    // GLW is short for GL Wrapper.
+    // Please use this class in stead of direct calls to functions in the GL toolkit.
+    public static partial class GLW
+    {
+
+        // Inlined by the compiler, so no function call overhead
+        public static void DrawLineLoopArrays(Vertex2Array array)
+        {
+            DrawArrays(PrimitiveType.LineLoop, array);
+        }
+
+        // Inlined by the compiler, so no function call overhead
+        public static void DrawLineStripArrays(Vertex2Array array)
+        {
+            DrawArrays(PrimitiveType.LineStrip, array);
+        }
+
+        private static void DrawArrays(PrimitiveType primitiveType, Vertex2Array vertex2Array)
+        {
+            if (vertex2Array == null) return;
+            vertex2Array.BindBuffer();
+
+            GL.VertexAttribPointer(0, 2, VertexAttribPointerType.Double, false, 0, 0);
+            GL.EnableVertexAttribArray(0);
+
+            GL.DrawArrays(primitiveType, 0, vertex2Array.Length);
+
+            GL.DisableVertexAttribArray(0);
+            GL.BindBuffer(BufferTarget.ArrayBuffer, 0);
+        }
+
+    }
+}
--- a/SourceCode/AgOpenGPS.Core/DrawLib/GLW.GeoCoordPrimitives.cs
+++ b/SourceCode/AgOpenGPS.Core/DrawLib/GLW.GeoCoordPrimitives.cs
@@ -73,7 +73,7 @@ namespace AgOpenGPS.Core.DrawLib
             if (vertices.Length >= MinVerticesForArray)
             {
                 Vertex2Array vertex2Array = new Vertex2Array(vertices);
-                GL.DrawArrays(primitiveType, 0, vertex2Array.Length);
+                DrawArrays(primitiveType, vertex2Array);
                 vertex2Array.Dispose();
             }
             else
@@ -101,11 +101,11 @@ namespace AgOpenGPS.Core.DrawLib
                 // background layer
                 SetLineWidth(backgroundStyle.Width);
                 SetColor(backgroundStyle.Color);
-                GL.DrawArrays(primitiveType, 0, vertex2Array.Length);
+                DrawArrays(primitiveType, vertex2Array);
                 // foreground layer
                 SetLineWidth(foregroundStyle.Width);
                 SetColor(foregroundStyle.Color);
-                GL.DrawArrays(primitiveType, 0, vertex2Array.Length);
+                DrawArrays(primitiveType, vertex2Array);
 
                 vertex2Array.Dispose();
             }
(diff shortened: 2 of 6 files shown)


SourceCode/AgOpenGPS.Core/Drawing/GeoViewportBase.cs
```diff
@@ -148,6 +144,30 @@ protected void OnMouseDown(GeoCoord mouseDownCoord)
...
+            {
+                double resizeNorthingFactor = scaleNorthingToY / scaleEastingToX;
+                projectedRect = BoundingBox.Resized(Zoom * resizeNorthingFactor, Zoom);
+            }
+            else
+            {
+                double resizeEastingFactor = scaleEastingToX / scaleNorthingToY;
+                projectedRect = BoundingBox.Resized(Zoom, Zoom * resizeEastingFactor);
+            }
+            GLW.CreateOrthoProjection((int)ViewportSize.DeltaX, (int)ViewportSize.DeltaY, projectedRect);
```
FantasyTeddy: Why not take `float` or `double` as width and height in `CreateOrthoProjection`?

SourceCode/AgOpenGPS.Core/Models/Base/GeoBoundingBox.cs
```diff
@@ -48,6 +48,18 @@ public bool IsInside(GeoCoord testCoord)
                 _minCoord.Easting <= testCoord.Easting && testCoord.Easting <= _maxCoord.Easting;
         }
 
+        public GeoBoundingBox Resized(double northingFactor, double eastingFactor)
```
FantasyTeddy: Maybe `Scaled` would be a more appropriate name for this method?

SourceCode/AgOpenGPS.Core/Drawing/GeoViewportBase.cs
```diff
@@ -69,14 +67,13 @@ public void BeginPaint()
             GLW.ClearColorAndDepthBuffer();
             GLW.LoadIdentity();
 
-            XyDelta viewportSize = ViewportSize;
-            GeoDelta boundingBoxSize = BoundingBox.MaxCoord - BoundingBox.MinCoord;
-            double scaleEastingToX = viewportSize.DeltaX / boundingBoxSize.EastingDelta;
-            double scaleNorthingToY = viewportSize.DeltaY / boundingBoxSize.NorthingDelta;
-            scaleBoundingBoxToViewport = Math.Min(scaleEastingToX, scaleNorthingToY);
+            UpdateProjection();
```
J005t67: For performance, UpdateProjection must be split in two parts:
- a part that only needs to be done when viewport or boundingbox changes
- a part that needs to be done for each frame.

SourceCode/AgOpenGPS.Core/Drawing/GeoViewportBase.cs
```diff
@@ -55,12 +52,13 @@ public void Resize(int width, int height)
 
         public abstract XyDelta ViewportSize { get; }
 
-        private double scaleBoundingBoxToViewport;
+        private double projectedRectToViewportFactor;
 
         public void SetBoundingBox(GeoBoundingBox boundingBox)
```
J005t67, line 72: The code will fit this boundingbox exactly to the viewport size. If the BoundingBox is for example the boundingbox of a field, a thick boundary line will be partially clipped on the edge of the viewport. We need to add a margin somewhere.

SourceCode/AgOpenGPS.Core/Drawing/GeoViewportBase.cs
J005t67: OpenGL requires a  viewport size is in integers (whole displayed pixels). I solved it the other way round. The variable ViewportSize is now in integers too.

SourceCode/AgOpenGPS.Core/Models/Base/XyDeltaInt.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace AgOpenGPS.Core.Models
+{
+    public struct XyDeltaInt
```
FantasyTeddy: I'm not a fan of this solution... I think the previous solution without this additional type was nicer... or maybe it is just the name of this `struct` that puts me off? Or maybe we could use one of the .NET types that represent a size here?

SourceCode/AgOpenGPS.Core/Drawing/GeoViewportBase.cs
J005t67: I did split UpdateProjection, but it turns out that nothing needs to be done for each frame.
Changes related to pan or zoom are directly taken care of during pan and zoom.

SourceCode/AgOpenGPS.Core/Drawing/GeoViewportBase.cs
```diff
@@ -53,30 +52,24 @@ public void Resize(int width, int height)
         public double Zoom { get; private set; }
         public GeoCoord ViewportCenter { get; set; }
 
-        public abstract XyDelta ViewportSize { get; }
+        public abstract XyDeltaInt ViewportSize { get; }
 
-        private double scaleBoundingBoxToViewport;
+        private GeoDelta DefaultDisplayedSize { get; set; }
+        private double DefaultBoundingBoxToViewportScale { get; set; }
```
FantasyTeddy: I think these should be fields instead of properties. (`private` -> field, `public` -> property)

SourceCode/AgOpenGPS.Core/Models/Base/XyDeltaInt.cs
J005t67: For future portability, I  wanted to avoid the Windows specific types like: System.Drawing.Size or System.Windows.Size, but I agree XyDeltaInt is not the way to go. I have removed XyDeltaInt and use a local struct now with a more appropriate name.

SourceCode/AgOpenGPS.Core/Drawing/GeoViewportBase.cs
```diff
@@ -15,31 +14,41 @@ namespace AgOpenGPS.Core.Drawing
 
     public abstract class GeoViewportBase
     {
-        private const double PanStep = 0.15;
-        private const double ZoomStep = 2.0;
+        public struct ViewportSize2D
```
FantasyTeddy: Maybe make this `private` or `protected`?

J005t67: Yes, protected is better! (Private is not an option, because subclasses need it too.)

SourceCode/AgOpenGPS.Core/Drawing/GeoViewportBase.cs
J005t67: I agree, changed.

PR #1148, opened 2026-01-04
Perferred order is 'public' before 'static'

J005t67:


FantasyTeddy (approved):
I also enabled the [IDE0036](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0036) analyzer rule to prevent any future ordering mistakes.

PR #1149, opened 2026-01-04
Log every popup and dialog to file

gunicsba:
Based on the discussion we had elsewhere I think we should also log all these popup messages to file.


richardklasens:
Closed since the formdialog has been fully refactored

PR #1152, opened 2026-01-05
Refactor/vertexbuffer

J005t67:
When I implemented the first version of VertexArrayBase and Vertex2Array, (probably more than a year ago), Vertex(2)Array(base) seemed a good name to me. From application side (calling side) it is indeed a array of vertices, but in OpenGL the same things is called VertexBuffer. Moreover, Vertex Array is a bit more high level comcept in OpenGL.

So that is why I renamed it to VertexBuffer.

### DIFF
--- a/SourceCode/AgOpenGPS.Core.Tests/Models/Base/GeoBoundingBoxTests.cs
+++ b/SourceCode/AgOpenGPS.Core.Tests/Models/Base/GeoBoundingBoxTests.cs
@@ -108,5 +108,22 @@ namespace AgOpenGPS.Core.Tests.Models
             Assert.That(bbA.IsEmpty, Is.True);
         }
 
+        [Test]
+        public void Test_Scaled()
+        {
+            // Arrange
+            GeoBoundingBox bb = GeoBoundingBox.CreateEmpty();
+            bb.Include(new GeoCoord(5.0, -3.0));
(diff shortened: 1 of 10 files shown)


SourceCode/AgOpenGPS.Core/DrawLib/GLW.GeoLineSegmentPrimitives.cs
```diff
@@ -25,9 +25,9 @@ private static void DrawPrimitive(PrimitiveType primitiveType, GeoLineSegment[]
             const int nVerticesPerSegment = 2;
             if (nVerticesPerSegment * lineSegments.Length >= MinVerticesForArray)
             {
-                Vertex2Array vertex2Array = new Vertex2Array(lineSegments);
-                DrawArrays(primitiveType, vertex2Array);
-                vertex2Array.Dispose();
+                VertexBuffer vertexBuffer = new VertexBuffer(lineSegments);
+                DrawArrays(primitiveType, vertexBuffer);
+                vertexBuffer.Dispose();
```
FantasyTeddy: Could we use `using` statements to handle the disposal of these objects?

J005t67: Yes! Changed at 6 places.

FantasyTeddy (approved):

PR #1153, opened 2026-01-05
Replace usages of `TimedMessageBox` where not appropriate

FantasyTeddy:
According to the discussion in #1147, I tried to replace all occurrences of `TimedMessageBox` with `FormDialog` where it is appropriate in my opinion.
I tried to follow this rule of thumb:
- If the message occurs in a user initiated workflow (e.g. in a specific form) then `FormDialog` is most likely the right choice.
- If the message occurs because of external factors (e.g. something steering related) then `TimedMessageBox` is still the best option.

Removing the calls to `TimedMessageBox` also has the additional benefit of removing another dependency to `FormGPS` from most forms.

### DIFF
--- a/SourceCode/GPS/Forms/Guidance/FormHeadAche.cs
+++ b/SourceCode/GPS/Forms/Guidance/FormHeadAche.cs
@@ -804,7 +804,7 @@ namespace AgOpenGPS
 
             if (crossings.Count != mf.hdl.tracksArr.Count * 2)
             {
-                FormDialog.Show("Crosings Error", "Make sure all ends cross and only once", MessageBoxButtons.OK);
+                FormDialog.Show("Crossings Error", "Make sure all ends cross and only once", MessageBoxButtons.OK);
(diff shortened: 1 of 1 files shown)


SourceCode/GPS/Forms/Guidance/FormHeadAche.cs
```diff
@@ -803,7 +804,7 @@ private void btnBndLoop_Click(object sender, EventArgs e)
 
             if (crossings.Count != mf.hdl.tracksArr.Count * 2)
             {
-                mf.TimedMessageBox(2000, "Crosings Error", "Make sure all ends cross and only once");
+                FormDialog.Show("Crosings Error", "Make sure all ends cross and only once", MessageBoxButtons.OK);
```
J005t67: Small typo in Crosings

FantasyTeddy: Fixed

Issue #1136, opened 2025-12-31
Performance: Set/send vertices only once


J005t67:
The current drawing code typically calls the function GL.Vertex2() or GLW.Vertex2() for all the verices again and again.
Mostly one time per vertex for each frame, but sometimes even two times per vertex per frame. For example for tramlines that have a dark outline, do call Vertex2 two times per vertex for each frame.

I expect that performance will improve when we store the vertices on the GPU in a buffer and only send them again if something changed. This will not be too difficult to implement, because I already have implemented Vertex2Array, that takes care of the intricities of buffers in OpenGL.

At the moment I do not know exactly what the new code will look like and what the performance gain will be. I like to try out a few things with relatively easy parts of the code first. Probably, either WorldGrid or TramLines.

PR #1154, opened 2026-01-06
Refactor: tramlinesvisual

J005t67:
I wanted to evaluate the performance improvements mentioned in:

#1136 Performance: Set/send vertices only once.

I choose to start with TramLines, because that code is relatively easy, (compared to for example the boundaries).

This is only a first step in refactoring TramLines. The plan is to follow up with the following PR's:
- Move all code that builds tram lines from CTram, CABLine and CABCurve to a new class TramLinesBuilder.
- Let the code in OpenGLDesigner ('//draw 245 green for the tram tracks') also use the new tramLinesVisual. 

Update 2026-01-07: Richard already tested the performance, and unfortunately the new code does not show good performance. I did not do measurements my self, so hopefully I phrase it correctly: the performance of the new TramLinesVisual is good!....... until you switch to guidance lines. Then the performance drops sharply, below the original  level without the new TramLinesVisual. 

Possible improvements I want to look at, (after doing some measurements myself):
- Currently GLW.DrawArrays creates a VAO (Vertex Array Object) from scratch for every frame. I want to make a small wrapper around VAO called VertexArray, (similar to class VertexBuffer that wraps VBO's) so we can store the VAO state and reuse it.
- Use floats instead of doubles.

### DIFF
--- a/SourceCode/GPS/Classes/CABCurve.cs
+++ b/SourceCode/GPS/Classes/CABCurve.cs
@@ -1310,8 +1310,8 @@ namespace AgOpenGPS
                     }
                 }
             }
-            List<GeoCoord[]> fillTrakcs = GeoRefactorHelper.ToGeoCoordArrays(mf.tram.tramList);
-            mf.tram.UpdateFillTracks(fillTrakcs);
+            List<GeoCoord[]> fillTracks = GeoRefactorHelper.ToGeoCoordArrays(mf.tram.tramList);
+            mf.tram.UpdateFillTracks(fillTracks);
         }
 
         //for calculating for display the averaged new line
--- a/SourceCode/GPS/Classes/CABLine.cs
+++ b/SourceCode/GPS/Classes/CABLine.cs
@@ -552,8 +552,8 @@ namespace AgOpenGPS
             }
 
             tramRef?.Clear();
-            List<GeoCoord[]> fillTrakcs = GeoRefactorHelper.ToGeoCoordArrays(mf.tram.tramList);
-            mf.tram.UpdateFillTracks(fillTrakcs);
+            List<GeoCoord[]> fillTracks = GeoRefactorHelper.ToGeoCoordArrays(mf.tram.tramList);
+            mf.tram.UpdateFillTracks(fillTracks);
         }
     }
 }


richardklasens:
It’s already a wip on my end as said in a message earlier. Have you tested the TramDetection?

SourceCode/AgOpenGPS.Core/Visuals/Field/TramLinesVisual.cs
richardklasens, line 1: The trams are not only a visual thing. It can really control a tram valve. The naming suggests it’s only a visual function

SourceCode/AgOpenGPS.Core/Visuals/Renderers/LineLoopRenderer.cs
richardklasens, line 1: Are we using this on more parts of the program? Or are you intending to do?

SourceCode/AgOpenGPS.Core/Visuals/Renderers/LineStripRenderer.cs
richardklasens, line 1: Same, are you using this on more parts?

SourceCode/GPS/Classes/CABCurve.cs
```diff
@@ -1317,6 +1310,8 @@ public void BuildTram()
                     }
                 }
             }
+            List<GeoCoord[]> fillTrakcs = GeoRefactorHelper.ToGeoCoordArrays(mf.tram.tramList);
```
richardklasens: Typo error on FillTracks

SourceCode/GPS/Classes/CABLine.cs
```diff
@@ -558,12 +552,8 @@ public void BuildTram()
             }
 
             tramRef?.Clear();
-            //outside tram
-
-            if (mf.bnd.bndList.Count == 0 || mf.tram.passes != 0)
-            {
-                //return;
-            }
+            List<GeoCoord[]> fillTrakcs = GeoRefactorHelper.ToGeoCoordArrays(mf.tram.tramList);
```
richardklasens: Same typo

SourceCode/GPS/Classes/CTram.cs
```diff
@@ -85,110 +89,49 @@ public void IsTramOuterOrInner()
...
-                    GL.End();
-                    GL.Begin(PrimitiveType.LineStrip);
-                    for (int h = 0; h < tramBndInnerArr.Count; h++)
-                    {
-                        GL.Vertex2(tramBndInnerArr[h].easting, tramBndInnerArr[h].northing);
-                    }
-                    GL.End();
-                }
-            }
+            LineStyle backgroundLineStyle = new LineStyle(
```
richardklasens, line 92: Why creating a new line style here? It’s always the same style.

SourceCode/AgOpenGPS.Core/Visuals/Field/TramLinesVisual.cs
J005t67: Currently, CTram is a bit of everything. The data itself, guidance related code, guidance visualization related code, code to build tramlines, (also partly in CABCurve and CABLine.) I only moved out the drawing related code to a well-named class. This PR will not touch the control of the valves.

SourceCode/AgOpenGPS.Core/Visuals/Renderers/LineLoopRenderer.cs
J005t67: Yes, if this PR is completed succesfully, I will introduce more visual classes that will reuse the Renderer classes. The idea is that the Visual classes are specific for something specific like fence lines, guidance lines, or whatever and will take care of the more high level things. I will keep the Renderer classes more simple and more reusable.

SourceCode/GPS/Classes/CTram.cs
J005t67: It is not exactly always the same, (alpha and camera distance both influance the line style), but I agree that it does not look good. I will move the color variable to the visual and only update it when alpha changes, so not every frame.
Also I will not longer combine the linewidth and color to a LineStyle for every frame, but keep them separated.

SourceCode/GPS/Classes/CABCurve.cs
J005t67: Fixed

SourceCode/GPS/Classes/CABLine.cs
J005t67: Fixed

richardklasens:
Based on our private conversation, the performance drops pretty hard when trams are active and you turn on a Guidance Line.

SourceCode/GPS/Classes/CTram.cs
```diff
@@ -12,7 +12,7 @@ namespace AgOpenGPS
     public class CTram
     {
         private readonly FormGPS mf;
-        private readonly TramLinesVisual _visual;
+        private readonly TramLinesVisual _tramLinesVisual;
```
J005t67, line 15: Bug: When a field is opened, the tramLinesVisual is not updated.

richardklasens:
The performance issue needs to be solved first (in a separate PR). That would be best.

richardklasens:
> * Move all code that builds tram lines from CTram, CABLine and CABCurve to a new class TramLinesBuilder.

First we need to create a unified guidance system for every line type.. that’s the basic part of the whole program. After that do headlands, after that, do trams. Until that these PR’s have to be open or closed until that part is done. 

I would also prefer a refactor plan in a structured MD file to see what the plans are. The plan should be added in a newly created documents folder where a refactor plans folder can be added. Then we can discuss the plans before we refactor or add stuff.

PR #1159, opened 2026-01-11
Add severity to `FormDialog`

FantasyTeddy:
This adds a severity to the `FormDialog`:

## Error
Screenshot 2026-01-11 190343

## Warning
Screenshot 2026-01-12 060802

## Info
Screenshot 2026-01-11 190450

---

If no severity is set, the dialog looks like this:

Screenshot 2026-01-11 185806


J005t67:
On my screen, the yellow border of the warning dialog looks a bit greenish. I would prefer a color that matches the color of the warning icon.

FantasyTeddy:
> On my screen, the yellow border of the warning dialog looks a bit greenish. I would prefer a color that matches the color of the warning icon.

I agree. I updated the image in the description.

SourceCode/GPS/Forms/FormDialog.cs
```diff
@@ -1,14 +1,37 @@
...
+        private static readonly Color ErrorBorderColor = Color.FromArgb(192, 0, 0);
+        private static readonly Color ErrorBackgroundColor = Color.FromArgb(255, 192, 192);
+        private static readonly Color WarningBorderColor = Color.FromArgb(192, 145, 0);
+        private static readonly Color WarningBackgroundColor = Color.FromArgb(227, 217, 152);
+        private static readonly Color InfoBorderColor = Color.FromArgb(0, 0, 192);
+        private static readonly Color InfoBackgroundColor = Color.FromArgb(192, 192, 255);
+
+        private Color _borderColor = Color.CornflowerBlue;
+
+        private FormDialog(string title, string message, bool showCancel, DialogSeverity? severity)
```
J005t67, line 26: To me, severity is an essential part of the dialogs. I prefer to make it not optional.

FantasyTeddy: For simple confirm-only dialogs (where it is only possible to press "OK") I totally agree and I could change that.
For the "Yes"/"No" type of dialog I found that a severity does not always (or almost never) makes sense. In fact, I was even considering adding a question mark icon for those cases.

What do you think? Do you disagree?

J005t67: Sorry, for my late respons. I was not aware of that 'problem' and I think using a question mark without severity is a good choice.

J005t67 (approved):
I think this PR already improves the usability as is and we can complete it as is.
For me it would be OK, to do any additional improvements (if any) in another pull request.

PR #1163, opened 2026-01-16
ISOBUS PGN out change

Pat-I:
Update PGN F2 to align more to ISOBUS standard, to eventually be able to send a command and element number.
If a change should be made, it should be now, before the AGOtaskcontroller is officially released. It will need a change there too.

First draft, someone will have to check the bit shifts.

This is the complete ISOBUS PGN 0xCB00:
image
image
image

Well, the default values of "command" and "elementNumber" sent by AOG should probably also be discussed. both are "0" now.


Pat-I:
Since this is only for the ISOBUS PGN 0xCB00 I will close it and try to make a new one were you could send diferent PGNs too.

PR #1164, opened 2026-01-17
sending Isobus type PGN

Pat-I:
This change the AOG PGN 0xF2 to also send the CANBUS PGN number, his priority, source  and destination address.

It now also send all 8 data byte of the ISOBUS Process Data Message (DDEntity)

The AOG message is now 7 bytes longer.

Right now it only send the "ISOBUS Process Data Message (DDEntity)", PGN 0xCB00(51968):
image
image

The goal of this change is to be able to exchange more isobus data in the future without having to create other PGNs (at the cost of more data sent today).

### DIFF
--- a/SourceCode/GPS/Forms/OpenGL.Designer.cs
+++ b/SourceCode/GPS/Forms/OpenGL.Designer.cs
@@ -1331,6 +1331,7 @@ namespace AgOpenGPS
                     number |= 1ul << j;
                 }
             }
+            if (!isJobStarted) number = 0; //
 
             //there has been a status change of section on/off
             if (number != lastNumber)
--- a/SourceCode/GPS/Classes/CISOBUS.cs
+++ b/SourceCode/GPS/Classes/CISOBUS.cs
@@ -200,7 +200,7 @@ namespace AgOpenGPS
             SectionControlEnabled = ReadBit(data[0], 0);
             int numberOfSections = data[1];
 
-            if (data.Length != 2 + (numberOfSections + 7) / 8)
+            if (data.Length < 2 + (numberOfSections + 7) / 8)
             {
                 // Make sure we have enough data to read all the section states
                 return false;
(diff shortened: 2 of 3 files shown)


Pat-I:
I made some changes
It's an 1:1 J1939 CANBUS message. These two integer would make it ready to send with the teensy library
id: the first 4 bytes make the identifier
data: the 8 next bytes are the payload

Edited to remove the inaccurate test results.

Pat-I:
Description of the proposed  modified AOG PGN F2:
0x80,0x81,SRC(AOG),0xF2,13(length),0(future use), ID3, ID2, ID1, ID0, DATA 0 TO 7, crc

Example: ISOBUS PGN: 0xCB00, priority: 3, DA(PS): F7, SA: 82
Data: Command: 3, element: 5, DDE: 75, data: 3847700

On CANBUS: 0xCCBF782, 53, 0, 4B, 0, 14, B6, 3A, 0
In AOG:
image

Overall the cost is 7 more bytes (+58%) and the benefit is to be able send (almost?) any CAN J1939 message between controllers in AOG

Pat-I:
The last commit allow "private void SendCanbusMessage(int id, byte[] data)" to send a variable number of "data" bytes. There's no limit of bytes for this PGN although CANBUS is 1 to 8 bytes.

SourceCode/GPS/Forms/OpenGL.Designer.cs
```diff
@@ -1331,6 +1331,7 @@ private void oglBack_Paint(object sender, PaintEventArgs e)
                     number |= 1ul << j;
                 }
             }
+            if (!isJobStarted) number = 0; //
```
richardklasens, line 1334: Can the variable number be changed to something appropriate?

Pat-I: I did not choose this name. I only set it to 0 in 2 more situations

PR #1168, opened 2026-02-24
Improves curve and boundary calculations, More fixes

richardklasens:
Addresses several issues related to curve and boundary calculations, focusing on performance and display enhancements:

- Caps curve side guidelines at a maximum of 3 to improve performance, especially on low end devices.
- Simplifies curve extensions by using a single point 200m out instead of multiple points.
- Hides curve display during recalculation to prevent flickering when switching tracks.
- Uses local coordinates to calculate curve point headings.
- Moves `oglMain.Refresh` call within the frame time calculation for more accurate timing.
- Reduces reference track size and fixes issues related to guideline extensions
- Fixes dialog resizing issues


richardklasens:
@codex please reveiw again

PR #1179, opened 2026-04-05
ISOBUS Heartbeat / usage improvement

gunicsba:
I've made some changes to TC and it'll send 0 sections in case we don't have an implement at sight. This should help us figure out if TC is running AND talking with AgOpenGPS.

If we fail to set the UDP OR not on the right bus OR no implement connected then we never had the ISOBUS SC button. Which is a pain in the ass.

Now if TC is running we have ISOBUS ready button.


gunicsba:
This button will appear:
image
When:
Task controller is running
Task controller is connected to the right UDP network and communicates
No implement is connected

When implement is connected it'll switch back to SC on/off

PR #1209, opened 2026-06-12
Improves boundary recording and simulator lockout

richardklasens:
Addresses artifacts in boundary recording and enhances simulator stability.

**Boundary Recording Improvements:**
- Prevents boundary overlap artifacts when an operator drives past the start point.
- Introduces `TrimDrivenOvershoot` to intelligently remove excess path by:
    - Stripping trailing points that loop straight back to the start.
    - Resolving angled overshoots by finding intersections and retaining the enclosed loop.
- Enhances `RemoveSelfIntersections` to prevent discarding valid boundaries if the cleanup process results in a disproportionately small or degenerate polygon.

**Simulator Safety & Lockout:**
- Fixes simulator coordinates from potentially leaking into live GNSS fixes, which could cause erratic position jumps.
- Implements a "sticky lock" (`isSimDisabledLocked`) that permanently disables the simulator for the remainder of the session once it's turned off (either manually or by the system).
- Ensures the simulator cannot be accidentally re-enabled and that its timer and fix quality spoofing mechanisms are completely halted when locked off, preventing interference with live data.

### DIFF
--- a/SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
+++ b/SourceCode/GPS/Classes/AgShare/Helpers/FieldParser.cs
@@ -45,7 +45,11 @@ namespace AgOpenGPS.Classes.AgShare.Helpers
                 Origin = origin
             };
 
-            var converter = new GeoConverter(origin.Latitude, origin.Longitude);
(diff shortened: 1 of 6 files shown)


richardklasens:
@codex please review

Issue #1217, opened 2026-09-02
Expose current guidance track/pass information over UDP for external integrations


gunicsba:
### Problem or Use Case

Feature request

It would be useful if AgOpenGPS exposed the current guidance-track context over its UDP protocol.

At the moment external applications can receive guidance-related information such as XTE, but they do not have enough information to identify the actual guidance pass / track that AOG is currently using.

This is needed for our ISOBUS TRACK / tramline-control implementation, but the information would likely also be useful for other external controllers and integrations.

Current limitation

From the external application's point of view, we currently do not have access to:

A stable ID for the currently active guidance reference / guidance pattern

Current guidance track / pass number

Track number immediately to the left

Track number immediately to the right

Today we have to approximate the current pass from the existing AOG tramline state.

That works as a temporary fallback, but it is inherently unreliable because the tramline state is not the same thing as the guidance-system track number.

Requested data

Ideally AOG would expose a guidance-track context containing at least:

guidanceReferenceId
currentTrackNumber
trackNumberLeft
trackNumberRight

guidanceReferenceId

A stable identifier for the active guidance reference / pattern.

For example, it should remain the same while travelling through:

... -3, -2, -1, 0, 1, 2, 3 ...

and change when the operator changes to another AB line / curve / guidance pattern.

It does not necessarily need to match any ISOBUS-specific representation. We only need a stable value that lets an external application distinguish one guidance reference from another.

currentTrackNumber

The signed pass / swath number relative to the active guidance reference.

It should support:

... -3, -2, -1, 0, +1, +2, +3 ...

trackNumberLeft / trackNumberRight

The neighboring guidance tracks relative to the current vehicle / implement orientation.

Providing these explicitly is preferable to requiring an external application to infer orientation from heading or vehicle movement.

Protocol options

There are two possible ways to expose this.

Option A — extend an existing guidance/XTE message

If there is sufficient room and this fits the existing protocol design, the current guidance/XTE message could also include the track context.

Option B — add a dedicated Guidance Track Context PGN

A separate message may be cleaner, for example:

Guidance Track Context

uint32 guidanceReferenceId
int32  currentTrackNumber
int32  trackNumberLeft
int32  trackNumberRight

An optional sequence counter / validity flags could also be useful so receivers can treat all four values as one coherent snapshot.

The exact field sizes and message layout are open for discussion.

Why left/right should be provided by AOG

An external application can calculate:

currentTrack - 1
currentTrack + 1

[...]

PR #1218, opened 2026-09-02
Add a new PGN for isobus Tracks.

gunicsba:
Aiming to solve https://github.com/AgOpenGPS-Official/AgOpenGPS/issues/1217 

Currently in draft until fully tested

### DIFF
--- a/SourceCode/GPS/Classes/CISOBUS.cs
+++ b/SourceCode/GPS/Classes/CISOBUS.cs
@@ -23,6 +23,15 @@ namespace AgOpenGPS
         private int lastTotalDistance;
         private DateTimeOffset totalDistanceTime;
 
+        // Guidance Track Context (PGN 0xF4)
+        private byte trackContextSequence;
+        private int lastTrackContextRefId;
+        private int lastTrackContextCurrent;
+        private int lastTrackContextLeft;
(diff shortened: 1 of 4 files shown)


gunicsba:
Tested it with KUHN ESPRO 6002 model year 2025.
The way ISOBUS Level 1 tramline works requires no tramlines to be set in agopengps. All we need is a properly placed AB reference line.

TC implementation:
https://github.com/AgOpenGPS-Official/AOG-TaskController/pull/73

gunicsba:
PGN 0xF4 now carries the guidance track spacing (commit 19f770fe), for DDI 512 GuidanceLineSwathWidth (AOG-TaskController#88).

- New bytes **15-16**: Swath Width, `uint16` little-endian, **millimetres**. Value is `0` when there is no active track (Valid flag clear).
- Payload length is now 12 (`message[4] = 12`), total message 18 bytes, CRC at byte 17 (sum of bytes 2-16).
- The value is tool width minus overlap, the same spacing used to compute the track numbers.
- A width or overlap change triggers a send (100 ms minimum interval unchanged).

Decoding: `ushort swathWidthMm = BitConverter.ToUInt16(data, 15);` See `docs/pgn-protocol.md`.

PR #1219, opened 2026-09-02
Fix NTRIP client silently accepting rejected connections

aortner:
- Re-resolve the caster hostname on every reconnect instead of only at startup, so a changed IP (e.g. dynamic DNS) is picked up
- Validate the caster's HTTP response before marking the connection as established, instead of assuming success right after sending the request
- Detect casters that reply "200 OK" but close the connection again before any real RTCM data (RTCM3 sync byte 0xD3) ever flows, and treat that as a rejection too
- Surface connection failures with a visible message instead of failing silently: bad mountpoint/credentials, an unresolvable caster URL, a caster closing without any response, and socket connect errors
- Fix async socket callbacks (OnConnect/OnRecievedData) operating on a stale Socket instance after a reconnect replaced clientSocket
- Avoid duplicate rejection popups when the caster's rejection text and connection close arrive almost simultaneously, by evaluating connection state on the UI thread instead of the IO callback thread
- Flush NTRIP-related log entries immediately instead of only on exit

Claude-Session: https://claude.ai/code/session_01F3qKLb85UESKFHSuLkX1uC


richardklasens:
@codex please review

gunicsba:
This would also help when the internet starts after AgIO started with an old profile.

SourceCode/AgIO/Source/Forms/NTRIPComm.Designer.cs
```diff
@@ -46,6 +46,13 @@ public partial class FormLoop
         public bool isNTRIP_Sending = false;
         public bool isRunGGAInterval = false;
 
+        //true once the HTTP request has been sent to the caster, reset on every reconnect
+        private bool isNTRIP_AuthSent = false;
+
+        //counts real RTCM bytes received since the caster accepted the connection - a caster
+        //can send "200 OK" and still reject the mountpoint by closing before any real data flows
+        private uint bytesSinceConnected = 0;
```
richardklasens, line 54: Small addition needed to support the throttle suggested below on `ResolveCasterIP()`.

```suggestion
        private uint bytesSinceConnected = 0;

        //throttle for ResolveCasterIP() - avoid a blocking DNS lookup on every reconnect retry
        private DateTime lastCasterResolveAttempt = DateTime.MinValue;
```

SourceCode/AgIO/Source/Forms/NTRIPComm.Designer.cs
```diff
@@ -199,10 +208,60 @@ public void ConfigureNTRIP()
...
+        // Resolve the caster hostname/URL to an IPv4 address. Updates broadCasterIP (and
+        // persists the setting) only when a new address is found; on failure the previous
+        // broadCasterIP is left untouched so a temporary DNS hiccup doesn't stop reconnects.
+        private bool ResolveCasterIP()
+        {
+            string actualIP = Properties.Settings.Default.setNTRIP_casterURL.Trim();
+
+            try
+            {
+                IPAddress[] addresslist = Dns.GetHostAddresses(actualIP);
```
richardklasens, line 220: **Blocking DNS lookup now runs on the UI thread on every reconnect attempt.**

`ResolveCasterIP()` is called from `StartNTRIP()`, which is invoked from `DoNTRIPSecondRoutine()` off `oneSecondLoopTimer.Tick` — i.e. the UI thread. Previously the DNS lookup only happened once, at form load. With this PR it now happens on *every* reconnect attempt (roughly every 5-9s while not connected).

`Dns.GetHostAddresses` is a synchronous, blocking call. If the network/DNS is actually down — exactly the scenario this PR is trying to surface better to the user — this call can block for several seconds before timing out, and it will do so repeatedly, freezing the AgIO UI every retry cycle for the whole duration of the outage. That seems to work against the PR's own goal.

A simple mitigation: throttle re-resolution so we don't hit DNS on every single retry once we already have a working IP (needs the `lastCasterResolveAttempt` field added in the other comment on this file):

```suggestion
        private bool ResolveCasterIP()
        {
            // Avoid a synchronous DNS lookup (runs on the UI thread and can block for
            // several seconds when the network/DNS is actually down) on every single
            // reconnect retry - only really re-resolve every 30s once we have a known-good IP.
            if (!string.IsNullOrEmpty(broadCasterIP) && (DateTime.Now - lastCasterResolveAttempt).TotalSeconds < 30)
            {
                return true;
            }
            lastCasterResolveAttempt = DateTime.Now;

            string actualIP = Properties.Settings.Default.setNTRIP_casterURL.Trim();

            try
            {
                IPAddress[] addresslist = Dns.GetHostAddresses(actualIP);
```

A more thorough fix would move the lookup off the UI thread entirely (`Dns.GetHostAddressesAsync` / `Task.Run` with the result marshalled back via `BeginInvoke`), but the throttle above is a minimal change that bounds the damage.

SourceCode/AgIO/Source/Forms/NTRIPComm.Designer.cs
```diff
@@ -421,9 +493,37 @@ private void SendAuthorization()
...
+                else
+                {
+                    string firstLine = response.Split('\r')[0];
+                    Log.EventWriter("NTRIP - Caster rejected connection: " + firstLine);
+                    TimedMessageBox(2500, "NTRIP Connection Rejected", firstLine);
+                    ReconnectRequest();
+                }
+                Log.FileSaveSystemEvents(); //flush immediately for live diagnostics
+                return;
+            }
```
richardklasens, line 518: **Possible loss of the first RTCM chunk if it arrives in the same read as the "200 OK".**

This block treats the *entire* first packet received after `isNTRIP_AuthSent` as the HTTP response and unconditionally `return`s once it's handled. If a caster sends the HTTP headers and the first RTCM3 frame back-to-back and they land in the same `BeginReceive` completion (quite possible - `NoDelay = true` reduces coalescing but doesn't prevent it, and casters aren't guaranteed to flush the header separately), that RTCM data is silently dropped: it's never handed to `SendNTRIP`/the RTCM parsing below.

Suggest splitting off any bytes after the header terminator and letting them fall through to the normal data path instead of always returning:

```suggestion
            if (isNTRIP_AuthSent && !isNTRIP_Connected && !Properties.Settings.Default.setNTRIP_isTCP)
            {
                string response = Encoding.ASCII.GetString(data);

                if (response.Contains("200"))
                {
                    isNTRIP_Connected = true;
                    isNTRIP_Connecting = false;
                    bytesSinceConnected = 0;
                    Log.EventWriter("NTRIP - Caster accepted connection");
                    Log.FileSaveSystemEvents(); //flush immediately for live diagnostics

                    //some casters push the first RTCM bytes in the same read as the HTTP
                    //response - don't drop them, hand them to the normal data path below
                    int headerEnd = response.IndexOf("\r\n\r\n", StringComparison.Ordinal);
                    if (headerEnd < 0 || headerEnd + 4 >= data.Length) return;

                    byte[] remaining = new byte[data.Length - (headerEnd + 4)];
                    Array.Copy(data, headerEnd + 4, remaining, 0, remaining.Length);
                    data = remaining;
                }
                else
                {
                    string firstLine = response.Split('\r')[0];
                    Log.EventWriter("NTRIP - Caster rejected connection: " + firstLine);
                    TimedMessageBox(2500, "NTRIP Connection Rejected", firstLine);
                    ReconnectRequest();
                    Log.FileSaveSystemEvents(); //flush immediately for live diagnostics
                    return;
                }
            }
```

SourceCode/AgIO/Source/Forms/NTRIPComm.Designer.cs
```diff
@@ -582,15 +682,36 @@ private void NTRIPtick(object o, EventArgs e)
...
+                    try { sock.Close(); } catch { /* already closed/disposed */ }
+
+                    string message = ex.Message;
+                    BeginInvoke((MethodInvoker)(() =>
+                    {
+                        TimedMessageBox(2500, "NTRIP Connection Failed", message);
+                        ReconnectRequest();
+                    }));
+                }
             }
```
richardklasens, line 715: **Logging call happens directly on the IO-completion thread, not marshalled to the UI thread.**

`Log.sbEvents` (in `AgLibrary.Logging.Log`) is a plain static `StringBuilder` with no locking, and `FileSaveSystemEvents()` iterates/clears it. Everywhere else in the codebase `Log.EventWriter`/`Log.FileSaveSystemEvents()` is called from the UI thread. Here, though, they're called directly in the `catch` block of an async socket callback - i.e. a thread-pool thread - *before* the `BeginInvoke` marshal that the rest of this method already does. If the UI thread happens to call `Log.EventWriter` at the same moment (e.g. from `DoNTRIPSecondRoutine`, which does exactly that once a second), you get a data race on the shared `StringBuilder`. Worst case that throws, and since it's already inside this `catch` block the exception would go unhandled on a thread-pool thread.

Suggest moving the logging into the existing `BeginInvoke` block:

```suggestion
            catch (Exception ex)
            {
                //only react if this is still the current connection attempt
                if (sock == clientSocket)
                {
                    try { sock.Close(); } catch { /* already closed/disposed */ }

                    string logMessage = ex.ToString();
                    string userMessage = ex.Message;
                    //marshal logging to the UI thread too - Log.sbEvents is a plain
                    //StringBuilder and isn't safe to touch concurrently from an IO thread
                    BeginInvoke((MethodInvoker)(() =>
                    {
                        Log.EventWriter("Catch -> NTRIP OnConnect: " + logMessage);
                        Log.FileSaveSystemEvents(); //flush immediately for live diagnostics
                        TimedMessageBox(2500, "NTRIP Connection Failed", userMessage);
                        ReconnectRequest();
                    }));
                }
            }
```

Same pattern applies to the `OnRecievedData` catch block below (separate comment).

SourceCode/AgIO/Source/Forms/NTRIPComm.Designer.cs
```diff
@@ -599,26 +720,62 @@ public void OnRecievedData(IAsyncResult ar)
...
+                    }));
                 }
             }
-            catch (Exception)
+            catch (Exception ex)
             {
-                //MessageBox.Show( this, ex.Message, "Unusual error druing Recieve!" );
+                Log.EventWriter("Catch -> NTRIP OnRecievedData: " + ex.ToString());
+                Log.FileSaveSystemEvents(); //flush immediately for live diagnostics
             }
```
richardklasens, line 779: Same issue as the `OnConnect` catch block above: `Log.EventWriter`/`Log.FileSaveSystemEvents()` run directly on the IO-completion thread here instead of being marshalled to the UI thread, racing with the UI thread's own use of the static, non-thread-safe `Log.sbEvents`.

```suggestion
            catch (Exception ex)
            {
                //marshal to the UI thread - Log.sbEvents is a plain StringBuilder and isn't
                //safe to touch concurrently with the UI-thread logging calls elsewhere
                string message = ex.ToString();
                BeginInvoke((MethodInvoker)(() =>
                {
                    Log.EventWriter("Catch -> NTRIP OnRecievedData: " + message);
                    Log.FileSaveSystemEvents(); //flush immediately for live diagnostics
                }));
            }
```

richardklasens (changes requested):
Thanks for this — the core fix (no longer marking the connection "up" the instant the GET is sent, regardless of what the caster actually replies) and the stale-`clientSocket`-in-async-callback fix are both real bugs worth fixing.

Left a few inline comments with suggested changes, roughly in order of importance:

1. **DNS resolution now happens synchronously on the UI thread on every reconnect attempt** (was previously only at form load). If the network/DNS is down - the exact scenario this PR wants to report better - this can repeatedly freeze the AgIO UI for the whole outage. Suggested a simple throttle.
2. **Possible silent loss of the first RTCM chunk** if a caster's "200 OK" and the first RTCM frame land in the same socket read - the current code always returns after handling the header, without forwarding any trailing bytes.
3. **Two `catch` blocks log directly on the socket IO-completion thread** instead of the UI thread, which races with the (non-thread-safe, static) `Log.sbEvents` `StringBuilder` used everywhere else from the UI thread.

None of these are directly caused by the main fix's logic (which looks correct), but #1 in particular seems worth addressing before merge since it can make the app feel broken in exactly the situation this PR is meant to help with.

aortner:
if it is ok please test

richardklasens (approved):

gunicsba:
@richardklasens @aortner this doesn't work with centipede NEAR...

Was it tested?

richardklasens:
> @richardklasens @aortner this doesn't work with centipede NEAR...
> 
> 
> 
> Was it tested?
> 
> 

I guess Andreas did?

aortner:
i did test it with different casters.

in general i gives more feedback if connection was not sucessfull.

aortner:
ok i see there is a problem with near. i will check it

gunicsba:
@aortner I dug into the NEAR failure — it's deterministic, and it explains why it worked for you against ordinary casters. Fix is in #1225.

Short version: `OnAddMessage()` returns early when the caster's `200 OK` arrives in a read of its own with no RTCM behind it, and that early return jumps over the only `NTRIP_Watchdog = 0;` in the file. The watchdog starts at `100` and is cleared nowhere else, so the very next `IncrementNTRIPWatchDog()` tick hits `NTRIP_Watchdog++ > 30 && isNTRIP_Connected` and calls `ReconnectRequest()` — one second after "Caster accepted connection" is logged. That also disposes `tmr`, so no GGA is ever sent.

Normal mount points get away with it because their first RTCM frame lands inside the same second and clears the watchdog in time. NEAR (and VRS mount points generally) send nothing at all until they have a position, so they never clear it and get torn down every time.

#1225 resets the watchdog on accept, sends a GGA straight away rather than waiting for the 5 second timer, plus two smaller things the same code path turned up: `SOURCETABLE 200 OK` (an unknown mount point) currently satisfies `Contains("200")` and is reported as a successful connection, and `data[0] == 0xD3` misses reads that start mid-frame, which can produce a false "check mountpoint/credentials" popup on a healthy stream.

I've tested it against a normal caster — connects and stays connected, no reconnect cycling. I don't have a NEAR setup to hand, so if you or @richardklasens can point a build at `caster.centipede.fr` / `NEAR` that would close it out properly.

PR #1221, opened 2026-09-17
Avoid blocking TC stdout/stderr readers on the AgIO UI thread

MarekJarczewwski:
## Summary

Keep Task Controller stdout/stderr readers independent of the WinForms UI loop.
Previously each received line synchronously invoked the UI thread and modified
the textbox. A busy UI could stop draining the redirected output pipe and block
TC on logging. Standard error was redirected but BeginErrorReadLine was missing.

This change:

- Drains both redirected streams asynchronously.
- Enqueues output in a thread-safe, bounded 100,000-character buffer.
- Flushes bounded batches from a 200-ms UI timer while the form is visible.
- Trims the display in chunks, limits oversized lines and reports dropped display
  lines; TC file logging remains enabled.
- Uses guarded BeginInvoke for process-exit UI updates and disposes the timer
  through the form's component container.

TC launch/stop policy, adapter selection, application settings and the installation
path lookup are unchanged. The local deployment-path override is not included.

## Validation

- Built current AgIO develop with this patch in Release: zero warnings/errors.
- Added tests/Test-TcLogQueue.ps1 and a Windows CI invocation on the published
  AgIO executable.
- PASS: 20,000 concurrent lines while the UI message loop is deliberately not
  pumped; bounded queue, oversized line, draining, recent-line retention and
  disposal. The test constructs an invisible form and does not launch TC or CAN.

The same queue logic was deployed during a nine-section Kverneland sprayer test.
After this and separate TC actual-feedback/subscription fixes, the operator
reported further tests on several fields without observed problems. This is
combined operator-reported experience, not proof that log backpressure caused
every previous timeout or that this patch alone fixed section mapping.

## Scope

No binaries, deployment scripts, raw field logs, credentials or machine-specific
data are included. This draft requests maintainer review before merge; related
TC fixes belong in AgOpenGPS-Official/AOG-TaskController.

Related actual-feedback and measurement-subscription draft:
https://github.com/AgOpenGPS-Official/AOG-TaskController/pull/76


gunicsba (approved):

gunicsba:
@MarekJarczewwski  we've also added a fix inside the TC: 
https://github.com/AgOpenGPS-Official/AOG-TaskController/pull/79

I think AgOpenGPS isn't the one to blame here at the first place.

PR #1225, opened 2026-09-20
Fix NTRIP client dropping VRS/nearest-base mount points (Centipede NEAR)

gunicsba:
Follow-up to #1219 — fixes the Centipede NEAR regression reported there.

## What breaks

`OnAddMessage()` returns early when the caster's `200 OK` arrives in a read of its own, with no RTCM behind it:

```csharp
int bodyStart = headerEnd >= 0 ? headerEnd + 4 : data.Length;
if (bodyStart >= data.Length) return;      // <-- skips the watchdog reset below
```

That early return jumps over the only `NTRIP_Watchdog = 0;` in the file. The watchdog is never cleared anywhere else and is initialised to `100`, so on the very next `IncrementNTRIPWatchDog()` tick:

```csharp
if (NTRIP_Watchdog++ > 30 && isNTRIP_Connected) ReconnectRequest();
```

…the freshly accepted connection is torn down, one second after `NTRIP - Caster accepted connection` is logged. `ReconnectRequest()` also disposes `tmr`, so the GGA timer never fires and no position is ever sent.

Ordinary mount points survive this by luck — their first RTCM frame lands inside the same second and clears the watchdog in time. VRS / nearest-base mount points send **nothing at all** until they know where the rover is, so they never clear it and fail every time. Hence "works with different casters" but not with NEAR.

## Changes

- **Reset `NTRIP_Watchdog` as soon as the caster accepts**, not only on first data. This is the actual fix.
- **Send a GGA immediately on accept** instead of waiting for the 5 second timer tick — a caster may drop an idle connection before that, and NEAR needs a position before it will pick a base at all.
- **Treat `SOURCETABLE 200 OK` as a rejection** and name the unknown mount point in the message. A caster answers an unknown mount point that way, and it used to satisfy the generic `headerText.Contains("200")` check and be reported as a successful connection.
- **Scan the whole read for the RTCM3 sync byte** instead of testing only `data[0]`. A read can start in the middle of a frame, so a perfectly healthy stream could leave `bytesSinceConnected == 0` and produce a false *"Caster closed the connection without sending any data (check mountpoint/credentials)"* popup on a later disconnect.

## Testing

Built the full solution in Release and ran AgIO against a normal caster (port 2101, mount KORMOS):

```
18:15:32-> NTRIP - Caster IP resolved to: <caster ip>
18:15:45-> NTRIP - Connecting to IP: <caster ip>:2101 To Port: 2233 Mount: KORMOS
18:15:46-> NTRIP - Caster accepted connection
```

Stable, no reconnect cycling, no rejection messages — so no regression on the ordinary path.

**Not yet verified against Centipede NEAR itself.** @aortner @richardklasens — if one of you can point a build at `caster.centipede.fr` / `NEAR` that would confirm it end to end. The failure mechanism above is deterministic from the code, but a real NEAR session is the proof.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


gunicsba:
I did try it during testing and it worked

PR #1226, opened 2026-09-20
Fix silent NTRIP reconnect loop on delayed-stream mountpoints

aortner:
The watchdog reset was skipped whenever the caster's "200 OK" response arrived without RTCM data bundled in the same socket read, since that path returned early before reaching the reset further down. The watchdog counter free-runs otherwise, so it would cross its threshold within a tick and silently reconnect (ReconnectRequest() logs nothing)
- breaking mountpoints where the caster needs a moment after accepting the connection before it starts streaming, e.g. a caster-side "nearest mountpoint" alias that waits for the client's GGA position first.


gunicsba (approved):
LGTM

aortner:
please add

gunicsba:
@richardklasens  ?

